PR #5, opened 2022-09-09
SAA-32 making DB connection SSL configurable for running RDS instancein secure mode and non secure on local containers.

matt-crowson-moj:


run-local.sh
```diff
@@ -0,0 +1,23 @@
...
+# around removing the SSL connection to the database and setting the DB properties, SERVER_PORT and client credentials
+# to match those used in the docker-compose files.
+#
+
+# Provide the DB connection details to local container-hosted Postgresql DB running
+export DB_SERVER=localhost
+export DB_NAME=activities-management-db
+export DB_USER=activities-management
+export DB_PASS=activities-management
+export DB_SSL_MODE=prefer
```
timharrison-moj, line 17: I've set these values for the DB in the UI `docker-compose` file - for when we run API + UI locally, but rely on DEV for auth, prison-api, prisoner-search etc.. I'll send you the .env file required.

timharrison-moj (approved):
LGTM

PR #6, opened 2022-09-09
SAA-32 add Postgres root cert to the Dockerfile to enable secure SSL connections to the RDS db.

matt-crowson-moj:


timharrison-moj (approved):
LGTM

PR #8, opened 2022-09-12
SAA-32 second attempt at adding postgres root cert to the DockerFile to enable secure SSL connections to the RDS db.

matt-crowson-moj:


Dockerfile
```diff
@@ -7,6 +7,10 @@ WORKDIR /app
 ADD . .
 RUN ./gradlew --no-daemon assemble
 
+# Grab AWS RDS Root cert
+RUN apt-get update && apt-get install -y curl
```
SimonMitchellMOJ, line 11: no need for line 22 apt-get install -y curl && \ now as installing here

Dockerfile
```diff
@@ -25,9 +29,8 @@ RUN addgroup --gid 2000 --system appgroup && \
     adduser --uid 2000 --system appuser --gid 2000
 
 # Install AWS RDS Root cert into Java truststore
-RUN mkdir /home/appuser/.postgresql \
-  && curl https://s3.amazonaws.com/rds-downloads/rds-ca-2019-root.pem \
-    > /home/appuser/.postgresql/root.crt \
+RUN mkdir /home/appuser/.postgresql
+COPY --from=builder --chown=appuser:appgroup /app/root.crt /home/appuser/.postgresql/root.crt
```
SimonMitchellMOJ, line 33: need to add `COPY --from=builder --chown=appuser:appgroup /app/run.sh /app` at line 38

matt-crowson-moj: We don't have this as part of build process at the moment.

SimonMitchellMOJ (approved):
LGTM

PR #10, opened 2022-09-13
SAA-20 renaming Postgres container for consistency.

matt-crowson-moj:


docker-compose.yml
```diff
@@ -4,7 +4,7 @@ services:
     image: postgres:latest
     networks:
       - hmpps
-    container_name: activities-management-database
+    container_name: activities-db
```
matt-crowson-moj, line 7: I appreciate we may change the way we spawn this but for now at least this is consistent.

timharrison-moj (approved):
LGTM

PR #13, opened 2022-09-15
SAA-59 adding repository and tests for creation and retrieval of activities.

matt-crowson-moj:


timharrison-moj (approved):
LGTM. In some projects, including CVL, we took the approach that the repositories (being standard components) were probably tested as part of the JPA release, so avoided specific repo tests. I don't mind either way though, as integration tests will verify that data is being persisted and retrieved correctly.

PR #14, opened 2022-09-16
SAA-60 changing ID's from Int to Long.

matt-crowson-moj:


timharrison-moj (commented):
Is there a DDL change too, to use `bigserial` over `serial` types?

matt-crowson-moj:
> Is there a DDL change too, to use `bigserial` over `serial` types?

So technically there should be but I avoided as you said you were going to change the model.

timharrison-moj (approved):
LGTM.
I'll add the bigserial bits as part of the checking shortly.

PR #15, opened 2022-09-16
SAA-60 (WIP) initial changes to support transformation of activity entity into the external API equivalents

matt-crowson-moj:
This is the initial work in progress commit to begin process of transforming a persisted activity into it's external API equivalent.

The model objects as they stand lack Swagger documentation.  This will be added in due course.


timharrison-moj (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/model/Activity.kt
```diff
@@ -0,0 +1,36 @@
...
+
+import java.time.LocalDate
+import java.time.LocalDateTime
+
+// TODO swagger docs
+data class Activity(
+
+  val id: Long,
+
+  val prisonCode: String,
```
matt-crowson-moj, line 11: So as mentioned I have a question over the duplication of prison code across the various entities.  For now I am only capturing/returning here.

PR #16, opened 2022-09-16
SAA-65 data model checks and amendments

timharrison-moj:
* Removed redundant `prison_code` from tables, entities and model objects that did not need it.
* Removed redundant location info - now only in ACTIVITY_SESSION.
* Linked the waiting list to the ACTIVITY instead of ACTIVITY_SESSION.
* Altered some column names for consistency of tense, meaning.
* Added transforms for entity-> model waiting list.
* Added indexes where needed - particularly on the referencing side of FK relationships.
* Adjusted ACTIVITY_TIER so it has an ID, CODE and DESC like categories (tiers can be 1, 2 or FOUNDATION)
* Adjusted tests to pass.

I've updated the model diagram to reflect the same changes - so they are now in line though we could use IJ to generate the  diagrams.

THIS IS A BREAKING CHANGE - so schema will need to be dropped prior to deploying.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Activity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Activity.kt
@@ -29,7 +29,7 @@ data class Activity(
   val activityCategory: ActivityCategory,
 
   @OneToOne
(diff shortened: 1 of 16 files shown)


matt-crowson-moj (dismissed):
LGTM

matt-crowson-moj:
One thing I noticed is the foreign keys are still of type `integer`.  They should probably be `bigint` as this is what `bigserial` will create for the primary keys types.

PR #17, opened 2022-09-20
SAA-66 Snake yaml security patch.

matt-crowson-moj:
The patch is needed again due to OWASP job picking this up.  Taking the next version of the dps gradle boot plugin should resolve this,


timharrison-moj (approved):
LGTM

PR #35, opened 2022-10-06
SAA-83 replace temporary property check with something more specific for turning JMS on/off.

matt-crowson-moj:


smcveigh941 (approved):
Not too sure what this is but cant be anything too bad :D

PR #40, opened 2022-10-12
SAA-108 running the service in dev with the spring DEV profile to only include seed data in that env.

matt-crowson-moj:


yasinmustafa (approved):
LGTM

PR #49, opened 2022-10-24
SAA-129 initial commit for batch marking of attendance records.

matt-crowson-moj:
This is a first cut of batch marking of attendance records.

This is taking a very thin slice in terms of what is actually updated, in this case it is purely updating the attendance reason for a pre-generated attendance record (note attendance records are not yet pre-generated).

There is no validation taking place as part of this PR e.g. unknown attendance reason codes, attendance record not found etc.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/model/request/AttendanceUpdateRequest.kt
```diff
@@ -0,0 +1,13 @@
+package uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.model.request
+
+import io.swagger.v3.oas.annotations.media.Schema
+
+@Schema(description = "Request object for updating an attendance record")
+data class AttendanceUpdateRequest(
```
markford-moj, line 6: Just a note for next iteration - we might consider a version attribute for optimistic locking

markford-moj (approved):
Reviewed in the context of a first cut s described (no validation etc)

PR #52, opened 2022-10-25
SAA-129 adding unique constraint to attendances. A prisoner can only have one record per scheduled instance.

matt-crowson-moj:
Contains breaking change to the DB in the base line script.  Do no merge!


src/main/resources/migration/common/V1__baseline.sql
```diff
@@ -168,6 +168,7 @@ CREATE TABLE attendance (
 CREATE INDEX idx_attendance_scheduled_instance_id ON attendance (scheduled_instance_id);
 CREATE INDEX idx_attendance_prisoner_number ON attendance (prisoner_number);
 CREATE INDEX idx_attendance_recorded_time ON attendance (recorded_time);
+CREATE UNIQUE INDEX idx_attendance_scheduled_instance_id_prison_number ON attendance (scheduled_instance_id, prisoner_number);
```
matt-crowson-moj, line 171: We can only have one attendance record per prisoner per scheduled instance.

smcveigh941 (approved):

PR #53, opened 2022-10-26
SAA-133 updating to latest DPS gradle spring boot to resolve security  warnings. Also bumped other dependencies as part of this change.

matt-crowson-moj:


yasinmustafa (approved):
LGTM

PR #56, opened 2022-10-28
SAA-135: Create session instances on a rolling 60 day basis

smcveigh941:
- Schedule instances of activity sessions on a rolling 60 day basis.
- Endpoint is secured on an ingress level
- Job is kicked off asynchronously. Response is served up and processing continues in background.
- `dayOfWeek` changed from `0000100` format to just a boolean representing a day of the week.
- Ability to suspend an activity schedule between 2 dates
- Reset identity counter after seeding data manually.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/ActivitySchedule.kt
```diff
@@ -50,9 +51,5 @@ data class ActivitySchedule(
 
   val capacity: Int,
 
-  val daysOfWeek: String
-) {
-  init {
-    // TODO - ensure the days of the week is valid or introduce value type?
-  }
-}
+  val dayOfWeek: Int
```
matt-crowson-moj, line 54: I am kinda on the fence with this.  I prefer richer types.  Given this is now essentially a DayOfWeek.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/job/CreateActivitySessionsJob.kt
```diff
@@ -0,0 +1,41 @@
...
+class CreateActivitySessionsJob(private val activityRepository: ActivityRepository) {
+
+  @Value("\${jobs.create-activity-sessions.days-in-advance}")
+  private val daysInAdvance: Long? = null
+
+  @Async("asyncExecutor")
+  fun execute() {
+    val day = LocalDate.now().plusDays(daysInAdvance!!)
+
+    activityRepository.getAllForDate(day).parallelStream().forEach { activity ->
```
matt-crowson-moj, line 21: Would be nice to make the filter logic a bit easier to read.  Some extension functions perhaps ...

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/job/CreateActivitySessionsJob.kt
```diff
@@ -0,0 +1,41 @@
...
+      activity.schedules.filter {
+        return@filter DayOfWeek.of(it.dayOfWeek).equals(day.dayOfWeek) && it.suspensions.find { suspension ->
+          !day.isBefore(suspension.suspendedFrom) && (suspension.suspendedUntil === null || !day.isAfter(suspension.suspendedUntil))
+        } === null
+      }.forEach { schedule ->
+        if (schedule.instances.find { it.sessionDate == day } === null) {
```
matt-crowson-moj, line 27: Assuming I understand the logic/intent here you can use `schedule.instances.none { it.sessionDate == day }` and do away with null check.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/repository/ActivityRepository.kt
```diff
@@ -1,8 +1,14 @@
...
+import org.springframework.data.repository.query.Param
 import org.springframework.stereotype.Repository
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.entity.Activity
+import java.time.LocalDate
 
 @Repository
-interface ActivityRepository : JpaRepository<Activity, Long>
+interface ActivityRepository : JpaRepository<Activity, Long> {
+  @Query(value = "from Activity a where a.startDate <= :date and (a.endDate is null or a.endDate >= :date)")
+  fun getAllForDate(@Param("date") date: LocalDate?): List<Activity>
```
matt-crowson-moj, line 13: Date param would never be null would it??? `date: LocalDate?`

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/config/ResourceServerConfiguration.kt
```diff
@@ -30,7 +30,8 @@ class ResourceServerConfiguration {
           "/v3/api-docs/**",
           "/swagger-ui/**",
           "/swagger-ui.html",
-          "/h2-console/**"
+          "/h2-console/**",
+          "/job/**" // This endpoint is secured in the ingress rather than the app so that it can be called from within the namespace without requiring authentication
```
matt-crowson-moj, line 34: Perhaps add comment to controller also?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/job/JobTrigger.kt
```diff
@@ -0,0 +1,22 @@
...
+import org.springframework.http.MediaType
+import org.springframework.web.bind.annotation.PostMapping
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.ResponseBody
+import org.springframework.web.bind.annotation.ResponseStatus
+import org.springframework.web.bind.annotation.RestController
+
+@RestController
+@RequestMapping("/job", produces = [MediaType.APPLICATION_JSON_VALUE])
+class JobTrigger(private val createActivitySessionsJob: CreateActivitySessionsJob) {
```
matt-crowson-moj, line 13: I guess why is the controller here and not in the resource package?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/job/CreateActivitySessionsJob.kt
matt-crowson-moj, line 21: Is the parallel stream really necessary at this point?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/job/CreateActivitySessionsJob.kt
```diff
@@ -0,0 +1,41 @@
...
+
+  @Value("\${jobs.create-activity-sessions.days-in-advance}")
+  private val daysInAdvance: Long? = null
+
+  @Async("asyncExecutor")
+  fun execute() {
+    val day = LocalDate.now().plusDays(daysInAdvance!!)
+
+    activityRepository.getAllForDate(day).parallelStream().forEach { activity ->
+      activity.schedules.filter {
```
matt-crowson-moj, line 22: Kinda feels like some of this logic can be pushed in to the entities/domain.  Would be nice to enrich the domain otherwise they just become glorified DTO's.  Just my opinion.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/job/CreateActivitySessionsJob.kt
```diff
@@ -0,0 +1,41 @@
...
+        return@filter DayOfWeek.of(it.dayOfWeek).equals(day.dayOfWeek) && it.suspensions.find { suspension ->
+          !day.isBefore(suspension.suspendedFrom) && (suspension.suspendedUntil === null || !day.isAfter(suspension.suspendedUntil))
+        } === null
+      }.forEach { schedule ->
+        if (schedule.instances.find { it.sessionDate == day } === null) {
+          schedule.instances.add(
+            ScheduledInstance(
```
matt-crowson-moj, line 29: Perhaps look at adding (secondary) constructor to the scheduled instance to so start time and end time can be applied in the constructor from the schedule as that is essentially where they are coming from anyway.

```
ScheduledInstance(
  activitySchedule = schedule,
  sessionDate = day
)
```

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/config/AsyncConfiguration.kt
```diff
@@ -0,0 +1,23 @@
...
+import org.springframework.scheduling.annotation.EnableAsync
+import org.springframework.scheduling.concurrent.ThreadPoolTaskExecutor
+import java.util.concurrent.Executor
+
+@Configuration
+@EnableAsync
+class AsyncConfiguration {
+
+  @Bean
+  fun asyncExecutor(): Executor? {
```
matt-crowson-moj, line 14: Just a Kotlin style suggestion.  A more functional(ish) way of doing it ...

 ```
 fun asyncExecutors(): Executor? = ThreadPoolTaskExecutor().apply {
    corePoolSize = 3
    maxPoolSize = 3
    queueCapacity = 100
    threadNamePrefix = "AsyncThread-"
    initialize()
  }
```

File: src/main/resources/migration/common/V1__baseline.sql [hunk 10]
```diff
@@ -134,6 +133,13 @@ CREATE INDEX idx_activity_schedule_end_time ON activity_schedule (end_time);
 CREATE INDEX idx_activity_schedule_internal_location_id ON activity_schedule (internal_location_id);
 CREATE INDEX idx_activity_schedule_internal_location_code ON activity_schedule (internal_location_code);
 
+CREATE TABLE activity_schedule_suspension (
+    activity_schedule_suspension_id bigserial NOT NULL CONSTRAINT activity_schedule_suspension_id PRIMARY KEY,
+    activity_schedule_id            bigint    REFERENCES activity_schedule (activity_schedule_id),
```
matt-crowson-moj, line 138: activity_schedule_id NOT NULL?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/ActivitySchedule.kt
markford-moj: https://docs.oracle.com/javase/8/docs/api/java/time/DayOfWeek.html ?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/job/CreateActivitySessionsJob.kt
```diff
@@ -0,0 +1,41 @@
...
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.entity.ScheduledInstance
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.repository.ActivityRepository
+import java.time.DayOfWeek
+import java.time.LocalDate
+
+@Component
+class CreateActivitySessionsJob(private val activityRepository: ActivityRepository) {
+
+  @Value("\${jobs.create-activity-sessions.days-in-advance}")
+  private val daysInAdvance: Long? = null
```
matt-crowson-moj, line 15: Why not just make this part of the constructor?  You can still wire in the property/env var.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/job/CreateActivitySessionsJobTest.kt
```diff
@@ -0,0 +1,108 @@
...
+class CreateActivitySessionsJobTest {
+  private val repository: ActivityRepository = mock()
+  private val job = CreateActivitySessionsJob(repository)
+
+  @Captor
+  private lateinit var activitySaveCaptor: ArgumentCaptor<Activity>
+
+  @BeforeEach
+  fun setup() {
+    ReflectionTestUtils.setField(job, "daysInAdvance", 60L)
```
matt-crowson-moj, line 35: If the days were part of the constructor then this wouldn't be necessary.  Feels unnecessary.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/job/CreateActivitySessionsJob.kt
```diff
@@ -0,0 +1,41 @@
...
+
+  @Async("asyncExecutor")
+  fun execute() {
+    val day = LocalDate.now().plusDays(daysInAdvance!!)
+
+    activityRepository.getAllForDate(day).parallelStream().forEach { activity ->
+      activity.schedules.filter {
+        return@filter DayOfWeek.of(it.dayOfWeek).equals(day.dayOfWeek) && it.suspensions.find { suspension ->
+          !day.isBefore(suspension.suspendedFrom) && (suspension.suspendedUntil === null || !day.isAfter(suspension.suspendedUntil))
+        } === null
```
matt-crowson-moj, line 25: A style question.  Is there a reason you are using the `'==='` ? I very rarely use this so curious to know if there is a reason for it.

matt-crowson-moj:
A general comment.  Perhaps add some logging.  I appreciate we should log only when necessary but might be useful in the early stages.

helm_deploy/hmpps-activities-management-api/values.yaml
```diff
@@ -15,6 +15,15 @@ generic-service:
...
     tlsSecretName: activities-api-cert
+    path: /
+    annotations:
+      # Secure the endpoints related to k8s jobs from outside the Kubernetes ingress
+      nginx.ingress.kubernetes.io/configuration-snippet: |
+        server_tokens off;
+        location /job/* {
+          deny all;
+          return 401;
+        }
```
timharrison-moj, line 27: Maybe we can secure the endpoint with a token/role, so it can be called by us?  We could then script a call to fire off the jobs as required?  Alternative is to use kubectl to schedule the job so fine for now.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/ActivityScheduleSuspension.kt
```diff
@@ -0,0 +1,26 @@
...
+  @GeneratedValue(strategy = GenerationType.IDENTITY)
+  val activityScheduleSuspensionId: Long? = null,
+
+  @ManyToOne
+  @JoinColumn(name = "activity_schedule_id", nullable = false)
+  val activitySchedule: ActivitySchedule,
+
+  val suspendedFrom: LocalDate,
+
+  val suspendedUntil: LocalDate? = null,
```
timharrison-moj, line 25: This is a great start to allow several planned suspensions to be active.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/model/Suspension.kt
```diff
@@ -0,0 +1,20 @@
...
+  @Schema(description = "The date from which the activity schedule was suspended", example = "02/09/2022")
+  @JsonFormat(pattern = "dd/MM/yyyy")
+  val suspendedFrom: LocalDate,
+
+  @Schema(
+    description = "The date until which the activity schedule was suspended. If null, the schedule is suspended indefinately",
+    example = "02/09/2022"
+  )
+  @JsonFormat(pattern = "dd/MM/yyyy")
+  val suspendedUntil: LocalDate? = null,
```
timharrison-moj, line 19: Wondering if we need a reason for suspension? Can be added later though - depends how the UI will create them.

src/main/resources/migration/common/V1__baseline.sql
```diff
@@ -118,14 +118,13 @@ CREATE TABLE activity_schedule (
...
-  suspend_until                 date,
   start_time                    time         NOT NULL,
   end_time                      time,
   internal_location_id          integer,
   internal_location_code        varchar(40),
   internal_location_description varchar(100),
   capacity                      integer      NOT NULL,
-  days_of_week                  character(7) NOT NULL
+  day_of_week                   integer      NOT NULL
 );
```
timharrison-moj, line 128: I'm not clear on how we're representing the days of week?

src/main/resources/migration/data/R__2_2__maths_level_1_activity_schedule.sql
```diff
@@ -1,15 +1,17 @@
...
-values (3, 1, 'Wednesday AM Houseblock 3', null, '10:00:00', '11:00:00', 1, 'EDU-ROOM-1', 'Education - R1', 10, '0010000');
+insert into activity_schedule(activity_schedule_id, activity_id, description, start_time, end_time, internal_location_id, internal_location_code, internal_location_description, capacity, day_of_week)
+values (3, 1, 'Wednesday AM Houseblock 3', '10:00:00', '11:00:00', 1, 'EDU-ROOM-1', 'Education - R1', 10, 3);
 
 -- Wednesday PM
-insert into activity_schedule(activity_schedule_id, activity_id, description, suspend_until, start_time, end_time, internal_location_id, internal_location_code, internal_location_description, capacity, days_of_week)
-values (4, 1, 'Wednesday PM Houseblock 3', null, '14:00:00', '15:00:00', 2, 'EDU-ROOM-2', 'Education - R2', 10, '0010000');
+insert into activity_schedule(activity_schedule_id, activity_id, description, start_time, end_time, internal_location_id, internal_location_code, internal_location_description, capacity, day_of_week)
```
timharrison-moj, line 14: Days of week probably needs to encompass several different days?

src/main/resources/migration/data/R__3_2__english_level_1_activity_schedule.sql
```diff
@@ -1,15 +1,17 @@
...
-values (7, 2, 'Thursday AM Houseblock 3', null, '10:00:00', '11:00:00', 1, 'EDU-ROOM-1', 'Education - R1', 10, '0001000');
+insert into activity_schedule(activity_schedule_id, activity_id, description, start_time, end_time, internal_location_id, internal_location_code, internal_location_description, capacity, day_of_week)
+values (7, 2, 'Thursday AM Houseblock 3', '10:00:00', '11:00:00', 1, 'EDU-ROOM-1', 'Education - R1', 10, 4);
 
 -- Thursday PM
-insert into activity_schedule(activity_schedule_id, activity_id, description, suspend_until, start_time, end_time, internal_location_id, internal_location_code, internal_location_description, capacity, days_of_week)
-values (8, 2, 'Thursday PM Houseblock 3', null, '14:00:00', '15:00:00', 2, 'EDU-ROOM-2', 'Education - R2', 10, '0001000');
+insert into activity_schedule(activity_schedule_id, activity_id, description, start_time, end_time, internal_location_id, internal_location_code, internal_location_description, capacity, day_of_week)
+values (8, 2, 'Thursday PM Houseblock 3', '14:00:00', '15:00:00', 2, 'EDU-ROOM-2', 'Education - R2', 10, 4);
```
timharrison-moj, line 15: Scheduling several days for a session?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/TransformFunctionsTest.kt
```diff
@@ -75,7 +75,7 @@ class TransformFunctionsTest {
           startTime = timestamp.toLocalTime(),
           endTime = timestamp.toLocalTime(),
           capacity = 1,
-          daysOfWeek = "0000001"
+          dayOfWeek = "1"
         )
```
timharrison-moj, line 79: Multiple days for this schedule?

timharrison-moj (commented):
Looking good! Great start to this, though added a few questions and comments to make sure we've got the requirement covered ok.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/ActivitySchedule.kt
smcveigh941: Sorted - using DayOfWeek type where the db column is integer 0-6 representing the ordinal of DayOfWeek

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/job/CreateActivitySessionsJob.kt
smcveigh941: Thanks didnt know about this

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/repository/ActivityRepository.kt
smcveigh941: Correct

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/config/ResourceServerConfiguration.kt
smcveigh941: Yes added

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/job/JobTrigger.kt
smcveigh941: I placed it into a `job` package to because I don't think its really classed as an resource - not sure these endpoints should even have Swagger docs because they are not visable/usable to consumers and only available from within the namespace

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/job/CreateActivitySessionsJob.kt
smcveigh941: At this point we only have a tiny number of activities so not necessary - but in future we will have lots of activities and this job could be slow running so why not use parallel stream from the off? Any disadvantages of introducing it now?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/config/AsyncConfiguration.kt
smcveigh941: Thank you

src/main/resources/migration/common/V1__baseline.sql
smcveigh941: Yes thank you

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/job/CreateActivitySessionsJob.kt
smcveigh941: Guess it makes sense, thanks

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/job/CreateActivitySessionsJob.kt
smcveigh941: What do you use instead? I think I'm being verbose just

helm_deploy/hmpps-activities-management-api/values.yaml
smcveigh941: Could do... but then the cronjob needs to call hmppsAuth to get a token. This is the method other projects have used.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/model/ActivitySchedule.kt
```diff
@@ -47,5 +46,5 @@ data class ActivitySchedule(
   val capacity: Int,
 
   @Schema(description = "The days of the week on which the schedule takes place", example = "Mon,Tue,Wed")
-  val daysOfWeek: String
```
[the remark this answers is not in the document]
smcveigh941: Does now :)

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/model/Suspension.kt
smcveigh941: Was thinking the same. Concluded that it can be added later based on UI

src/main/resources/migration/common/V1__baseline.sql
smcveigh941: Each row in `activity_schedule` would be a repeating occurance for a single day of the week. If you need it on multiple days in the week then its just added as a new row with a different dayOfWeek value. This also allows for different times, location, capacity etc on different days of the week.

src/main/resources/migration/data/R__2_2__maths_level_1_activity_schedule.sql
smcveigh941: Multiple rows for multiple days - each with their own values for location, capacity, dayOfWeek, times etc

src/main/resources/migration/data/R__3_2__english_level_1_activity_schedule.sql
smcveigh941: As above - message me to discuss if not clear and/or not happy

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/TransformFunctionsTest.kt
smcveigh941: ⬆️

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/job/CreateActivitySessionsJob.kt
smcveigh941: Did the suggested thanks

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/job/CreateActivitySessionsJob.kt
smcveigh941: Added extension functions and it looks much nicer thanks

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/job/CreateActivitySessionsJob.kt
smcveigh941: Pushed some of this into the entity, thanks

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/job/CreateActivitySessionsJob.kt
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  @Async("asyncExecutor")
+  fun execute() {
+    val day = LocalDate.now().plusDays(daysInAdvance!!)
+
+    log.info("Scheduling all activities on $day")
+
+    activityRepository.getAllForDate(day).parallelStream().forEach { activity ->
+      activity
```
matt-crowson-moj, line 29: It's just occurred to me an Activity has an active flag. I presume we would skip these when scheduling? The repository query does not distinguish at the mo. @timharrison-moj

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/job/JobTrigger.kt
matt-crowson-moj: Agreed, no to swagger docs.  I guess I am used to see controllers/resources in a similar location.  We could have another package in controllers e.g. internal?

I am not going to loose sleep over it, it is just a gut feel.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/repository/ActivityRepository.kt
```diff
@@ -1,8 +1,14 @@
...
+import org.springframework.data.jpa.repository.Query
+import org.springframework.data.repository.query.Param
 import org.springframework.stereotype.Repository
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.entity.Activity
+import java.time.LocalDate
 
 @Repository
-interface ActivityRepository : JpaRepository<Activity, Long>
+interface ActivityRepository : JpaRepository<Activity, Long> {
+  @Query(value = "from Activity a where a.startDate <= :date and (a.endDate is null or a.endDate >= :date)")
```
matt-crowson-moj, line 12: See comment on the job.  Should this only be returning active activities? or include a param to only include active?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/ActivitySchedule.kt
```diff
@@ -50,9 +52,30 @@ data class ActivitySchedule(
 
   val capacity: Int,
 
-  val daysOfWeek: String
+  val mondayFlag: Boolean,
```
matt-crowson-moj, line 55: Could probably just call them monday, tuesday etc, meh!

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/ActivitySchedule.kt
```diff
@@ -50,9 +52,30 @@ data class ActivitySchedule(
...
+
+  val fridayFlag: Boolean,
+
+  val saturdayFlag: Boolean,
+
+  val sundayFlag: Boolean,
 ) {
-  init {
-    // TODO - ensure the days of the week is valid or introduce value type?
+  fun getDaysOfWeek(): List<DayOfWeek> {
```
matt-crowson-moj, line 69: A more Kotlin way of doing the same (meh):

```
  fun daysOfWeek(): List<DayOfWeek> =
    mutableListOf<DayOfWeek>().apply {
      if (mondayFlag) add(DayOfWeek.MONDAY)
      if (tuesdayFlag) add(DayOfWeek.TUESDAY)
      ......
    }
```

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/job/CreateActivitySessionsJob.kt
smcveigh941: We are scheduling 60 days in advance. Just because an activity is not active now doesnt mean it wont be active in 60 days time. If the active field is just calculated based on start and end dates then it shouldnt matter

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/repository/ActivityRepository.kt
smcveigh941: We are scheduling 60 days in advance. Just because an activity is not active now doesnt mean it wont be active in 60 days time. If the active field is just calculated based on start and end dates then it shouldnt matter

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/job/CreateActivitySessionsJob.kt
```diff
@@ -0,0 +1,47 @@
...
+class CreateActivitySessionsJob(
+  private val activityRepository: ActivityRepository,
+  @Value("\${jobs.create-activity-sessions.days-in-advance}") private val daysInAdvance: Long? = null
+) {
+
+  companion object {
+    private val log = LoggerFactory.getLogger(this::class.java)
+  }
+
+  @Async("asyncExecutor")
```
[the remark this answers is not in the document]
smcveigh941: Only need to add a custom async exception handler if you want to do something other than print out the error to logs. For the case of this job, the response is 201 indicating that the job has kicked off and there continues to be processing taking place after the response has been served up. Errors will still continue to be logged in app insights and slack-alerts I'm pretty sure

matt-crowson-moj (approved):
LGTM.  I have a few comments but does not stop it from being merged IMO.  You will need to drop the dev DB prior to merging into main.

timharrison-moj (approved):
LGTM

PR #62, opened 2022-11-03
SAA-156 Add validation to API Endpoint: (prisoner scheduled instances)

markford-moj:
Validate the following

- Incorrect date formats - example error - Error converting 'endDate' (01/10/2022): Failed to convert value of type 'java.lang.String' to required type 'java.time.LocalDate'
- Exceeding 3 month date range limit - Validation failure: Date range cannot exceed 3 months
- Missing start date param - Required request parameter 'startDate' for method parameter type LocalDate is not present
- Missing end date param - Required request parameter 'endDate' for method parameter type LocalDate is not present

Changes

- Make controller advice extend ResponseEntityExceptionHandler and override handlerExceptionInternal
- New controller advice function - handleMethodArgumentTypeMismatchException
- Validate date range

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/resource/ActivityScheduleInstanceController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/resource/ActivityScheduleInstanceController.kt
@@ -20,6 +20,7 @@ import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.config.ErrorRes
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.model.ActivityScheduleInstance
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.service.ScheduledInstanceService
 import java.time.LocalDate
+import javax.validation.ValidationException
 
 @RestController
 @RequestMapping("/prisons/{prisonCode}/prisoners/{prisonerNumber}/scheduled-instances", produces = [MediaType.APPLICATION_JSON_VALUE])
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/resource/ActivityScheduleInstanceController.kt
```diff
@@ -59,9 +60,12 @@ class ActivityScheduleInstanceController(private val scheduledInstanceService: S
     @PathVariable("prisonCode") prisonCode: String,
     @PathVariable("prisonerNumber") prisonerNumber: String,
     @RequestParam(value = "startDate", required = true) @DateTimeFormat(iso = DateTimeFormat.ISO.DATE) @Parameter(description = "Start date of query") startDate: LocalDate,
-    @RequestParam(value = "endDate", required = true) @DateTimeFormat(iso = DateTimeFormat.ISO.DATE) @Parameter(description = "End date of query") endDate: LocalDate,
+    @RequestParam(value = "endDate", required = true) @DateTimeFormat(iso = DateTimeFormat.ISO.DATE) @Parameter(description = "End date of query (max 3 months from start date)") endDate: LocalDate,
   ): List<ActivityScheduleInstance> {
     val dateRange = LocalDateRange(startDate, endDate)
+    if (endDate.isAfter(startDate.plusMonths(3))) {
+      throw ValidationException("Date range cannot exceed 3 months")
+    }
```
smcveigh941, line 68: Why are we making this restriction?

smcveigh941 (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/resource/ActivityScheduleInstanceController.kt
markford-moj: just an arbitrary recognition that we should protect the service with some sort of limit. No real business requirements in this area yet. Dave mentioned pagination would be another option. Im sure we will refine at some point

PR #67, opened 2022-11-08
SAA-176 add minimal attendance creation to the attendance service for the upcoming cron job.

matt-crowson-moj:
Adding simple implementation for attendance creation in the service layer for the scheduled cron job to use.  There are business rules missing for the first iteration e.g. not checking for suspensions, inactive allocations and not applying pay rates.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/repository/ScheduledInstanceRepository.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/repository/ScheduledInstanceRepository.kt
@@ -33,4 +33,6 @@ interface ScheduledInstanceRepository : JpaRepository<ScheduledInstance, Long> {
     startDate: LocalDate,
     endDate: LocalDate
   ): List<ScheduledInstance>
+
+  fun findAllBySessionDate(date: LocalDate): List<ScheduledInstance>
 }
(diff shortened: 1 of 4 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/repository/ScheduledInstanceRepository.kt
```diff
@@ -33,4 +33,6 @@ interface ScheduledInstanceRepository : JpaRepository<ScheduledInstance, Long> {
     startDate: LocalDate,
     endDate: LocalDate
   ): List<ScheduledInstance>
+
+  fun findAllBySessionDate(date: LocalDate): List<ScheduledInstance>
```
matt-crowson-moj, line 37: This will likely be replaced with a dedicated query to identify only those that don't have suspensions for a given date.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/AttendancesService.kt
```diff
@@ -44,9 +47,35 @@ class AttendancesService(
...
+        instance,
+        allocation.prisonerNumber
+      )
+    ) {
+      log.info("Attendance record already exists for allocation ${allocation.allocationId} and scheduled instance ${instance.scheduledInstanceId}")
+      return
+    }
+
+    attendanceRepository.save(
+      Attendance(
```
matt-crowson-moj, line 74: Also not setting status.

PR #68, opened 2022-11-08
SAA-162: Add endpoint to return the list of categories and capacities within a prison

smcveigh941:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/CapacityService.kt
```diff
@@ -0,0 +1,35 @@
...
+class CapacityService(private val activityRepository: ActivityRepository) {
+  fun getActivityCategoryCapacityAndAllocated(
+    prisonCode: String,
+    categoryId: Long
+  ): CapacityAndAllocated {
+    val activities = activityRepository.getAllByPrisonCodeAndActivityCategoryActivityCategoryId(
+      prisonCode,
+      categoryId
+    ).takeIf { it.isNotEmpty() }
+      ?: throw EntityNotFoundException("Category with ID $categoryId not found at prison $prisonCode")
```
matt-crowson-moj, line 20: Assuming I am understanding this correctly.  We are throwing an exception if no activities found under a particular category?

If that is the case I don't think we should be throwing an exception, it just might be there are no activities in that category.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/ActivityCategory.kt
```diff
@@ -16,4 +17,11 @@ data class ActivityCategory(
...
   val description: String
-)
+) {
+  fun toModel() = ModelActivityCategory(
+    id = activityCategoryId!!,
+    description = description
+  )
+}
+
+fun List<ActivityCategory>.toModel() = map { it.toModel() }
```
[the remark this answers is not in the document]
matt-crowson-moj: Yeah I agree it is getting rather large, not sure it should be done in the entity though as it is a presentation layer thing.  

Perhaps move it into the actual model objects?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/resource/ActivityCategoryController.kt
```diff
@@ -0,0 +1,49 @@
...
+import org.springframework.web.bind.annotation.ResponseBody
+import org.springframework.web.bind.annotation.RestController
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.config.ErrorResponse
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.entity.toModel
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.model.response.ActivityCategory
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.repository.ActivityCategoryRepository
+
+@RestController
+@RequestMapping("/activity-categories", produces = [MediaType.APPLICATION_JSON_VALUE])
+class ActivityCategoryController(private val activityCategoryRepository: ActivityCategoryRepository) {
```
matt-crowson-moj, line 21: Splitting hairs a little here.  IMO I don't think repositories should be used outside of the service layer.

I do understand though adding a service here could be considered overkill. Meh!

matt-crowson-moj: I guess another way to think of it, I see the service layer as setting the transaction boundaries e.g. in this case it could read only.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/resource/PrisonController.kt
```diff
@@ -54,16 +52,15 @@ class PrisonController(
...
-      ApiResponse(
-        responseCode = "404",
-        description = "The prison for this code was not found.",
-        content = [Content(mediaType = "application/json", schema = Schema(implementation = ErrorResponse::class))],
       )
     ]
   )
-  fun getPrisonByCode(@PathVariable("prisonCode") prisonCode: String): RolloutPrison =
-    prisonService.getByPrisonCode(prisonCode)
+  @GetMapping(value = ["/{prisonCode}/activity-categories/{categoryId}/capacity"])
```
matt-crowson-moj, line 58: Need to document a 404 can be thrown given the capacity service can throw an `EntityNotFoundException`.

matt-crowson-moj (approved):
LGTM

PR #70, opened 2022-11-09
SAA-176 adding integration test for attendance creation.

matt-crowson-moj:


src/test/resources/test_data/seed-activity-id-4.sql
```diff
@@ -0,0 +1,26 @@
+insert into activity(activity_id, prison_code, activity_category_id, activity_tier_id, summary, description, start_date, end_date, active, created_time, created_by)
+values (4, 'PVI', 1, 1, 'Maths', 'Maths Level 1', current_date, null, true, current_timestamp, 'SEED USER');
```
matt-crowson-moj, line 2: I am using current_date and current_timestamp for these.  This just simplifies the temporal nature of attendances relying on the schedule fitting in with the day it actually runs.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/integration/CreateAttendanceRecordsJobIntegrationTest.kt
```diff
@@ -0,0 +1,79 @@
...
+        assertThat(instances.first().attendances).hasSize(2)
+      }
+    }
+
+    assertThat(attendanceRepository.count()).isEqualTo(4)
+  }
+
+  @Sql("classpath:test_data/seed-activity-id-4.sql")
+  @Test
+  fun `Multiple calls on same day does not result in duplicate attendances`() {
```
matt-crowson-moj, line 59: Technically the DB actually won't allow this to happen, due to a unique constraint.

I guess what we really need to say here is we can rerun the process on the same day should the need arise, so perhaps this is misleading?

markford-moj (approved):

PR #72, opened 2022-11-10
Saa 176 only create attendance records for active allocations

matt-crowson-moj:
Attendances should only be created for offender allocations that are active on the given date the attendances are due to be created.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Activity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Activity.kt
@@ -57,6 +57,7 @@ data class Activity(
 
   var endDate: LocalDate? = null,
 
+  @Deprecated(message = "To be removed. Use start and end dates in its place")
   var active: Boolean = true,
 
   val createdTime: LocalDateTime,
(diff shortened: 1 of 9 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Activity.kt
```diff
@@ -57,6 +57,7 @@ data class Activity(
 
   var endDate: LocalDate? = null,
 
+  @Deprecated(message = "To be removed. Use start and end dates in its place")
```
matt-crowson-moj, line 60: There is a separate JIRA ticket to deal with the remove of the active flag attributes.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/model/Allocation.kt
```diff
@@ -22,7 +22,7 @@ data class Allocation(
 
   @Schema(description = "The date when the prisoner will start the activity", example = "10/09/2022")
   @JsonFormat(pattern = "dd/MM/yyyy")
-  val startDate: LocalDate? = null,
+  val startDate: LocalDate,
```
matt-crowson-moj, line 25: This cannot be null the schema definition for this is already correct.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Allocation.kt
```diff
@@ -27,10 +28,11 @@ data class Allocation(
 
   var payBand: String? = null,
 
-  var startDate: LocalDate? = null,
+  var startDate: LocalDate,
```
matt-crowson-moj, line 31: This cannot be null the schema definition for this is already correct.

PR #81, opened 2022-11-15
SAA-181 AM Events Endpoint: Client for (prison api) activities endpoint

markford-moj:
New client for prison api activities
Add prison api activities call to ScheduledEventService


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/model/PrisonerScheduledEvents.kt
```diff
@@ -24,6 +24,9 @@ data class PrisonerScheduledEvents(
   @Schema(description = "A list of scheduled appointments for this prisoner in this date range")
   val appointments: List<ScheduledEvent>? = null,
```
matt-crowson-moj, line 25: Can these just be empty lists instead of null for consistency.  Asking because on all models they are empty lists/arrays.

matt-crowson-moj: Perhaps a question for wider the team.

matt-crowson-moj (approved):
LGTM

PR #85, opened 2022-11-18
SAA-160 API endpoint to get scheduled events (combined) BUGFIX

markford-moj:
Corrected prison api request param names
prison api Alert and Assessment data coming through with null offender numbers from dev - made property nullable. Throw 404 if prison code and prisoner number don't match query params 
Swagger endpoint docs fix


timharrison-moj (approved):
LGTM

matt-crowson-moj (approved):
LGTM

PR #88, opened 2022-11-23
SAA-198 adding the domain model and DB DDL for event priorities.

matt-crowson-moj:
**DO NOT MERGE, CONTAINS DB CHANGE**

This change introduces the concept of event priorities for prisons in the domain i.e. when an event takes priority over another should there be a clash because they are both at the same time.

If no priorities exist for a prison then sensible defaults are provided.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/EventPriority.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/EventPriority.kt
@@ -1,0 +1,25 @@
+package uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.entity
+
+import javax.persistence.Entity
+import javax.persistence.EnumType
+import javax.persistence.Enumerated
+import javax.persistence.GeneratedValue
+import javax.persistence.GenerationType
+import javax.persistence.Id
+import javax.persistence.Table
+
+// TODO this will need to contain more information e.g. subtype/subcategories.
+@Entity
+@Table(name = "event_priority")
+data class EventPriority(
+  @Id
+  @GeneratedValue(strategy = GenerationType.IDENTITY)
+  val eventPriorityId: Long? = null,
+
+  val prisonCode: String,
+
+  @Enumerated(EnumType.STRING)
+  val eventType: EventType,
+
+  val priority: Int
+)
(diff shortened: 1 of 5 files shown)


src/main/resources/migration/common/V1__baseline.sql
```diff
@@ -229,3 +229,11 @@ CREATE TABLE activity_pay (
 
 CREATE INDEX idx_activity_pay_activity_id ON activity_pay (activity_id);
 
+CREATE TABLE event_priority (
+  event_priority_id bigserial   NOT NULL CONSTRAINT event_priority_pk PRIMARY KEY,
```
markford-moj, line 233: just a suggestion that we call the pk column (and entity property)  'id'

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/EventPriority.kt
```diff
@@ -0,0 +1,25 @@
...
+import javax.persistence.Id
+import javax.persistence.Table
+
+// TODO this will need to contain more information e.g. subtype/subcategories.
+@Entity
+@Table(name = "event_priority")
+data class EventPriority(
+  @Id
+  @GeneratedValue(strategy = GenerationType.IDENTITY)
+  val eventPriorityId: Long? = null,
```
markford-moj, line 17: just a suggestion that we call the pk column (and entity property)  'id'

markford-moj (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/PrisonRegimeService.kt
```diff
@@ -2,13 +2,19 @@ package uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.service
...
+class PrisonRegimeService(private val eventPriorityRepository: EventPriorityRepository) {
 
-  // TODO this is stubbed for now until we have the DB, Entity and repository in place
   fun getEventPrioritiesForPrison(code: String) =
-    EventType.values().associateWith { setOf(Priority(it.defaultPriority)) }
+    eventPriorityRepository.findByPrisonCode(code)
+      .groupBy { it.eventType }
+      .mapValues { it.value.map { ep -> Priority(ep.priority) } }
+      .ifEmpty { defaultPriorities() }
+
```
timharrison-moj, line 15: Like it.

timharrison-moj (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/EventPriority.kt
matt-crowson-moj: Noted. Probably something to look at in the wider context.  All id's use a similar pattern at the moment,

PR #90, opened 2022-11-24
SAA-199 Decorate scheduled event endpoint with event priorities

markford-moj:
Altered ScheduleEventService to call PrisonRegimeService for event priorities
Added event priorities to response


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/model/ScheduledEvent.kt
```diff
@@ -52,4 +52,7 @@ data class ScheduledEvent(
   @Schema(description = "The end time for this scheduled instance", example = "10:00")
   @JsonFormat(pattern = "HH:mm")
   val endTime: LocalTime?,
+
+  @Schema(description = "The event priority")
+  val priority: Int?,
```
matt-crowson-moj, line 57: I don't think this can be null.  There should always be defaults.

matt-crowson-moj (approved):
LGTM

PR #99, opened 2022-12-01
SAA-220 Migrate whereabouts location service layer

markford-moj:
Migrate whereabouts location service layer to Activities Management API

Whereabouts location regex patterns config- BCI, CDI, EYI, HEI, MDI, NHI, RSI, WNI property files groups.properties

Location service layer

Prison API client calls

GET locations for type: /agencies/{agencyId}/locations/type/{type}

GET locations for type (Unrestricted - Does not check the invoker has the selected agency in their caseload): /agencies/{agencyId}/locations?eventType={locationType}

GET location groups: /agencies/{agencyId}/locations/groups


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/LocationService.kt
```diff
@@ -0,0 +1,48 @@
...
+    return LocationPrefixDto(locationPrefix)
+  }
+
+  fun getVideoLinkRoomsForPrison(agencyId: String): List<LocationIdAndDescription>? =
+    prisonApiClient
+      .getLocationsForTypeUnrestricted(agencyId, "APP").block()
+      ?.filter { it.locationType == "VIDE" }
+      ?.map { LocationIdAndDescription(it.locationId, it.userDescription ?: it.description) }
+
+  fun getCellLocationsForGroup(agencyId: String, groupName: String): List<Location>? =
```
matt-crowson-moj, line 38: Sorry don't know how much of this was a straight copy.

Looks to me like this might work as well without the toMutableList and toList at the end ...

```
  fun getCellLocationsForGroup(agencyId: String, groupName: String): List<Location>? =
    prisonApiClient.getLocationsForType(agencyId, "CELL").block()
      ?.filter(locationGroupService.locationGroupFilter(agencyId, groupName)::test)
      ?.map { it.copy(description = it.description.formatLocation()) }

```

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/LocationGroupService.kt
```diff
@@ -0,0 +1,10 @@
+package uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.service
+
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.client.prisonapi.model.Location
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.client.prisonapi.model.LocationGroup
+import java.util.function.Predicate
+
+interface LocationGroupService {
```
matt-crowson-moj, line 7: Again this is a lift and shift so I appreciate this may have already been like this.  I can't see the point of this interface unless I am missing something.

matt-crowson-moj:
Perhaps some additional logging in places would be useful.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/LocationGroupService.kt
markford-moj: Agreed I sort of see where they intended to go with it - dynamically injecting the appropriate implementation of the interface - but then they've just injected both implementations and chose which function to call with an if / else

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/LocationService.kt
markford-moj: Gonna take a snip of this and look to see if it works in next subtask

PR #111, opened 2022-12-06
Allow jobs to be triggered by HTTP in dev only

smcveigh941:
Environments currently block all requests to the `/jobs` route from outside of the k8s namespace. It would be convenient to hit these endpoints in dev only.


timharrison-moj (approved):
LGTM. Not that I understand what this is doing exactly!  Will allow ANY http request in DEV??

PR #125, opened 2022-12-16
SAA-247 have identified two attributes returned by prison api can actually be null/empty. This does not align with the API's docs.

matt-crowson-moj:
There is a risk here if we were to generate the DTO's for prison API again say due to a new endpoint these changes may be overwritten.

The offending endpoint : `/api/bookings/offenderNo/{prisonerNumber}?fullInfo=true`

**_Note: there may be more, these are the two I am aware of._**


timharrison-moj (approved):
LGTM

PR #132, opened 2022-12-20
SAA-258 activity category

markford-moj:
New property 'name'


matt-crowson-moj (approved):
LGTM - don't forget you will need to drop the DB prior to merging into main.

timharrison-moj (approved):
LGTM

PR #135, opened 2022-12-21
SAA-261 some minor refactoring to reduce some boiler place exception handling.

matt-crowson-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/repository/JpaRepositoryExt.kt
```diff
@@ -0,0 +1,10 @@
+package uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.repository
+
+import org.springframework.data.jpa.repository.JpaRepository
+import javax.persistence.EntityNotFoundException
+
+inline fun <reified T> JpaRepository<T, Long>.findOrThrowIllegalArgument(id: Long): T =
+  this.findById(id).orElseThrow { IllegalArgumentException("${T::class.java.simpleName} $id not found") }
+
+inline fun <reified T> JpaRepository<T, Long>.findOrThrowNotFound(id: Long): T =
```
matt-crowson-moj, line 9: We can change this if we want to allow us to override the message but I think this is fine as is in all honesty.

timharrison-moj (approved):
LGTM. Takes out the repetitive `.orElseThrow { exception(msg) }` and exception messages in each place used.

timharrison-moj (approved):
LGTM

PR #136, opened 2022-12-21
SAA-259 activity data model changes

markford-moj:
tier and description become optional

Also SAA-262 - unique constraint added to prison code and summary

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Activity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Activity.kt
@@ -34,8 +34,8 @@ data class Activity(
   val activityCategory: ActivityCategory,
 
   @OneToOne
-  @JoinColumn(name = "activity_tier_id", nullable = false)
-  val activityTier: ActivityTier,
+  @JoinColumn(name = "activity_tier_id")
+  val activityTier: ActivityTier?,
 
   @OneToMany(mappedBy = "activity", fetch = FetchType.EAGER, cascade = [CascadeType.ALL], orphanRemoval = true)
   @Fetch(FetchMode.SUBSELECT)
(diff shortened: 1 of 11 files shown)


src/main/resources/migration/common/V1__baseline.sql
```diff
@@ -89,16 +89,17 @@ CREATE TABLE activity (
...
+  description          varchar(300),
   start_date           date         NOT NULL,
   end_date             date,
   risk_level           varchar(10),
   minimum_incentive_level      varchar(10),
   created_time         timestamp    NOT NULL,
-  created_by           varchar(100) NOT NULL
+  created_by           varchar(100) NOT NULL,
+  UNIQUE (prison_code, summary)
 );
```
timharrison-moj, line 103: Is this to prevent the same activity name being created more than once per prison? Makes sense.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Activity.kt
```diff
@@ -57,7 +57,7 @@ data class Activity(
 
   val summary: String,
 
-  val description: String,
+  val description: String?,
```
timharrison-moj, line 61: Do the users have anywhere to enter a longer description at present?

src/main/resources/migration/common/V1__baseline.sql
markford-moj: It is - but not totally fool proof as @matt-crowson-moj pointed out - could still be bypassed by letter case, spurious spaces etc. He also made some good points re activity dates in this context which are worth further discussion. This might be a good starting point tho

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Activity.kt
markford-moj: No - thats why @smcveigh941 requested this change

timharrison-moj (approved):
LGTM.

PR #139, opened 2022-12-22
SAA-261 publishing activity creation event and start introduction of localstock for working/testing queues locally.

matt-crowson-moj:
We need to save and flush data to ensure timing of data being persisted and event being intercepted are in sync.

**Still to do:**

Integration test (this requires test containers for localstack and updating the CI config to pull in a localstack container).


markford-moj (dismissed):

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/ActivityPay.kt
```diff
@@ -39,7 +39,7 @@ data class ActivityPay(
     return activityPayId != null && activityPayId == other.activityPayId
   }
 
-  override fun hashCode(): Int = id.hashCode()
+  override fun hashCode(): Int = activityPayId.hashCode()
```
matt-crowson-moj, line 42: This actually showed up as an error (which it is) after changing to save and flush.

markford-moj (approved):

PR #140, opened 2022-12-22
SAA-261 disabling publishing of events to avoid raising alerts until we have something consume the message.

matt-crowson-moj:


timharrison-moj (approved):
LGTM

PR #141, opened 2022-12-23
SAA-272 Data Model Changes 14/12/2022 - DM / API Work

markford-moj:
Update the Flyway migrations with these changes + any seed data. 
Update Entity / Model (DTO) classes for these changes. 
Update service methods to build responses from the altered source entities. 
Endpoints/controller changes - the multiple slots, the additional columns. 
Tests / Integration tests / test data updates to reflect the changes.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/ActivitySchedule.kt
```diff
@@ -74,50 +75,21 @@ data class ActivitySchedule(
...
-    if (wednesdayFlag) add(DayOfWeek.WEDNESDAY)
-    if (thursdayFlag) add(DayOfWeek.THURSDAY)
-    if (fridayFlag) add(DayOfWeek.FRIDAY)
-    if (saturdayFlag) add(DayOfWeek.SATURDAY)
-    if (sundayFlag) add(DayOfWeek.SUNDAY)
-  }
-
   fun getAllocationsOnDate(date: LocalDate): List<Allocation> = this.allocations.filter {
     !date.isBefore(it.startDate) && (it.endDate == null || !date.isAfter(it.endDate))
   }
```
timharrison-moj, line 95: Should this function still be at the activity schedule level? Or at the slot level now?

timharrison-moj (approved):
Very good. Seems to have gone swimmingly well!

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/ActivitySchedule.kt
markford-moj: decision was made to leave allocations at schedule level when we discussed
I.e. when your allocated to a schedule your allocated to all the slots

PR #145, opened 2023-01-04
SAA-275 Add booking id to allocations

markford-moj:
Add booking id to allocations
Also make rollout prison date optional


timharrison-moj (approved):
LGTM

PR #147, opened 2023-01-06
SAA-283: Peer review comments and schedule changes

timharrison-moj:
Contains a few minor things:

* Address peer review comments from previous PR.
* Schedules activities for 14 days in advance in DEV (so we can see real things)
* Fixes warnings in the UniqueProperyId class used in the view


matt-crowson-moj (approved):
LGTM

PR #148, opened 2023-01-09
SAA-288 pay_per_session

markford-moj:
New Activity property payPerSession. An enum 'H' Half day or 'F' Full day - indicating whether activity payment is for a full day or half day. DB / Entity / Model updates with test data.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Activity.kt
```diff
@@ -126,3 +133,5 @@ data class Activity(
 }
 
 fun List<Activity>.toModelLite() = map { it.toModelLite() }
+
+enum class PayPerSession { H, F }
```
matt-crowson-moj, line 137: This is very minor point and more of preference on my part for clarity.

`enum class PayPerSession(val label: String) { H("Half day"), F("Full day") }`

or

`enum class PayPerSession { HALF_DAY, FULL_DAY }`

I know NOMIS uses H and F but I don't like this single character meaning business.

matt-crowson-moj (approved):
LGTM

PR #149, opened 2023-01-09
SAA-261 event needs to be raised at the schedule level and not activity level for mapping to NOMIS.

matt-crowson-moj:
We realised after doing this piece of work events around activity creation actually need to take place at the schedule level for mapping to NOMIS.


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/integration/ActivityIntegrationTest.kt
```diff
@@ -48,15 +39,6 @@ class ActivityIntegrationTest : IntegrationTestBase() {
       assertThat(eligibilityRules.size).isEqualTo(1)
       assertThat(pay.size).isEqualTo(2)
     }
-
-    verify(eventsPublisher).send(eventCaptor.capture())
-
-    with(eventCaptor.firstValue) {
```
matt-crowson-moj, line 54: We cannot test this at the integration test level at present because we do not have an endpoint to support schedule creation.

It will need to done retrospectively.

markford-moj (approved):

markford-moj (approved):

PR #150, opened 2023-01-09
SAA-292 3rd party dependency updates.

matt-crowson-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/util/TransformFunctions.kt
```diff
@@ -140,7 +140,7 @@ fun List<PrisonerScheduledActivity>.toModelScheduledEvents(
     ModelScheduledEvent(
       prisonCode = prisonCode,
       eventId = it.scheduledInstanceId,
-      bookingId = it.bookingId?.toLong() ?: 0L, // Change allocation to include bookingId
+      bookingId = it.bookingId.toLong(), // Change allocation to include bookingId
```
matt-crowson-moj, line 143: Null check unnecessary as no longer nullable.

timharrison-moj (approved):
LGTM

PR #151, opened 2023-01-09
SAA-293 recording the (missing) pricinal name against allocation creations.

matt-crowson-moj:
I missed this when I did the create allocations work and it was on my TODO/tidy up list, so here it is.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/resource/ActivityScheduleController.kt
```diff
@@ -235,10 +236,12 @@ class ActivityScheduleController(
   )
   @PreAuthorize("hasAnyRole('ACTIVITY_HUB', 'ACTIVITY_HUB_LEAD', 'ACTIVITY_ADMIN')")
   fun allocate(
+    principal: Principal,
     @PathVariable("scheduleId") scheduleId: Long,
```
timharrison-moj, line 240: Is this making an assumption that it will be called (always) with a user token rather than an admin client token?

timharrison-moj (approved):
LGTM. Just one more thing... (in Columbo style..)

timharrison-moj (approved):
LGTM

PR #152, opened 2023-01-10
SAA-279 Basic activities dashboard

markford-moj:
New activities endpoint - all activities for prison

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/resource/PrisonController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/resource/PrisonController.kt
@@ -37,6 +37,34 @@ class PrisonController(
 ) {
 
   @Operation(
+    summary = "Get list of activities at a specified prison",
+  )
+  @ApiResponses(
+    value = [
+      ApiResponse(
+        responseCode = "200",
+        description = "Activities",
+        content = [Content(mediaType = "application/json", array = ArraySchema(schema = Schema(implementation = ActivityLite::class)))],
+      ),
+      ApiResponse(
+        responseCode = "401",
+        description = "Unauthorised, requires a valid Oauth2 token",
+        content = [Content(mediaType = "application/json", schema = Schema(implementation = ErrorResponse::class))],
+      ),
+      ApiResponse(
+        responseCode = "403",
+        description = "Forbidden, requires an appropriate role",
+        content = [Content(mediaType = "application/json", schema = Schema(implementation = ErrorResponse::class))],
+      )
+    ]
+  )
+  @GetMapping(value = ["/{prisonCode}/activities"])
+  @ResponseBody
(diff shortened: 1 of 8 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/resource/PrisonController.kt
```diff
@@ -36,6 +36,39 @@ class PrisonController(
...
+        content = [Content(mediaType = "application/json", schema = Schema(implementation = ErrorResponse::class))],
+      ),
+      ApiResponse(
+        responseCode = "403",
+        description = "Forbidden, requires an appropriate role",
+        content = [Content(mediaType = "application/json", schema = Schema(implementation = ErrorResponse::class))],
+      ),
+      ApiResponse(
+        responseCode = "404",
+        description = "Category ID not found",
```
matt-crowson-moj, line 61: Copy paste error?  This would be a prison in this case if a 404 is thrown.  I don't think it is thrown though is it? if not I guess this can be removed.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/resource/PrisonController.kt
```diff
@@ -36,6 +36,39 @@ class PrisonController(
   private val scheduleService: ActivityScheduleService,
 ) {
 
+  @Operation(
+    summary = "Get list of activities at a specified prison",
```
timharrison-moj, line 40: I'm not sure this should be in a PrisonController? As the subject matter is a list of Activities, I think would be better placed into an activities controller. The fact that it is by-prison is probably additional information. 
e.g. /activities/prison/{prisonCode} ?

We could enrich later with query params like `?isActive=true`, or `?hasAllocations=true`

markford-moj: In the activities controller - the next uri token after /activities is a path param {activitiesId} - adding a path like this /activities/prison creates a confusing scenario where 'prison' could be confused with and activities id.

markford-moj: personally i would have added /prison/{prisonCode} to the start of the activity controller path - so all activity endpoints include prison code on the path.

timharrison-moj (commented):
Updated comments only.

markford-moj:
This sounds like a significant refactor task impacting the UI as well - probably best in its own PR?

timharrison-moj:
> This sounds like a significant refactor task impacting the UI as well - probably best in its own PR?

Yes, that's fine to go in as is. It can be part of the general API refactor to present consumers with logically grouped endpoints.

timharrison-moj (approved):
LGTM.

PR #155, opened 2023-01-12
SAA-300: App insights sample size and PR review comments addressed

timharrison-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/job/CreateActivitySessionsJob.kt
```diff
@@ -30,55 +30,60 @@ class CreateActivitySessionsJob(
...
+
+        filteredForDay.forEach { activity ->
+          var activityChanged = false
+          val schedules = activity.getSchedulesOnDay(day, includeSuspended = false)
+          val withNoPreExistingInstances = schedules.filterActivitySchedulesWithNoPreExistingInstance(day)
+
+          withNoPreExistingInstances.forEach { schedule ->
+            val filteredSlots = schedule.slots.filter { day.dayOfWeek in it.getDaysOfWeek() }
+            filteredSlots.filterActivityScheduleSlotsForBankHoliday(day).forEach { slot ->
+              schedule.instances.add(
```
matt-crowson-moj, line 72: Sorry just one other thought on this.  We could go a bit further and push some of this into the Domain (Entity).

For example:

`schedule.addInstance(day, slot)`

This would mean we could in future easily add business rules into this. Just a thought.

matt-crowson-moj: Indeed.  To be honest there are a few places now where this occurs so it might be better to decide how we want to approach this for all entities.

matt-crowson-moj (approved):
LGTM

PR #161, opened 2023-01-20
SAA-303 updating the scheduled activites view on the back of start and end dates being added to the schedule.

matt-crowson-moj:
**DO NOT MERGE WITHOUT DROPPING DEV DB FIRST**

### DIFF
--- a/src/main/resources/migration/common/V1__baseline.sql
+++ b/src/main/resources/migration/common/V1__baseline.sql
@@ -288,14 +288,16 @@ SELECT si.scheduled_instance_id,
            true
        END suspended
 FROM scheduled_instance si
-         JOIN activity_schedule schedule ON schedule.activity_schedule_id = si.activity_schedule_id
+         JOIN activity_schedule schedule
+              ON schedule.activity_schedule_id = si.activity_schedule_id
+                  AND si.session_date >= schedule.start_date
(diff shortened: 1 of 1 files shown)


src/main/resources/migration/common/V1__baseline.sql
```diff
@@ -288,14 +288,16 @@ SELECT si.scheduled_instance_id,
...
+                  AND si.session_date >= schedule.start_date
+                  AND (schedule.end_date is null OR schedule.end_date <= si.session_date)
          JOIN allocation alloc
               ON alloc.activity_schedule_id = si.activity_schedule_id
                   AND si.session_date >= alloc.start_date
                   AND (alloc.end_date is null or alloc.end_date <= si.session_date)
          JOIN activity act
               ON act.activity_id = schedule.activity_id
-                  AND si.session_date >= act.start_date
                   AND (act.end_date is null OR act.end_date <= si.session_date)
```
matt-crowson-moj, line 301: Still need to look at the activity end date as the schedule end can be null.

timharrison-moj (approved):
LGTM. Yes - I think this is the check needed. I have some other things to address here too - I think even where sessions are cancelled or suspended, we still need to return them and created the scheduled instances - currently we're excluding. However, they need to have attendances generated so that they are paid, so the instances are still needed.  Just FYI notes - probably not the place to put them on this PR! Sorry

PR #164, opened 2023-01-25
SAA-296: Upgrade to Spring Boot 3

gerryallan-moj:


build.gradle.kts
```diff
@@ -17,19 +17,21 @@ configurations {
   testImplementation { exclude(group = "org.junit.vintage") }
 }
 
+repositories {
```
matt-crowson-moj, line 20: Might be worth adding a comment why mavenCentral is needed ...

PR #166, opened 2023-01-27
SAA-320 initial commit for scheduled instance job to use start and end dates of the schedule as well as the activity.

matt-crowson-moj:
Initial changes to include the start and end dates of a schedule when creating instances of a schedule.

Includes adding some extra validation on the entities themselves around dates and also adding missing unit tests for pulling out schedules from activities for scheduling.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Activity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Activity.kt
@@ -92,18 +92,13 @@ data class Activity(
     if (endDate != null) date.between(startDate, endDate) else (date.isEqual(startDate) || date.isAfter(startDate))
 
   fun getSchedulesOnDay(day: LocalDate, includeSuspended: Boolean = true): List<ActivitySchedule> {
-    val byDayOfWeek = this.schedules.filter { schedule ->
-      schedule.slots().any { day.dayOfWeek in it.getDaysOfWeek() }
-    }
+    val byDayOfWeek = this.schedules
+      .filter { it.isActiveOn(day) }
+      .filter { schedule -> schedule.slots().any { day.dayOfWeek in it.getDaysOfWeek() } }
     return if (includeSuspended) {
       byDayOfWeek
     } else {
-      byDayOfWeek.filter {
-        it.suspensions.none { suspension ->
-          !day.isBefore(suspension.suspendedFrom) &&
-            (suspension.suspendedUntil == null || !day.isAfter(suspension.suspendedUntil))
-        }
-      }
+      byDayOfWeek.filter { it.isSuspendedOn(day).not() }
     }
   }
 
@@ -200,7 +195,7 @@ data class Activity(
     return schedules.last()
   }
 
-  fun failIfScheduleDatesClashWithActivityDates(startDate: LocalDate, endDate: LocalDate?) {
+  private fun failIfScheduleDatesClashWithActivityDates(startDate: LocalDate, endDate: LocalDate?) {
     if (startDate.isBefore(this.startDate)) {
       throw IllegalArgumentException("The schedule start date '$startDate' cannot be before the activity start date ${this.startDate}")
     }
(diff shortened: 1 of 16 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Activity.kt
```diff
@@ -92,18 +92,13 @@ data class Activity(
     if (endDate != null) date.between(startDate, endDate) else (date.isEqual(startDate) || date.isAfter(startDate))
 
   fun getSchedulesOnDay(day: LocalDate, includeSuspended: Boolean = true): List<ActivitySchedule> {
-    val byDayOfWeek = this.schedules.filter { schedule ->
-      schedule.slots().any { day.dayOfWeek in it.getDaysOfWeek() }
-    }
+    val byDayOfWeek = this.schedules
+      .filter { it.isActiveOn(day) }
```
matt-crowson-moj, line 96: This is the key bit checking the dates of the schedule are included when returning schedules.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Activity.kt
```diff
@@ -214,7 +209,7 @@ data class Activity(
     }
   }
 
-  fun failIfScheduleWithDescriptionAlreadyPresentOnActivity(description: String) {
+  private fun failIfScheduleWithDescriptionAlreadyPresentOnActivity(description: String) {
     if (schedules.any { it.description.trim().uppercase() == description.trim().uppercase() }) {
       throw IllegalArgumentException("A schedule with the description '$description' already exists.")
```
timharrison-moj, line 214: When I was checking this part - whether something already exists on the day - it seemed to sometimes create duplicates of the same activity. Perhaps something to check locally?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/ActivitySchedule.kt
```diff
@@ -128,6 +131,47 @@ data class ActivitySchedule(
...
+    }
+  }
+
+  private fun failIfMatchingInstanceAlreadyPresent(
+    sessionDate: LocalDate,
+    slot: ActivityScheduleSlot
+  ) {
+    if (instances.any { it.sessionDate == sessionDate && it.startTime == slot.startTime && it.endTime == slot.endTime }) {
+      throw IllegalArgumentException("An instance for date '$sessionDate', start time '${slot.startTime}' and end time '${slot.endTime}' already exists")
+    }
```
timharrison-moj, line 172: Do we want to throw exceptions here? These would be thrown up to the running job and potentially stop it, as it stands?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/ActivityScheduleTest.kt
```diff
@@ -369,4 +373,116 @@ class ActivityScheduleTest {
...
+      description = "description",
+      capacity = 1,
+      startDate = today,
+    ).apply {
+      addSlot(
+        startTime = today.atStartOfDay().toLocalTime(),
+        endTime = today.atStartOfDay().toLocalTime().plusHours(1),
+        setOf(DayOfWeek.MONDAY)
+      )
+    }
```
timharrison-moj, line 476: This might help with the duplication of scheduled instances that I was seeing earlier - before your changes?

timharrison-moj (approved):
LGTM. Couple of questions only.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/ActivitySchedule.kt
matt-crowson-moj: I'd still be keen for the exceptions to stay and the job to handle (runtime) exceptions.  It things are working as they should then it should never happen.

I was kinda hoping to see some exceptions when I ran the job locally but didn't see any.  To be fair I tried this after my changes.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Activity.kt
matt-crowson-moj: Hmm OK.  Obvs it shouldn't but will check.

timharrison-moj (approved):
LGTM

PR #172, opened 2023-02-01
SAA-330 removing query constraint on scheduled instances requiring an allocation when pulling back a list.

matt-crowson-moj:
This is a subtle change the endpoint which returns a list of scheduled instances at a given prison on given dates and time.

If no there were no allocations to a scheduled instance then it would be would be excluded from the results.  This restriction has now been removed.

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/integration/ActivityScheduleInstanceIntegrationTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/integration/ActivityScheduleInstanceIntegrationTest.kt
@@ -4,6 +4,9 @@ import org.assertj.core.api.Assertions.assertThat
 import org.junit.jupiter.api.Test
 import org.springframework.http.MediaType
 import org.springframework.test.context.jdbc.Sql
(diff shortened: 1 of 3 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/integration/ActivityScheduleInstanceIntegrationTest.kt
```diff
@@ -31,61 +34,78 @@ class ActivityScheduleInstanceIntegrationTest : IntegrationTestBase() {
...
+    val endDate = LocalDate.of(2022, 11, 5)
+
+    val scheduledInstances = webTestClient.getScheduledInstancesBy(moorlandPrisonCode, startDate, endDate, TimeSlot.AM)
+
+    assertThat(scheduledInstances).hasSize(1)
+  }
+
+  @Test
+  @Sql("classpath:test_data/seed-activity-id-10.sql")
+  fun `getActivityScheduleInstances - returns no instance when match on time slot`() {
```
matt-crowson-moj, line 80: Making a change to the query had no impact on the integration tests so added this just prove a point.

PR #173, opened 2023-02-01
SAA-256- unlock list - check and correct transforms for courts and visits

timharrison-moj:
Tracking down why locationId and other details were not being populated.


matt-crowson-moj (approved):
LGTM

PR #175, opened 2023-02-03
SAA-352 bring the outbound domain even message inline with the ADR, requires AdditionalInfo object as part of message.

matt-crowson-moj:
The ADR for message formats related to this change can be found in Confluence.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityService.kt
@@ -70,7 +70,7 @@ class ActivityService(
       activityCategory = category,
       activityTier = tier,
       attendanceRequired = request.attendanceRequired,
-      summary = request.summary!!,
+      summary = request.summary,
       description = request.description,
       startDate = request.startDate ?: LocalDate.now(),
       endDate = request.endDate,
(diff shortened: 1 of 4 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityService.kt
```diff
@@ -70,7 +70,7 @@ class ActivityService(
       activityCategory = category,
       activityTier = tier,
       attendanceRequired = request.attendanceRequired,
-      summary = request.summary!!,
+      summary = request.summary,
```
matt-crowson-moj, line 73: Unrelated change.  Noticed when compiling the code.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/EventsService.kt
```diff
@@ -26,40 +26,47 @@ class OutboundEventsService(private val publisher: EventsPublisher) {
...
+        additionalInformation = additionalInformation,
+        description = "A new activity schedule has been created in the activities management service"
+      )
+  };
+
+  abstract fun event(additionalInformation: AdditionalInformation): OutboundHMPPSDomainEvent
 }
 
-// TODO resolve message formats
+interface AdditionalInformation
```
matt-crowson-moj, line 47: Pulled out so we can have different implementations of this depending on the nature of the event.  In some cases will we need more information and sometimes less.

PR #181, opened 2023-02-08
SAA-357. Expose the full pay bands instead of just the ID

sioldham:
Expose the full pay bands over the API instead of just the ID. 

I've added the Pay Bands used by the integration tests into a single file as they need to match those in the DB seed file and, should that seed file change, the corresponding Pay Band expectations need only be changed in one place (as opposed to being scattered across the tests)

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/PrisonPayBand.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/PrisonPayBand.kt
@@ -31,7 +31,7 @@ data class PrisonPayBand(
   @Column(name = "nomis_pay_band", nullable = false)
   val nomisPayBand: Int
 ) {
-  fun toModel(): ModelPrisonPayBand = ModelPrisonPayBand(
+  fun toModel() = ModelPrisonPayBand(
 
     id = prisonPayBandId,
     displaySequence = displaySequence,
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/PrisonPayBand.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/PrisonPayBand.kt
@@ -6,6 +6,7 @@ import jakarta.persistence.GeneratedValue
 import jakarta.persistence.GenerationType
 import jakarta.persistence.Id
 import jakarta.persistence.Table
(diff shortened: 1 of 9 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/integration/testdata/IntegrationTestPayBands.kt
```diff
@@ -0,0 +1,7 @@
+package uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.integration.testdata
+
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.model.PrisonPayBand
+
+val testPayBand1 = PrisonPayBand(1, 1, "Pay band 1 (lowest)", "Pay band 1 (lowest)", 1, "PVI")
```
matt-crowson-moj, line 5: Given these are prison specific it might pay to include the prison name in these as well?  PVI = Pentonville for example.

Or provide via function call perhaps?

```
fun payBandOne(prisonCode: String) = 
  PrisonPayBand(1, 1, "Pay band 1 (lowest)", "Pay band 1 (lowest)", 1, prisonCode)

```

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/PrisonPayBand.kt
```diff
@@ -29,4 +30,14 @@ data class PrisonPayBand(
 
   @Column(name = "nomis_pay_band", nullable = false)
   val nomisPayBand: Int
-)
+) {
+  fun toModel(): ModelPrisonPayBand = ModelPrisonPayBand(
```
matt-crowson-moj, line 34: Minor point.  Don't need to specify return type.  Meh.

matt-crowson-moj (approved):
LGTM with a couple of minor comments.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/integration/testdata/IntegrationTestPayBands.kt
sioldham: I've just renamed the variables for the time being because, looking at the seed data, it's not always possible to accurately create the alias and description solely from the pay band id and the prison code without addition logic.

For example. The alias/description for PVI Band Three is "Pay band 3" but for MDI it is "Pay band 3 (highest)"

matt-crowson-moj (approved):
LGTM

PR #187, opened 2023-02-15
SAA-336: Add scheduleId to the Allocation response model

smcveigh941:
At the moment, the only way to know the schedule which an allocation is referring to is by the `scheduleDescription`. This is to add the schedule identifier to the response for more reliable mapping.


matt-crowson-moj (approved):
LGTM

PR #188, opened 2023-02-15
SAA-372: Make riskLevel a mandatory field on an activity

smcveigh941:
- `riskLevel` is mandatory
- Updated seeded data to reflect
- Frontend already supports this
- Syscon need not be informed


matt-crowson-moj (approved):
LGTM

PR #192, opened 2023-02-22
SAA-348: Date and time should not be in the past

iterative-it:
Prevents the creation of appointments where the start date and time are in the past. Previously the code would allow same day appointments that started earlier than the current time.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/model/request/AppointmentCreateRequest.kt
```diff
@@ -97,6 +98,9 @@ data class AppointmentCreateRequest(
   @AssertTrue(message = "Internal location id must be supplied if in cell = false")
   private fun isInternalLocationId() = inCell || internalLocationId != null
 
+  @AssertTrue(message = "Start time must be in the future")
+  private fun isStartTime() = startDate == null || startTime == null || LocalDateTime.of(startDate, startTime) > LocalDateTime.now()
```
matt-crowson-moj, line 102: Personally (if you haven't already) I would also consider enforcing this type of logic on the Domain entity.

matt-crowson-moj (approved):
LGTM

PR #196, opened 2023-02-27
SAA-401 adding new seeded activity category to support allocating prisoners to unemployed activities e.g. retired.

matt-crowson-moj:
**REQUIRES DB TO BE DROPPED PRIOR TO MERGE**

This will display as follows on activity creation in the frontend:

image


timharrison-moj (approved):
Also LGTM.

PR #202, opened 2023-03-03
SAA-411 dependency updates.  Linting has changed hence the large number formatting changes in this commit.

matt-crowson-moj:
As a result of moving to a more update plugin version we have inherited a linting formatting change hence the large number of files affected by it in this PR.  It's really all extra commas, there are no actual functional changes.

Also see the Kotlin docs  [here](https://kotlinlang.org/docs/coding-conventions.html#trailing-commas) on trailing commas.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,5 +1,5 @@
 plugins {
-  id("uk.gov.justice.hmpps.gradle-spring-boot") version "5.1.0"
+  id("uk.gov.justice.hmpps.gradle-spring-boot") version "5.1.1-beta-2"
   kotlin("plugin.spring") version "1.8.10"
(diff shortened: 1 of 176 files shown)


build.gradle.kts
```diff
@@ -91,3 +91,11 @@ tasks.named<JacocoReport>("jacocoTestReport") {
     html.required.set(true)
   }
 }
+
+ktlint {
+  filter {
+    // exclude open api generated types
```
matt-crowson-moj, line 97: The code comment says as well.  Essentially don't control the format of the generated types erg we want to ignore them.

matt-crowson-moj: I did wonder about repackaging these into a folder structure prefixed with `/generated/...`

PR #204, opened 2023-03-03
SAA-411 dependency updates. Bumping DPS spring boot gradle version for security patches.

matt-crowson-moj:


yasinmustafa (approved):
LGTM

PR #205, opened 2023-03-06
SAA-346: Add flag to the allocation object indicating whether the allocation is for employment or not

smcveigh941:
Includes fix for broken integration test


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Allocation.kt
```diff
@@ -50,6 +50,8 @@ data class Allocation(
 ) {
   fun isActive(date: LocalDate) = date.between(startDate, endDate)
 
+  fun isUnemployment() = activitySchedule.activity.activityCategory.code == "SAA_NOT_IN_WORK"
```
matt-crowson-moj, line 53: It kinda feels to me that our categories could just be enums.  I don't see the value (anymore) in these being JPA entities.

That as an aside I would not chain calls this deep.

I'd rather we had something like `activity.isUnemployed()` which would in turn delegate to the category on the activity.

Allocation should not really know anything about the actual category code values.

matt-crowson-moj (approved):
LGTM

PR #206, opened 2023-03-07
SAA-390 - Add allocation data source toggle

sioldham:
The purpose of this change is to make configurable the data source for appointments data.

There are a couple of points that I'd like to draw your attention to:

- I've added the `getTimeRangeForPrisonAndTimeSlot` method to the `PrisonRegimeService`. It may look odd at first glance but we noticed that there are gaps between the AM/PM/ED time slots which meant that, when using SQL (as opposed to the Prison API), certain appointments would fall in the gaps and not be retrieved. Please do double check the logic though.
- It occurred to me that the `TransformFunctions` file was getting a bit long so I put my new mappers in a new `map` package. Happy to move it into `TransformFunctions` though if that's preferred.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/AppointmentInstanceService.kt
```diff
@@ -0,0 +1,102 @@
...
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.repository.AppointmentInstanceRepository
+import java.time.LocalDate
+import java.time.LocalTime
+
+/**
+ * Fetches appointment data from either the Prison API or the Appointment Instance Repository, depending on the rollout
+ * status of the prison.
+ */
+@Service
+class AppointmentInstanceService(
```
matt-crowson-moj, line 26: We have a mixture of blocking and not blocking calls and they shouldn't really be mixed like this.

Probably easier to have a chat about.

sioldham: I've updated the service to block on the Prison API. The methods now return simple lists and the `ScheduledEventService` wraps them in monos which should suffice for now, pending a separate refactor ticket.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/model/AppointmentsDataSource.kt
```diff
@@ -0,0 +1,7 @@
+package uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.model
+
+enum class AppointmentsDataSource {
```
matt-crowson-moj, line 3: Minor point. I'd probably say this belongs within the entities package.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/AppointmentInstanceService.kt
```diff
@@ -0,0 +1,109 @@
...
+          .associateBy { it.prisonerNumber }
+      } else {
+        emptyMap()
+      }
+
+      prisonersWithAppointments.toPrisonerSchedule(prisonerMap, locationMap)
+    }
+  }
+
+  /**
```
matt-crowson-moj, line 95: Feels a little bit heavy on the documentation for what is essentially a private method.  Isn't the method name enough?

Also I would consider using an extension function which is a bit more Kotlin idiomatic e.g.

```
  private fun RolloutPrison.shouldUsePrisonApi() = 
    !this.active || this.appointmentsDataSource == AppointmentsDataSource.PRISON_API
```

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/AppointmentInstanceService.kt
```diff
@@ -0,0 +1,109 @@
...
+      val prisonersWithAppointments = appointmentInstanceRepository.findByPrisonCodeAndPrisonerNumberAndDateAndTime(
+        prisonCode,
+        prisonerNumbers,
+        date,
+        earliestStartTime,
+        latestStartTime,
+      )
+      log.info("Fetched [${prisonersWithAppointments.size}] prisoners with appointments from the Appointment Instance Repository")
+
+      val prisonerMap = if (prisonersWithAppointments.isNotEmpty()) {
```
matt-crowson-moj, line 84: It will only follow through on the other calls if the appointments aren't empty so this should suffice:

```
      val prisonerMap = prisonersWithAppointments
        .map { it.prisonerNumber }
        .let { prisonerSearchApiClient.findByPrisonerNumbers(it).block()!! }
        .associateBy { it.prisonerNumber }

```

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/AppointmentInstanceService.kt
```diff
@@ -0,0 +1,109 @@
...
+
+  fun getPrisonerSchedules(
+    prisonCode: String,
+    prisonerNumbers: Set<String>,
+    rolloutPrison: RolloutPrison,
+    date: LocalDate,
+    timeSlot: TimeSlot?,
+  ): List<PrisonerSchedule> {
+    log.info(
+      "Fetching prisoner schedules for Rollout Prison [$rolloutPrison], Prison Code [$prisonCode], " +
```
matt-crowson-moj, line 60: Logging feels a little overkill.  Could get away with one log message in either branch of the `if / else`.

A word of caution, be very careful logging an entire object (the rollout prison in this case).  There is the potential to log PII when doing things like this particularly with data classes.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/AppointmentInstanceService.kt
sioldham: Haha. You can't win with documentation; some like it, some don't. Will remove

sioldham: Good shout on the idiomatic Kotlin. Thanks. Will do

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/PrisonRegimeService.kt
```diff
@@ -45,6 +48,22 @@ class PrisonRegimeService(
   fun getPrisonRegimeByPrisonCode(code: String) = transform(
     prisonRegimeRepository.findByPrisonCode(code) ?: throw EntityNotFoundException(code),
   )
+
+  fun getTimeRangeForPrisonAndTimeSlot(prisonCode: String, timeSlot: TimeSlot): LocalTimeRange {
```
matt-crowson-moj, line 52: Suggestied change, not essential by any means but again a bit more Kotliny perhaps:

```
  fun getTimeRangeForPrisonAndTimeSlot(prisonCode: String, timeSlot: TimeSlot) =
    getPrisonRegimeByPrisonCode(prisonCode)
      .let { pr ->
        val (start, end) = when(timeSlot) {
          TimeSlot.AM -> LocalTime.MIDNIGHT to pr.pmStart
          TimeSlot.PM -> pr.pmStart to pr.edStart
          TimeSlot.ED -> pr.edStart to LocalTime.of(23, 59)
        }

        LocalTimeRange(start, end)
      }
```

Instead of using the pair here you could also just return LocalTimeRange in the when statement.  Meh.

matt-crowson-moj, line 52: This is one of those methods where some documentation would be of assistance perhaps given it needs explaining in the PR?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/map/AppointmentInstanceMapper.kt
```diff
@@ -0,0 +1,63 @@
...
+private const val EVENT_TYPE = "APP"
+private const val EVENT_TYPE_DESC = "Appointment"
+
+/**
+ * Maps a List<AppointmentInstance> to a List<PrisonerSchedule>
+ *
+ *   @param prisonerLookup Map of prisonerNumber -> Prisoner to facilitate data lookup
+ *   @param locationLookup Map of locationId -> Location to facilitate data lookup
+ */
+fun List<AppointmentInstance>.toPrisonerSchedule(
```
matt-crowson-moj, line 23: Yeah I am not sure what is best with these.  This adds another way of doing it.  I am not a total fan but perhaps just put these directly within the AppointmentInstance entity?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/PrisonRegimeService.kt
sioldham: Nice. Thanks. Will do

sioldham: Good point .Yep. Will add

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ScheduledEventService.kt
```diff
@@ -45,7 +47,7 @@ class ScheduledEventService(
     }
     ?.let { prisonerDetail ->
       val bookingId = prisonerDetail.bookingId!!.toLong()
-      val prisonRolledOut = rolloutPrisonRepository.findByCode(prisonCode)?.active ?: false
+      val prisonRolledOut = rolloutPrisonRepository.findByCode(prisonCode)!!
```
matt-crowson-moj, line 50: Hmm this is definitely a change in behaviour.  Not saying it isn't warranted but not sure the `!!` is the best option here.

We would probably be better off with an IllegalArgumentException or EntityNotFoundException exception and meaningful message.  This would just throw an NPE with no real context.

So I tend to use the double exclamation when working Java (platform types) and even then I don't really like it.

sioldham: That's fair. Yes. I'll change it to throw a meaningful exception (I think I've done it in another place too)

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/model/LocalTimeRange.kt
```diff
@@ -0,0 +1,8 @@
+package uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.model
+
+import java.time.LocalTime
+
+data class LocalTimeRange(
```
matt-crowson-moj, line 5: This probably doesn't belong in this package either really.  As far as I can tell it is not used here or by any models?

We do have a common package.  Perhaps it would fit there?

src/main/resources/migration/data/R__1_1__rollout_prisons.sql
```diff
@@ -1,5 +1,5 @@
-insert into rollout_prison (rollout_prison_id, code, description, active, rollout_date) select 1, 'LEI', 'HMP Leeds', true, '2022-12-19' where not exists (select 1 from rollout_prison where rollout_prison_id = 1);
-insert into rollout_prison (rollout_prison_id, code, description, active, rollout_date) select 2, 'MDI', 'HMP Moorland', true, '2022-11-20' where not exists (select 1 from rollout_prison where rollout_prison_id = 2);
-insert into rollout_prison (rollout_prison_id, code, description, active, rollout_date) select 3, 'PVI', 'HMP Pentonville', true, '2022-10-21' where not exists (select 1 from rollout_prison where rollout_prison_id = 3);
+insert into rollout_prison (rollout_prison_id, code, description, active, rollout_date, appointments_data_source) select 1, 'LEI', 'HMP Leeds', true, '2022-12-19', 'PRISON_API' where not exists (select 1 from rollout_prison where rollout_prison_id = 1);
```
matt-crowson-moj, line 1: I don't know the answer to this, just asking.  Don't we want any of the prison in dev to use the activities service?

If the answer is yes in the activities space we tend to favour MDI (Moorland).

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/AppointmentInstanceServiceTest.kt
```diff
@@ -0,0 +1,248 @@
...
+  private val appointmentInstanceRepository: AppointmentInstanceRepository = mock()
+  private val appointmentInstanceService = AppointmentInstanceService(
+    prisonApiClient,
+    prisonerSearchApiClient,
+    locationService,
+    appointmentInstanceRepository,
+    prisonRegimeService,
+  )
+
+  @Nested
```
matt-crowson-moj, line 50: Good shout on using **Nested** by the way.  Something we haven't been using but probably should consider using.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/map/AppointmentInstanceMapper.kt
sioldham: Moved into TransformFunctions for now

src/main/resources/migration/data/R__1_1__rollout_prisons.sql
sioldham: Agreed to use Moorland

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ScheduledEventService.kt
```diff
@@ -115,7 +118,7 @@ class ScheduledEventService(
     timeSlot: TimeSlot? = null,
   ): PrisonerScheduledEvents? {
     val eventPriorities = prisonRegimeService.getEventPrioritiesForPrison(prisonCode)
-    val prisonRolledOut = rolloutPrisonRepository.findByCode(prisonCode)?.active ?: false
+    val prisonRolledOut = rolloutPrisonRepository.findByCode(prisonCode)!!
```
matt-crowson-moj, line 121: I missed this one.  This is also changing behaviour?  NPE not very friendly.

matt-crowson-moj: Yeah sorry I missed it last time.  I think sort this though and we are done and we can possibly merge it later today with my changes also.

matt-crowson-moj (approved):
LGTM - don't merge though yet as need to coordinate with Syscon for DB changes.

PR #207, opened 2023-03-08
SAA-413: Appointment details endpoint

iterative-it:
The reasoning behind this change is in the [ticket](https://dsdmoj.atlassian.net/browse/SAA-413). At a high level it's due to the source of the connected display data e.g. categories, locations and prisoners potentially changing as domains are extracted from NOMIS. It also supports caching at the API layer and other optimisations.

There will be an associated UI change once this PR is approved.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/client/prisonapi/api/PrisonApiClient.kt
```diff
@@ -243,4 +244,12 @@ class PrisonApiClient(private val prisonApiWebClient: WebClient) {
...
   fun getEducationLevel(educationLevelCode: String): Mono<ReferenceCode> =
     getReferenceCode("EDU_LEVEL", educationLevelCode)
+
+  fun getUserDetailsList(usernames: List<String>): Mono<List<UserDetail>> {
+    return prisonApiWebClient.post()
+      .uri("/api/users/list")
+      .bodyValue(usernames)
+      .retrieve()
+      .bodyToMono(typeReference<List<UserDetail>>())
+  }
```
timharrison-moj, line 254: Does this need to be a Mono<List>?  Empty list if no items?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/AppointmentOccurrence.kt
```diff
@@ -77,6 +84,29 @@ data class AppointmentOccurrence(
...
+      if (inCell) null else locationMap.getOrDefault(internalLocationId, null).toAppointmentLocationSummary(internalLocationId!!, prisonCode),
+      inCell,
+      startDate,
+      startTime,
+      endTime,
+      comment ?: appointmentComment,
+      isEdited = false,
+      isCancelled = false,
+      updated = updated,
+      updatedBy = if (updatedBy == null) null else userMap.getOrDefault(updatedBy, null).toSummary(updatedBy!!),
```
matt-crowson-moj, line 100: A bit more idiomatic. Meh.
```
updatedBy?.let { userMap.getOrDefault(updatedBy, null).toSummary(updatedBy!!) }
```

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/AppointmentOccurrence.kt
```diff
@@ -77,6 +84,29 @@ data class AppointmentOccurrence(
     allocations = allocations.toModel(),
     instances = instances.toModel(),
   )
+
+  fun toSummary(prisonCode: String, locationMap: Map<Long, Location>, userMap: Map<String, UserDetail>, appointmentComment: String) =
+    AppointmentOccurrenceSummary(
+      appointmentOccurrenceId,
+      if (inCell) null else locationMap.getOrDefault(internalLocationId, null).toAppointmentLocationSummary(internalLocationId!!, prisonCode),
+      inCell,
```
timharrison-moj, line 92: I was going to suggest ways to avoid the if / else with ?.let or ?.also - but when I tried them, I think yours is clearer, so not commenting on it now.

matt-crowson-moj (dismissed):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/AppointmentOccurrence.kt
matt-crowson-moj: It's a bit awkward with booleans.

I did play with this a little wasn't convinced it was worth it e.g.

```
fun <T> Boolean.ifTrueOrNull(block: () -> T) = if (this) block() else null

inCell.ifTrueOrNull { locationMap.getOrDefault(internalLocationId, null).toAppointmentLocationSummary(internalLocationId!!, prisonCode) },

```

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/AppointmentOccurrence.kt
iterative-it: Like it! Changed

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/client/prisonapi/model/UserDetail.kt
```diff
@@ -0,0 +1,73 @@
...
+ * @param expiryDate Date the user account has expired
+ * @param lockedFlag The User account is locked
+ * @param expiredFlag Indicates the user account has expired
+ */
+data class UserDetail(
+  @Schema(example = "231232", required = true, description = "Staff Id")
+  @get:JsonProperty("staffId", required = true) val staffId: Long,
+
+  @Schema(example = "DEMO_USER1", required = true, description = "Username")
+  @get:JsonProperty("username", required = true) val username: String,
```
timharrison-moj, line 26: Do you need the `@get:JsonProperty`  annotations?  They look like standard mappings of names and types that would be handled by default behaviour.

iterative-it: This class was generated by OpenAPI Generator. I can remove them?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/client/prisonapi/api/PrisonApiClient.kt
iterative-it: Changed the response type

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/client/prisonapi/model/PrisonApiTypesTest.kt
```diff
@@ -24,4 +24,11 @@ class PrisonApiTypesTest {
 
     assertThat(classificationField.returnType.isMarkedNullable).isTrue
   }
+
+  @Test
+  fun `lockDate field on generated User Detail DTO type should be nullable`() {
+    val field = UserDetail::class.declaredMembers.first { it.name == "lockDate" }
+
+    assertThat(field.returnType.isMarkedNullable).isTrue
+  }
```
timharrison-moj, line 33: I wonder if these types of tests are starting to test the class generators and framework, rather than our code? Might save ourselves some time by limiting tests to just our code?

iterative-it: Ah this one is because the API documentation is incorrect. That property is nullable based on the data coming back in dev at least. I had to manually change the generated model and this test is to ensure that if it’s regenerated, it’s set back to nullable again

timharrison-moj (approved):
LGTM.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/client/prisonapi/model/UserDetail.kt
timharrison-moj: It ok - leave if generated - it'll only put them back again :-(

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/client/prisonapi/model/PrisonApiTypesTest.kt
matt-crowson-moj: Yes we need these tests @timharrison-moj, the overhead is tiny.

PR #210, opened 2023-03-10
SAA-417 ensure both attendance creation and capacities take deallocated offenders into account.

matt-crowson-moj:
Changes to attendance record creation and capacity calculations on the back of introducing a status field to allocations.

Having a status field on allocations removes the need to check the dates on them.

The status field will be managed going forwards to determine an allocations status.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Allocation.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Allocation.kt
@@ -11,7 +11,6 @@ import jakarta.persistence.JoinColumn
 import jakarta.persistence.ManyToOne
 import jakarta.persistence.OneToOne
 import jakarta.persistence.Table
-import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.common.between
 import java.time.LocalDate
 import java.time.LocalDateTime
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.model.Allocation as ModelAllocation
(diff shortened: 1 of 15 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Allocation.kt
```diff
@@ -66,7 +66,7 @@ data class Allocation(
   var prisonerStatus: PrisonerStatus = PrisonerStatus.ACTIVE
     private set
 
-  fun isActive(date: LocalDate) = date.between(startDate, endDate)
+  fun isActive(date: LocalDate) = date.between(startDate, endDate) && status(PrisonerStatus.ENDED).not()
```
matt-crowson-moj, line 69: @timharrison-moj 

Can you just confirm this is correct.  Should this be checking if only ACTIVE?  This will include SUSPENSION as it stands.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Allocation.kt
```diff
@@ -66,7 +66,7 @@ data class Allocation(
...
```
timharrison-moj, line 70: Might need an `isSuspended()` soon?   Active ACTIVE and Active SUSPENDED might need to be separated out at some point?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/CapacityService.kt
```diff
@@ -58,6 +58,6 @@ class CapacityService(
   }
 
   private fun ActivitySchedule.getAllocationsForToday(): List<Allocation> {
-    return this.getAllocationsOnDate(LocalDate.now())
+    return this.getActiveAllocationsOnDate(LocalDate.now())
   }
```
timharrison-moj, line 62: Question whether SUSPENDED is "active" or not? For capacity, I think it's probably right to be included here - even though suspended, they are still allocated .. strictly speaking.

PR #214, opened 2023-03-14
SAA-420: Cancel schedule instance

leejacobson-moj:
## Overview

Adds cancel scheduled instance endpoint, `/scheduled-instances/{instanceId}/cancel`

Calling this endpoint will cancel the schedule instance, and update the status and pay of any associated attendance records.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Attendance.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Attendance.kt
@@ -58,6 +58,25 @@ data class Attendance(
   override fun toString(): String {
     return this::class.simpleName + "(attendanceId = $attendanceId )"
   }
+
+  fun cancel(reason: AttendanceReason, payIncentiveCode: String?) {
+    status = AttendanceStatus.COMPLETED
+    payAmount = if (payIncentiveCode != null) getPay(payIncentiveCode)?.rate ?: 0 else 0
+    attendanceReason = reason
+    comment = scheduledInstance.cancelledReason
+    recordedTime = scheduledInstance.cancelledTime
+    recordedBy = scheduledInstance.cancelledBy
+  }
+
+  private fun getPay(incentiveCode: String): ActivityPay? {
+    val currentAllocation = scheduledInstance.activitySchedule.allocations()
+      .filter { it.isAllocated() }
+      .find { it.prisonerNumber == prisonerNumber }
+
+    return scheduledInstance.activitySchedule.activity.activityPay()
+      .filter { it.payBand.prisonPayBandId == currentAllocation?.payBand?.prisonPayBandId }
+      .find { it.incentiveNomisCode == incentiveCode }
+  }
 }
 
 enum class AttendanceStatus {
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/model/request/ScheduleInstanceCancelRequest.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/model/request/ScheduleInstanceCancelRequest.kt
@@ -1,0 +1,20 @@
+package uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.model.request
+
+import io.swagger.v3.oas.annotations.media.Schema
+import jakarta.validation.constraints.NotEmpty
+import jakarta.validation.constraints.Size
+
+@Schema(description = "Request object for cancelling a schedule instance")
+data class ScheduleInstanceCancelRequest(
+  @field:NotEmpty(message = "Cancellation reason must be supplied")
+  @field:Size(max = 100, message = "Cancellation reason must not exceed {max} characters")
+  @Schema(description = "The reason for cancelling the schedule instance", example = "No tutor available")
+  val reason: String,
+
+  @field:NotEmpty(message = "Username must be supplied")
+  @Schema(description = "The username of the user cancelling the schedule instance", example = "RJ56DDE")
+  val username: String,
+
+  @Schema(description = "A field for any additional comments", example = "Resume tomorrow")
+  val comment: String?,
+)
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ScheduledInstanceService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ScheduledInstanceService.kt
@@ -1,15 +1,22 @@
 package uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.service
 
 import org.springframework.stereotype.Service
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.client.prisonersearchapi.api.PrisonerSearchApiClient
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.common.LocalDateRange
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.common.TimeSlot
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.entity.toModel
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.model.ActivityScheduleInstance
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.model.request.ScheduleInstanceCancelRequest
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.repository.AttendanceReasonRepository
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.repository.ScheduledInstanceRepository
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.repository.findOrThrowNotFound
 
 @Service
-class ScheduledInstanceService(private val repository: ScheduledInstanceRepository) {
+class ScheduledInstanceService(
+  private val repository: ScheduledInstanceRepository,
+  private val attendanceReasonRepository: AttendanceReasonRepository,
+  private val prisonerSearchApiClient: PrisonerSearchApiClient,
+) {
   fun getActivityScheduleInstanceById(id: Long): ActivityScheduleInstance =
     repository.findOrThrowNotFound(id).toModel()
 
@@ -35,5 +42,26 @@ class ScheduledInstanceService(private val repository: ScheduledInstanceReposito
     val scheduledInstance = repository.findOrThrowNotFound(id)
     scheduledInstance.uncancel()
     repository.save(scheduledInstance)
+  }
+
+  fun cancelScheduledInstance(instanceId: Long, scheduleInstanceCancelRequest: ScheduleInstanceCancelRequest) {
+    val scheduledInstance = repository.findOrThrowNotFound(instanceId)
+
+    scheduledInstance.cancelSession(
+      reason = scheduleInstanceCancelRequest.reason,
+      by = scheduleInstanceCancelRequest.username,
+      cancelComment = scheduleInstanceCancelRequest.comment,
+    ) { attendanceList ->
+      val attendanceReason = attendanceReasonRepository.findByCode("CANC")
+      val prisonerNumbers = scheduledInstance.attendances.map { it.prisonerNumber }
+      val prisoners = prisonerSearchApiClient.findByPrisonerNumbers(prisonerNumbers).block()?.associateBy { it.prisonerNumber }
+
+      attendanceList.forEach {
+        val payIncentiveCode = prisoners?.get(it.prisonerNumber)?.currentIncentive?.level?.code
+        it.cancel(attendanceReason, payIncentiveCode)
+      }
+    }
+
+    repository.saveAndFlush(scheduledInstance)
   }
 }
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/AttendanceTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/AttendanceTest.kt
@@ -3,9 +3,16 @@ package uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.entity
 import org.assertj.core.api.Assertions.assertThat
 import org.junit.jupiter.api.Test
 import org.mockito.kotlin.mock
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.helpers.activityEntity
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.helpers.attendanceReasons
 import java.time.LocalDateTime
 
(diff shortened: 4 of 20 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/model/request/ScheduleInstanceCancelRequest.kt
```diff
@@ -0,0 +1,20 @@
+package uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.model.request
+
+import io.swagger.v3.oas.annotations.media.Schema
+import jakarta.validation.constraints.NotNull
+import jakarta.validation.constraints.Size
+
+@Schema(description = "Request object for cancelling a schedule instance")
+data class ScheduleInstanceCancelRequest(
+  @field:NotNull(message = "Cancellation reason must be supplied")
```
matt-crowson-moj, line 9: Suggestion.

Favour **NotEmpty** over **NotNull** if a string is required.  Not null allows empty strings and/or whitespace.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ScheduledInstanceService.kt
```diff
@@ -30,4 +37,24 @@ class ScheduledInstanceService(private val repository: ScheduledInstanceReposito
...
+
+  fun cancelScheduledInstance(instanceId: Long, scheduleInstanceCancelRequest: ScheduleInstanceCancelRequest) {
+    val scheduledInstance = repository.findOrThrowNotFound(instanceId)
+
+    if (scheduledInstance.cancelled) throw IllegalArgumentException("The schedule instance has already been cancelled")
+
+    val today = LocalDateTime.now().withNano(0)
+    if (scheduledInstance.sessionDate < today.toLocalDate()) throw IllegalArgumentException("The schedule instance has ended")
+
+    scheduledInstance.cancelled = true
```
matt-crowson-moj, line 49: Suggestion:

I would add a cancel function to the scheduled instance entity i.e. push it down into the Domain.

So you end up with something like.  Really the attributes on the instance for cancellation should not be publicly settable.

```
scheduledInstance.cancel(
  cancelledReason = scheduleInstanceCancelRequest.reason
  cancelledTime = today
  cancelledBy = scheduleInstanceCancelRequest.username
  comment = scheduleInstanceCancelRequest.comment
)
```

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/AttendancesService.kt
```diff
@@ -69,6 +74,36 @@ class AttendancesService(
...
+      it.apply {
+        it.status = AttendanceStatus.CANCELLED
+        it.attendanceReason = attendanceReason
+        it.comment = scheduledInstance.comment
+        it.recordedTime = scheduledInstance.cancelledTime
+        it.recordedBy = scheduledInstance.cancelledBy
+        it.payAmount = payAmount?.rate ?: 0
+      }
+    }
+    attendanceRepository.saveAll(updatedAttendances)
```
matt-crowson-moj, line 104: I don't think this is needed. By saving the instance it will propagate.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ScheduledInstanceService.kt
```diff
@@ -30,4 +37,24 @@ class ScheduledInstanceService(private val repository: ScheduledInstanceReposito
...
+    scheduledInstance.cancelledReason = scheduleInstanceCancelRequest.reason
+    scheduledInstance.cancelledTime = today
+    scheduledInstance.cancelledBy = scheduleInstanceCancelRequest.username
+    scheduledInstance.comment = scheduleInstanceCancelRequest.comment
+    repository.save(scheduledInstance)
```
matt-crowson-moj, line 54: Change to use saveAndFlush() because we will probably be raising an event off the back of this.  This says save it now and flush it the db.  Save can mean a slight delay.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ScheduledInstanceService.kt
```diff
@@ -30,4 +38,31 @@ class ScheduledInstanceService(private val repository: ScheduledInstanceReposito
       activities
     }
   }
+
+  fun cancelScheduledInstance(instanceId: Long, scheduleInstanceCancelRequest: ScheduleInstanceCancelRequest) {
+    val scheduledInstance = repository.findOrThrowNotFound(instanceId)
+
+    if (scheduledInstance.cancelled) throw IllegalArgumentException("The schedule instance has already been cancelled")
```
matt-crowson-moj, line 45: Suggestion.

This exception for checking if already cancelled can probably be pushed into the cancel function itself.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ScheduledInstanceServiceTest.kt
```diff
@@ -65,4 +76,67 @@ class ScheduledInstanceServiceTest {
     result = service.getActivityScheduleInstancesByDateRange(prisonCode, dateRange, TimeSlot.ED)
     assertThat(result).isEmpty()
   }
+
+  @Test
+  fun `cancelScheduledInstance - success`() {
```
matt-crowson-moj, line 81: Suggestion.

It would be worth having coverage where we have multiple attendance records to make sure it cancels them all.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ScheduledInstanceService.kt
```diff
@@ -30,4 +38,31 @@ class ScheduledInstanceService(private val repository: ScheduledInstanceReposito
...
+
+    val today = LocalDateTime.now().withNano(0)
+    if (scheduledInstance.sessionDate < today.toLocalDate()) throw IllegalArgumentException("The schedule instance has ended")
```
matt-crowson-moj, line 48: I am still not sure I fully understand the need for looking back in time.

If the session has passed then cancelling would not make any sense?  May be I am missing something.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ScheduledInstanceServiceTest.kt
leejacobson-moj: Done

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ScheduledInstanceService.kt
leejacobson-moj: I think that's the point? If a session has passed then it wouldn't make sense for a user to cancel it. This check simply enforces that business rule at the API layer.

The check is part of the ticket ACs, so it might be worth us discussing this with Tim if you're still unsure?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/ScheduledInstanceTest.kt
```diff
@@ -38,4 +39,26 @@ class ScheduledInstanceTest {
   fun `instance session is not running on date when cancelled`() {
     assertThat(instance.copy(cancelled = true, sessionDate = LocalDate.MIN).isRunningOn(LocalDate.MIN)).isFalse
   }
+
+  @Test
+  fun `can cancel scheduled instance`() {
```
matt-crowson-moj, line 44: Could do with unit test for one that is already cancelled.

```
instance.cancel (...)

assertThatThrownBy { instance.cancel(...) }
  .isInstanceOf(IllegalArgumentException::class.java)
  .hasMessage("Expected message here ...'")
```

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Attendance.kt
```diff
@@ -49,6 +49,24 @@ data class Attendance(
   override fun toString(): String {
     return this::class.simpleName + "(attendanceId = $attendanceId )"
   }
+
+  fun cancel(reason: AttendanceReason, payIncentiveCode: String?) {
```
matt-crowson-moj, line 53: You could argue this should also fail if already cancelled?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/AttendanceTest.kt
```diff
@@ -0,0 +1,28 @@
...
+  private val today = LocalDateTime.now()
+
+  @Test
+  fun `can cancel attendance`() {
+    val attendanceReason = AttendanceReason(1, "NREQ", "Not Required")
+    val canceledInstance = instance.copy(cancelledBy = "USER1", cancelledTime = today)
+    val attendanceWithCanceledInstance = attendance.copy(scheduledInstance = canceledInstance)
+
+    attendanceWithCanceledInstance.cancel(attendanceReason, "BAS")
+    Assertions.assertThat(attendanceWithCanceledInstance.status).isEqualTo(AttendanceStatus.CANCELLED)
```
matt-crowson-moj, line 22: Remove `Assertions.` from the assertion statements.  This happens when you do a copy and paste in intellij ;)

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/ScheduledInstanceTest.kt
leejacobson-moj: done

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Attendance.kt
leejacobson-moj: This has since been updated to reflect the removal of the cancel status. Cancelled attendances now have a status of "COMPLETE". A attendance marked as "COMPLETE" can be overwritten as per ACs in SAA-420.

It could instead check the attendance reason, but will leave this for now as I'm unsure of the exact logic here. For example, if a scheduled instance already has one cancelled attendance, and the schedule is then cancelled, should that cancelled attendance now be updated with a new cancellation time / by or should it skip the record?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Attendance.kt
```diff
@@ -49,6 +49,25 @@ data class Attendance(
...
+    status = AttendanceStatus.COMPLETED
```
matt-crowson-moj, line 54: Thought/suggestion:

Should this be checking the instance has been cancelled prior to cancelling the attendance?

We are using the (nullable) values from the owning instance.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ScheduledInstanceService.kt
timharrison-moj: We can't cancel sessions that were scheduled for before today. Only on-the-day or in the future.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Attendance.kt
timharrison-moj: I think it should update the cancelled time/user if someone cancels and there were existing cancelled attendances. I don't think there ever will be though. I think the ACs have the rule - you can only cancel a scheduled instance which is not currently cancelled?

src/test/resources/test_data/seed-activity-id-13.sql
```diff
@@ -42,6 +42,3 @@ values (1, now(), '10:00:00', '11:00:00', true, now(), 'Old Canceller', 'Nobody
 
 insert into attendance(attendance_id, scheduled_instance_id, prisoner_number, attendance_reason_id, comment, posted, recorded_time, recorded_by, status, pay_amount, bonus_amount, pieces)
 values (1, 1, 'A22222A', 1, 'Attendance Comment', null, now(), 'Old Recorder', 'SCHEDULED', 150, null, null);
```
timharrison-moj, line 44: Should the status be "WAIT" not "SCHEDULED"?

src/test/resources/test_data/seed-activity-id-16.sql
```diff
@@ -0,0 +1,53 @@
...
+
+insert into scheduled_instance(scheduled_instance_id, activity_schedule_id, session_date, start_time, end_time, cancelled, cancelled_time, cancelled_by, cancelled_reason, comment)
+values (2, 2, '2022-10-10', '14:00:00', '15:00:00', false, null, null, null, null);
+
+insert into scheduled_instance(scheduled_instance_id, activity_schedule_id, session_date, start_time, end_time, cancelled, cancelled_time, cancelled_by, cancelled_reason, comment)
+values (3, 2, now()::date, '14:00:00', '15:00:00', true, now(), 'USER1', 'Location unavailable', null);
+
+insert into attendance(attendance_id, scheduled_instance_id, prisoner_number, attendance_reason_id, comment, posted, recorded_time, recorded_by, status, pay_amount, bonus_amount, pieces)
+values (1, 1, 'A11111A', null, null, false, null, null, 'SCHEDULED', null, null, null);
+
```
timharrison-moj, line 51: WAIT status?

timharrison-moj (commented):
LGTM. Couple of questions added. And a question I've just thought of, should we allow any user to cancel sessions? Or should we be looking for the ACTIVITY_HUB role - I'll have a think on this overnight.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Attendance.kt
leejacobson-moj: There is a check on the scheduled instance. If the instance has been canceled, it will throw an error as per the ACs.

But to your point, this is why I wasn't sure if we would want to add a check at the attendance level. If for some reason an attendance is marked as cancelled but the instance is not, then canceling the instance can update any attendance records after marking the instance itself as cancelled.

We can always revisit this later, but for now it seems like we don't need / want any checks here.

timharrison-moj (approved):
LGTM

PR #219, opened 2023-03-15
SAA-396: Mark non attendance on scheduled instances (sub task SAA-448)

gerryallan-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/model/Attendance.kt
```diff
@@ -42,4 +39,16 @@ data class Attendance(
   val bonusAmount: Int? = null,
 
   val pieces: Int? = null,
+
+  @Schema(description = "Should payment be issued for SICK, REST or OTHER", example = "true")
+  val issuePayment: Boolean?,
```
matt-crowson-moj, line 44: Can these booleans be null?  Either true or false?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/model/request/AttendanceUpdateRequest.kt
```diff
@@ -8,6 +8,21 @@ data class AttendanceUpdateRequest(
   @Schema(description = "The internally-generated ID for this attendance", example = "123456")
   val id: Long,
 
-  @Schema(description = "The reason codes- ABS, ACCAB, ATT, CANC, NREQ, SUS, UNACAB, REST", example = "ATT")
+  @Schema(description = "The reason codes- SICK, REFUSED, NREQ, REST, CLASH, OTHER, SUSP, CANC, ATT", example = "ATT")
```
matt-crowson-moj, line 11: Does the attendance reason (now) need a validation constraint adding to it.

Not sure what else would require it but this one should be mandatory right?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/AttendancesService.kt
```diff
@@ -43,6 +43,9 @@ class AttendancesService(
         attendanceReason =
```
matt-crowson-moj, line 43: Can we introduce a method on the attendance to deal with completing/marking attendance?  i.e. push this kinda of stuff into the entity/domain.

So in the apply function you would end up with something like

```
it.apply{ attendance ->
  attendance.mark(...pass necessary params in here ..)
}
```

Also

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/model/request/AttendanceUpdateRequest.kt
gerryallan-moj: I believe the idea is to leave this data driven from the attendance_reasons table, so that new reasons can be added and included in the page with little or no code changes

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/AttendancesService.kt
gerryallan-moj: Corrected

matt-crowson-moj (approved):
LGTM

timharrison-moj (approved):
Superb Gerry. Thanks. LGTM.

PR #227, opened 2023-03-22
SAA-490 get scheduled events in parallel using Kotlin coroutines

timharrison-moj:
This PR:
* Replaces the Mono.zip and Mono<List<T>> return types from prison API with Kotlin coroutines and Deferred<List<T>>
* It uses runBlocking {} enclosures, suspend functions and async {} calls for visits, activities, transfers, and court hearings.
* Removes the confusing, multi-layered transform functions for scheduled events - much clearer (I think).
* Allows (as a next step) the individual transform of court hearings, activities, appointments, visits and transfers when needed.
* We already know that we want to treat these events differently, show different detail, so this will be required.
* Updates all affected tests.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ScheduledEventService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ScheduledEventService.kt
@@ -1,6 +1,9 @@
 package uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.service
 
 import jakarta.persistence.EntityNotFoundException
+import kotlinx.coroutines.async
+import kotlinx.coroutines.coroutineScope
+import kotlinx.coroutines.runBlocking
 import org.springframework.stereotype.Service
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.client.prisonapi.api.PrisonApiClient
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.client.prisonapi.model.CourtHearings
@@ -19,9 +22,6 @@ import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.util.prisonApiP
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.util.prisonApiScheduledEventToScheduledEvents
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.util.transformPrisonerScheduledActivityToScheduledEvents
 import java.time.LocalDate
-import kotlinx.coroutines.async
-import kotlinx.coroutines.coroutineScope
-import kotlinx.coroutines.runBlocking
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.client.prisonapi.model.ScheduledEvent as PrisonApiScheduledEvent
 
 @Service
@@ -146,8 +146,7 @@ class ScheduledEventService(
     val activities = async {
       if (prisonRolledOut.active) {
         emptyList()
-      }
-      else {
+      } else {
         prisonApiClient.getScheduledActivitiesAsync(prisoner.first, dateRange)
       }
     }
@@ -161,7 +160,7 @@ class ScheduledEventService(
       courtHearings.await(),
       visits.await(),
       activities.await(),
-      transfers.await()
+      transfers.await(),
     )
   }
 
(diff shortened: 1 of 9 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ScheduledEventServiceTest.kt
```diff
@@ -1,18 +1,21 @@
 package uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.service
 
 import jakarta.persistence.EntityNotFoundException
+import kotlinx.coroutines.runBlocking
```
matt-crowson-moj, line 4: Still one hell of a big unit test.  Over 1200 lines.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ScheduledEventService.kt
```diff
@@ -51,22 +59,51 @@ class ScheduledEventService(
...
+              prisonCode,
+              EventType.EXTERNAL_TRANSFER.name,
+              EventType.EXTERNAL_TRANSFER.defaultPriority,
+              eventPriorities[EventType.EXTERNAL_TRANSFER],
+            ),
           )
-        }.block()
-        ?.apply {
+        }
+        .apply {
```
matt-crowson-moj, line 106: I realise this was already here.  I never quite liked the fact it mutated activities after the fact (shrug).

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ScheduledEventService.kt
```diff
@@ -51,22 +59,51 @@ class ScheduledEventService(
       val prisonRolledOut = rolloutPrisonRepository.findByCode(prisonCode)
         ?: throw EntityNotFoundException("Unable to get scheduled events. Could not find prison with code $prisonCode")
       val eventPriorities = prisonRegimeService.getEventPrioritiesForPrison(prisonCode)
-      getSinglePrisonerEventCalls(bookingId to prisonerNumber, prisonRolledOut, dateRange)
-        .map { t ->
-          transformToPrisonerScheduledEvents(
-            bookingId,
+
+      runBlocking {
```
matt-crowson-moj, line 63: I think the use of runBlocking here is OK.  This is our bridge between the async and synchronous world.

I guess how much effort would it be to make this a suspend function?

Obviously this would propagate up the call stack.  It would however mean we could use coroutineScope if we did.

matt-crowson-moj:
Overall this looks good to me!

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ScheduledEventServiceTest.kt
timharrison-moj: I know. I think this wants breaking down to test the various types of event, and their variants. Another ticket?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ScheduledEventService.kt
timharrison-moj: Again, not ideal, and maybe an area we can look at to refine?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/client/prisonapi/api/PrisonApiClientTest.kt
```diff
@@ -110,21 +120,25 @@ class PrisonApiClientTest {
     val dateRange = LocalDateRange(LocalDate.of(2022, 10, 1), LocalDate.of(2022, 11, 5))
 
     prisonApiMockServer.stubGetScheduledActivitiesNotFound(bookingId, dateRange.start, dateRange.endInclusive)
-    assertThatThrownBy { prisonApiClient.getScheduledActivities(bookingId, dateRange).block() }
+
```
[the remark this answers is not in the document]
timharrison-moj: Let me check these again - I think there was a problem in runBlocking at the function level when using the `assertThrows` check.

PR #229, opened 2023-03-24
SAA-499: Retrieve appointment instance endpoint for sync

iterative-it:
Key points:

- Have replaced the appointment_instance table with a view to reduce data duplication
- This is being merged into the SAA-484 PR as both contain database changes and I'd like both to go out at once


matt-crowson-moj (approved):
LGTM

PR #230, opened 2023-03-24
SAA-500: Publish appointment instance created event

iterative-it:


matt-crowson-moj (approved):
LGTM - worth checking if Syscon are ready to consume.  Otherwise will start raising alerts.

matt-crowson-moj (approved):
LGTM

PR #231, opened 2023-03-27
SAA-522: add attendance_history table

gerryallan-moj:
Add table in preparation for multiple API changes.


timharrison-moj (approved):
LGTM.
Is the plan to only get a row here when the original attendance is changed?  So most attendances will have no history row?

matt-crowson-moj:
Do we need this PR?  Looks like it is also on the later one.

timharrison-moj (approved):
LGTM

PR #234, opened 2023-03-28
SAA-407: Refactor migration scripts to allow for ref data to be loaded into different environments

smcveigh941:


src/main/resources/application.yml
```diff
@@ -61,7 +61,7 @@ spring:
       validationTimeout: 500
 
   flyway:
-    locations: classpath:/migration/common
+    locations: classpath:/migration/baseline,classpath:/migration/changes,classpath:/reference_data/common,classpath:/reference_data/prod
```
matt-crowson-moj, line 64: Not sure we really need the baseline folder anymore, meh!

I would change the repeatable migrations to be V__ instead of R__.

The original intent with the R__ files was to support repeatable migrations but we are not using this feature so perhaps just change them to V__ ?

matt-crowson-moj (approved):
LGTM

gerryallan-moj (approved):

PR #237, opened 2023-03-29
SAA-523: Create history record when marked attendance records are updated

gerryallan-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/AttendancesService.kt
```diff
@@ -99,4 +108,24 @@ class AttendancesService(
 
   fun getAttendanceById(id: Long): ModelAttendance =
     attendanceRepository.findOrThrowNotFound(id).toModel()
+
+  fun createHistory(
+    attendance: Attendance,
+  ) {
```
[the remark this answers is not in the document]
matt-crowson-moj: Ideally we should be letting the attendance entity itself manage its history.

So when the attendance changes e.g. it is updated a history record is created when we persist the attendance record.  

Any changes to that attendance will be propagated via JPA/Hibernate.

As Steve has suggested (though not using pre-persist) this could be done in the `@PreUpdate` annotation on an `AttendanceEntityListener` class.  In our case the attendance record will always exists ergo it will be an update to the attendance.

The end result being we do not need the call save on the history repo.  It is done when we save the attendance itself.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Attendance.kt
```diff
@@ -95,6 +96,7 @@ data class Attendance(
   }
 
   fun mark(
+    principal: Principal,
```
matt-crowson-moj, line 99: I'd probably just pass who did the marking hrough as a String.  Just keeps things simpler.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/AttendancesService.kt
```diff
@@ -99,4 +108,24 @@ class AttendancesService(
...
+    if (attendance.status != AttendanceStatus.WAITING) {
```
matt-crowson-moj, line 115: I can't see any integration test coverage for this?

When going down the entity listener pre-update route I'd suggest you add coverage to verity the listener is doing what it should and creating a history record.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Attendance.kt
gerryallan-moj: corrected

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/AttendancesService.kt
gerryallan-moj: Integration test added

PR #239, opened 2023-03-30
SDIT-423 Add endpoint for consolidated attendance sync details

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/AttendanceSync.kt
```diff
@@ -0,0 +1,61 @@
+package uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.entity
```
mikehalmamoj, line 1: I considered putting all of the sync related stuff into its own package so that it is separate from the main application (and I do like a domain based package structure). On the other hand it's good for the sync stuff to be visible so left it alongside the main code. I don't really have a strong opinion either way but thought I'd mention it.

matt-crowson-moj (approved):
LGTM

PR #243, opened 2023-03-31
SDIT-423 Stick to springdoc version 2.0.2

mikehalmamoj:
Version 2.0.3 contains a breaking change that reverses the default of the required flag for non-nullable fields from true to false. This means we'd need to add `required = true` to all non-nullable fields with a @Schema annotation - which is a lot of work and seems completely wrong to me. So I'm trying to get springdoc to reverse this breaking change, but in the meantime we need to stick to v2.0.2 so that all non-nullable fields are required.


matt-crowson-moj (approved):
LGTM

PR #244, opened 2023-04-01
SAA-471 - Add local audit functionality

sioldham:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/ActivityEntityListener.kt
```diff
@@ -0,0 +1,36 @@
...
+  private lateinit var applicationContext: ApplicationContext
+
+  companion object {
+    private val log: Logger = LoggerFactory.getLogger(this::class.java)
+  }
+
+  @PostPersist
+  fun onCreate(activity: Activity) {
+    // TODO Is there a way to make this work??
+    val auditService = applicationContext.getBean(AuditService::class.java)
```
sioldham, line 25: Trying to autowire anything into an entity listener that references the database (even a data source) is problematic; it just doesn't work. In fact it stops _anything_ being autowired into the entity listener (I autowired the `AuditService` into the `AllocationEntityListener` by way of an experiment and not only was the `AuditService` null but so was the `OutboundEventService` that had hitherto autowired correctly)

Originally, I tried to autowire in the `AuditService` (late init etc) and it was always null _unless_ I removed the reference to the `AuditRepository` from the `AuditService` in which case it autowired correctly.

All the solutions I have seen online are hacks of this nature but perhaps someone else in the team has an idea.

helm_deploy/values-dev.yaml
```diff
@@ -14,7 +14,8 @@ generic-service:
     APPLICATIONINSIGHTS_CONFIGURATION_FILE: applicationinsights.dev.json
     SPRING_PROFILES_ACTIVE: dev
     SCHEDULE_AHEAD_DAYS: 14
-    FEATURE_AUDIT_SERVICE_ENABLED: true
+    FEATURE_AUDIT_SERVICE_HMPPS_ENABLED: false
```
sioldham, line 17: Disable for the time being and do the integration with HMPPS audit under a separate ticket. We can demo Local Audit in the meantime

helm_deploy/values-preprod.yaml
```diff
@@ -10,7 +10,8 @@ generic-service:
     PRISON_API_URL: https://api-preprod.prison.service.justice.gov.uk
     PRISONER_SEARCH_API_URL: https://prisoner-offender-search-preprod.prison.service.justice.gov.uk
     APPLICATIONINSIGHTS_CONFIGURATION_FILE: applicationinsights.dev.json
-    FEATURE_AUDIT_SERVICE_ENABLED: true
+    FEATURE_AUDIT_SERVICE_HMPPS_ENABLED: false
```
sioldham, line 13: See comment above

helm_deploy/values-prod.yaml
```diff
@@ -10,7 +10,8 @@ generic-service:
     PRISON_API_URL: https://api.prison.service.justice.gov.uk
     PRISONER_SEARCH_API_URL: https://prisoner-offender-search.prison.service.justice.gov.uk
     APPLICATIONINSIGHTS_CONFIGURATION_FILE: applicationinsights.json
-    FEATURE_AUDIT_SERVICE_ENABLED: true
+    FEATURE_AUDIT_SERVICE_HMPPS_ENABLED: false
```
sioldham, line 13: See comment above

helm_deploy/values-dev.yaml
matt-crowson-moj: Not sure we have this quite right. Probably best to chat about it.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/ActivityEntityListener.kt
sioldham: Yeah, it's having a reference to anything database related in the chain of objects being injected 😢

matt-crowson-moj: OK well I am not sure there is a clean way around this.  I would pull the accessing of the context like this out and wrap it in something audit specific and then wire that in.  It is likely we will use this in other places for auditing other changes so we can reuse the wrapper.  For example:

```
@Component
class AuditUtil {

  @Autowired
  private lateinit var applicationContext: ApplicationContext

  fun logEvent(event: AuditableEvent) {
    applicationContext.getBean(AuditService::class.java).logEvent(event)
  }
}

@Component
class ActivityEntityListener {

  @Autowired
  private lateinit var auditing: AuditUtil
....
}
```

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/ActivityEntityListener.kt
```diff
@@ -0,0 +1,36 @@
...
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.context.ApplicationContext
+import org.springframework.stereotype.Component
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.service.AuditService
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.util.toActivityCreatedEvent
+
+@Component
+class ActivityEntityListener {
```
matt-crowson-moj, line 13: Should this just be called something like AuditingEntitiyListener?

It can then be reused for other entities as well?  Not sure we need a separate listener class for each entity???

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/model/audit/AuditModelUtils.kt
```diff
@@ -0,0 +1,37 @@
+package uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.model.audit
+
+import net.minidev.json.JSONObject
+import java.time.LocalDate
+import java.time.LocalDateTime
+import java.time.LocalTime
+import java.time.format.DateTimeFormatter
+
+object AuditModelUtils {
```
matt-crowson-moj, line 9: This is really a Kotlin style thing. Don't really need single `object`  to wrap this function, it doesn't really add anything having it in an `object`.

You could just have the function in a file by itself.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/ActivityEntityListener.kt
sioldham: Sure, makes sense (assuming it will work with multiple, overloaded `@PostPersist` methods)

helm_deploy/values-prod.yaml
sioldham: Fixed as discussed

helm_deploy/values-preprod.yaml
sioldham: Fixed as discussed

helm_deploy/values-dev.yaml
sioldham: Fixed as discussed

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/ActivityEntityListener.kt
sioldham: Done

matt-crowson-moj (approved):
LGTM

PR #245, opened 2023-04-02
SDIT-423 Use latest springdoc including required fix

mikehalmamoj:
springdoc fixed in https://github.com/springdoc/springdoc-openapi/issues/2185


matt-crowson-moj (approved):
LGTM - wow quick fix that one.

PR #247, opened 2023-04-04
SAA-548: Added appointment types (INDIVIDUAL|GROUP)

leejacobson-moj:
## Overview

- Adds new column `appointment_type` to `appointments` table
- Adds a required field, `appointmentType` to the create appointment request.
- Returns `appointmentType` in endpoints which return appointment or appointment occurrence details


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/AppointmentOccurrence.kt
```diff
@@ -130,6 +134,12 @@ data class AppointmentOccurrence(
       },
       prisoners.toSummary(),
     )
+
+  private fun failIfIndividualAppointmentAlreadyAllocated() {
+    if (allocations.size > 0 && appointment.appointmentType == AppointmentType.INDIVIDUAL) {
```
matt-crowson-moj, line 139: Minor point.

Perhaps swap these round so it checks the appointment type first?

Just one less hop/check if the appointment type is not individual i.e it will only do the size check if it is individual.  Also can use `isNotEmpty()`

```
if (appointment.appointmentType == AppointmentType.INDIVIDUAL && allocations.isNotEmpty() {
...
}
```

src/main/resources/migration/V2.0__appointments_type.sql
```diff
@@ -0,0 +1,17 @@
+ALTER TABLE appointment ADD COLUMN appointment_type varchar(10);
```
matt-crowson-moj, line 1: Not sure if we really need the migration yet.  I know there is one for the sync migration but I would still add this column in the appointments baseline script for now.  And yes it would require a DB drop prior to merging.

matt-crowson-moj (commented):
LGTM

iterative-it (approved):
LGTM

iterative-it (approved):
LGTM

matt-crowson-moj (approved):
LGTM

PR #251, opened 2023-04-05
SAA-554 small change to the event listener and refactoring the auditing to use the feature switch component.

matt-crowson-moj:


timharrison-moj (approved):
LGTM

PR #252, opened 2023-04-05
SAA-502: Add previous and next instance IDs and Dates to the Instance

gerryallan-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/ActivitySchedule.kt
```diff
@@ -242,6 +242,16 @@ data class ActivitySchedule(
   override fun toString(): String {
     return this::class.simpleName + "(activityScheduleId = $activityScheduleId )"
   }
+
+  fun previous(scheduledInstance: ScheduledInstance): ScheduledInstance? =
```
matt-crowson-moj, line 246: Look good but you should really have unit test coverage for these ...

matt-crowson-moj (approved):
LGTM approved but can you add some test coverage even if in a separate PR.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/ActivitySchedule.kt
```diff
@@ -242,6 +242,16 @@ data class ActivitySchedule(
...
+    instances()
+      .sortedWith(compareBy<ScheduledInstance> { it.sessionDate }.thenBy { it.startTime })
+      .let { sorted -> sorted.getOrNull(sorted.indexOf(scheduledInstance) - 1) }
+
```
timharrison-moj, line 250: Can this attempt to get the item before the 0th element?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/ActivitySchedule.kt
```diff
@@ -242,6 +242,16 @@ data class ActivitySchedule(
...
+  fun next(scheduledInstance: ScheduledInstance): ScheduledInstance? =
+    instances()
+      .sortedWith(compareBy<ScheduledInstance> { it.sessionDate }.thenBy { it.startTime })
+      .let { sorted -> sorted.getOrNull(sorted.indexOf(scheduledInstance) + 1) }
 }
```
timharrison-moj, line 255: Could this go off the end of the list?

matt-crowson-moj: My understanding is it will simply return null in both cases.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/ActivitySchedule.kt
gerryallan-moj: Will add in another PR

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/ActivitySchedule.kt
gerryallan-moj: Matt is correct - I have tested at both ends of the list

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/ActivitySchedule.kt
gerryallan-moj: It will just return null

matt-crowson-moj (dismissed):
LGTM don't forget to add tests though for previous and next ...

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/ActivityScheduleTest.kt
```diff
@@ -489,4 +489,35 @@ class ActivityScheduleTest {
...
+    }
+
+    scheduleWithInstances.addInstance(today, scheduleWithInstances.slots().first())
+    scheduleWithInstances.addInstance(today.plusDays(1L), scheduleWithInstances.slots().first())
+    scheduleWithInstances.addInstance(today.plusDays(2L), scheduleWithInstances.slots().first())
+
+    assertThat(scheduleWithInstances.instances()).hasSize(3)
+
+    assertThat(scheduleWithInstances.previous(scheduleWithInstances.instances().first())?.toModel()?.previousScheduledInstanceDate).isNull()
+    assertThat(scheduleWithInstances.next(scheduleWithInstances.instances().first())?.toModel()?.nextScheduledInstanceDate).isNotNull()
```
matt-crowson-moj, line 518: I'd suggest you check actual expected values versus them being not null.

For example in this case the dates should be different and we can't say for certain they are with correct saying `isNotNull()`.

gerryallan-moj: Updated

matt-crowson-moj (approved):
LGTM

PR #261, opened 2023-04-10
SAA-547 Improvements for scheduled event switching, retrieval, transform and display

timharrison-moj:
** Will need API types regenerating in the UI service **

This PR contains a number of change and improvements:
* Standardises the appointment events from prison API (single and multiple) to use async functions like the other events.
* Moves the appointment switch and transform logic from the `AppointmentInstanceService` into the `ScheduledEventService`.
* Avoids the need for appointments to be transformed into prison API events first, which loses much of their detail.
* Provides a direct appointment -> scheduled event transform to retain and enrich the detail.
* Allow either activities or appointments to be rolled out first.
* Extends the `ScheduledEvent` model to provide more detail for all events (IDs, cancelled, category, location, in-cell, outside)
* Removes the deprecated prison API blocking calls and tests.
* Simplified transforms, switches, services, and tests for all scheduled events.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/model/ScheduledEvent.kt
```diff
@@ -11,54 +11,78 @@ data class ScheduledEvent(
   @Schema(description = "The prison code for this scheduled event", example = "MDI")
   val prisonCode: String?,
```
timharrison-moj, line 13: This is the extended version of the `ScheduledEvent` model that is returned to clients. It now provides specific IDs for activities, appointments, adjudication hearings so that our UI can provides links and better descriptions of the type and category of event.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/resource/ScheduledEventController.kt
```diff
@@ -21,12 +21,19 @@ import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.common.LocalDat
...
 import java.time.LocalDate
 
 @RestController
 @RequestMapping("/scheduled-events")
-class ScheduledEventController(private val scheduledEventService: ScheduledEventService) {
+class ScheduledEventController(
+  private val scheduledEventService: ScheduledEventService,
+  private val referenceCodeService: ReferenceCodeService,
+  private val locationService: LocationService,
+) {
```
timharrison-moj, line 37: The reference code and location services are required to enrich appointment types and locations with more detail - appointments only store the codes for these things, not the descriptions too.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/AppointmentInstanceService.kt
```diff
@@ -1,105 +1,11 @@
 package uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.service
 
-import org.slf4j.Logger
```
timharrison-moj, line 3: This service is no longer responsible for switching the appointment source, nor retrieving their details. It has all been moved into the scheduled event service and operates in the same way for both activities and appointments.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/util/PrisonApiTransformations.kt
```diff
@@ -98,64 +111,138 @@ fun PrisonApiCourtHearings.prisonApiCourtHearingsToScheduledEvents(
...
+    bookingId = it.bookingId,
+    internalLocationId = it.eventLocationId,
+    internalLocationCode = it.eventLocation,
+    internalLocationDescription = it.eventLocation,
+    cancelled = it.eventStatus == "CANC",
+    inCell = false,
+    categoryCode = it.eventSubType,
+    categoryDescription = it.eventSubTypeDesc,
+    summary = "Appointment ${it.eventSubTypeDesc}",
+    comments = it.outcomeComment,
```
iterative-it, line 175: This should map to comment rather than outcomeComment I think. outcomeComment is always null or empty for appointments looking at the prison API code

src/test/resources/test_data/seed-appointment-single-id-3.sql
```diff
@@ -0,0 +1,8 @@
+INSERT INTO appointment (appointment_id, appointment_type, category_code, prison_code, internal_location_id, in_cell, start_date, start_time, end_time, comment, created, created_by)
+VALUES (3, 'INDIVIDUAL', 'AC1', 'MDI', 123, false, '2022-10-01', '09:00', '10:30', 'Appointment level comment', now()::timestamp, 'TEST.USER');
```
iterative-it, line 2: Is this test data needed in addition to the `seed-appointment-single-id-1.sql` file? Mainly to slightly reduce the maintenance of this seed data if the schema changes. Main difference seems to be the prison code and prisoner number. Perhaps the other file and related tests could be updated? Maybe in a tech debt ticket?

iterative-it (commented):
Looking good and a really positive change in a key area. Particularly like the move to async, coroutines and parallelism.

I would change the comments mapping and personally I'd switch to enums but the rest are comments and suggestions only

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/model/ScheduledEvent.kt
```diff
@@ -11,54 +11,78 @@ data class ScheduledEvent(
...
 
-  @Schema(description = "The event id for this scheduled event", example = "10001")
+  @Schema(description = "The source of this event - valid values are NOMIS or SAA (scheduling activities and appointments)", example = "NOMIS")
+  val eventSource: String?,
+
+  @Schema(description = "The event type (APPOINTMENT, ACTIVITY, COURT, TRANSFER, ADJUDICATION, VISIT)", example = "APPOINTMENT")
+  val eventType: String?,
```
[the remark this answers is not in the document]
timharrison-moj: Agree with this, Dave. Can I do as a small follow-on? I've run through all tests in API, UI, cypress and running-locally, so would prefer to get this in as-is and then do these updates if that's ok?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/util/PrisonApiTransformations.kt
timharrison-moj: Also agree with this, Dave. Same request - update as part of the small follow-on?

src/test/resources/test_data/seed-appointment-single-id-4.sql
```diff
@@ -0,0 +1,11 @@
+INSERT INTO appointment (appointment_id, appointment_type, category_code, prison_code, internal_location_id, in_cell, start_date, start_time, end_time, comment, created, created_by)
+VALUES (3, 'INDIVIDUAL', 'AC1', 'MDI', 123, false, '2022-10-01', '09:00', '10:30', 'Appointment level comment', now()::timestamp, 'TEST.USER');
```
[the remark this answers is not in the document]
timharrison-moj: I did think that at the time! Ok, same follow-on PR?

iterative-it (approved):
Significant and positive change

iterative-it (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/util/PrisonApiTransformations.kt
timharrison-moj: There is no comment field returned in the prison API response for appointments by a single booking id I'm afraid - it uses the ScheduledEvent response class, and it's absent. When we get appointments from NOMIS for multiple people, this uses the PrisonerSchedule class, and this does have the comment field, and is used.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/model/ScheduledEvent.kt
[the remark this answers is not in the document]
timharrison-moj: Defined an enum type with the values.

PR #265, opened 2023-04-13
SAA-584: Appointment location and review comments.

timharrison-moj:
Minor changes only.
* Use the correct location description for appointments (`userDescription` in place of `description` on a Location object)
* Addressed one or two review comments.


iterative-it (approved):
LGTM

PR #268, opened 2023-04-17
SAA-587: minor issues with the attend journey

gerryallan-moj:
Display history records in reverse chronological order
Pay content should be the Amount or ‘none’

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Attendance.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Attendance.kt
@@ -184,7 +184,7 @@ data class Attendance(
       .sortedWith(compareBy { this.recordedTime })
       .reversed()
       .map { attendanceHistoryRow -> transform(attendanceHistoryRow) },
-    )
+  )
 }
 
 enum class AttendanceStatus {
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Attendance.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Attendance.kt
@@ -181,9 +181,10 @@ data class Attendance(
     issuePayment = this.issuePayment,
     incentiveLevelWarningIssued = this.incentiveLevelWarningIssued,
     attendanceHistory = this.attendanceHistory
-      .map { attendanceHistory -> transform(attendanceHistory) }
-      .reversed(),
-  )
+      .sortedWith(compareBy { this.recordedTime })
+      .reversed()
+      .map { attendanceHistoryRow -> transform(attendanceHistoryRow) },
+    )
 }
 
 enum class AttendanceStatus {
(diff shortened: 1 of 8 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Attendance.kt
```diff
@@ -100,6 +100,7 @@ data class Attendance(
     newStatus: AttendanceStatus,
     newComment: String?,
     newIssuePayment: Boolean?,
+    newPayAmount: Int?,
     newIncentiveLevelWarningIssued: Boolean?,
```
timharrison-moj, line 104: When marking attendance, can a user specify a pay amount?  I ask because I don't think NOMIS will understand this - it will only look at the pay rates set against this activity by incentive level and pay band, and award those amounts. Might need to review what the design is doing for this?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/util/TransformFunctions.kt
```diff
@@ -305,6 +305,9 @@ fun transform(attendance: EntityAttendance): ModelAttendance =
     pieces = attendance.pieces,
     issuePayment = attendance.issuePayment,
     incentiveLevelWarningIssued = attendance.incentiveLevelWarningIssued,
+    attendanceHistory = attendance.history()
+      .map { attendanceHistory: EntityAttendanceHistory -> transform(attendanceHistory) }
+      .reversed(),
   )
```
timharrison-moj, line 311: Is this to return the history in reverse date/time order? Looks good.

timharrison-moj (commented):
LGTM. Just the one question about why the user is allowed to specify an actual amount?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Attendance.kt
gerryallan-moj: No - this is just to allow the pay_amount to be nullified when the attendance is unmarked.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/util/TransformFunctions.kt
gerryallan-moj: Yes

matt-crowson-moj: Is the history sorted by date and time though first?  Reversing will just reverse a list won't it?

gerryallan-moj: Yes - the history is already chronological, so reversing it gives us what we need

matt-crowson-moj: Sorry to go on but how so?  I appreciate they are created chronologically but can you guarantee the order on which they will be retrieved from the database?

matt-crowson-moj (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/util/TransformFunctions.kt
gerryallan-moj: Added a sortedWith to make sure

PR #271, opened 2023-04-18
SAA-517 part one of adding ability to reactivate auto-suspended allocations on receipt of prisoner back into prison from temporary release.

matt-crowson-moj:
This is the first PR of two for reactivation of auto-suspended allocations when a prisoner comes back into prison after temporary release.

The second PR to come will intercept the Domain event to call this code.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/InboundEventsProcessor.kt
```diff
@@ -0,0 +1,63 @@
...
+
+import org.slf4j.LoggerFactory
+import org.springframework.stereotype.Component
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.entity.Allocation
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.entity.PrisonerStatus
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.repository.AllocationRepository
+import java.time.LocalDateTime
+
+@Component
+class InboundEventsProcessor(private val repository: AllocationRepository) {
```
matt-crowson-moj, line 11: To be used by the already existing InboundEventsService.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/InboundEventsProcessor.kt
```diff
@@ -0,0 +1,63 @@
...
+
+  companion object {
+    private val log = LoggerFactory.getLogger(this::class.java)
+  }
+
```
timharrison-moj, line 16: Do you think we need an initial function to determine if this event is "interesting"? Rules might be - is the prisoner at a prison which is rolled out? Does the prisoner have any active allocations? Is the event type interesting/applicable? I think these current events (release/transfer) are ALWAYS interesting, so perhaps its a task for a later ticket around change of circumstances, thought the prison roll-out status is definitely something we'd want to check, even for these?

timharrison-moj (approved):
LGTM. One question only, and that could be an iteration on next PR?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/InboundEventsProcessor.kt
matt-crowson-moj: The rollout prison check is a good point.  Having one entry point (function) will certainly make this easy to check from here.

If the prisoner isn't at the prison in question then the repo call will not find anything to process.  Is this OK?  I think so though there are bound to be edge cases where this is not OK.

We could extend the repo call to only return the rows we are interested in changing as well i.e. where the status is ACTIVE, AUTO_SUSPENDED, ENDED?  This would negate the need for the filter checks further on.

PR #278, opened 2023-04-19
SAA-603 implement outbound event for scheduled instance amended

timharrison-moj:
* Entity listener on amend of scheduled instance
* New event for `activities.scheduled-instance.amended`
* This is intended to catch the scenario where a scheduled instance is cancelled or uncancelled.


helm_deploy/hmpps-activities-management-api/values.yaml
```diff
@@ -35,8 +35,9 @@ generic-service:
     DB_SSL_MODE: "verify-full"
     SCHEDULE_AHEAD_DAYS: 60
     SCHEDULE_AHEAD_DAYS_ONLINE: 14
-    FEATURE_EVENTS_SNS_ENABLED: true
+    FEATURE_EVENTS_SNS_ENABLED: false
     FEATURE_EVENT_ACTIVITIES_ACTIVITY_SCHEDULE_CREATED: true
```
timharrison-moj, line 39: Correcting a previous issue - FEATURE_EVENTS_SNS_ENABLED should be false in values.yaml, and is overridden in environment-specific files.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/ScheduledInstanceEntityListener.kt
```diff
@@ -0,0 +1,37 @@
...
+import jakarta.persistence.PostUpdate
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.stereotype.Component
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.service.OutboundEvent
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.service.OutboundEventsService
+
+@Component
+class ScheduledInstanceEntityListener {
```
matt-crowson-moj, line 12: Just a thought

I do wonder if we could get a way with one listener which then just checks the type and decides what to do e.g.

```
@Component
class EventPublisherListener {

  @Autowired
  private lateinit var outboundEventsService: OutboundEventsService

  companion object {
    private val log: Logger = LoggerFactory.getLogger(this::class.java)
  }

  @PostPersist
  fun onCreate(schedule: Any) {
    when(entity) {
      is Activity -> blah bah
      is ScheduledInstance -> blah blah
    }
  }

  @PostUpdate
  fun onUpdate(entity: Any) {
    when(entity) {
      is Activity -> blah bah
      is ScheduledInstance -> blah blah
    }
  }
  ....
}
```

matt-crowson-moj (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/ScheduledInstanceEntityListener.kt
timharrison-moj: Interesting - would produce far fewer modules / listeners and send() functions. One for a trial refactor later?

matt-crowson-moj (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/ScheduledInstanceEntityListener.kt
matt-crowson-moj: This is already being done for auditing.

PR #283, opened 2023-04-21
SAA-517 unknown release events should not impact on existing allocations.

matt-crowson-moj:
These changes extend the coverage on the prison release event to cover the case where the release reason is unknown.  If this is the case then we cannot make an informed decision on changing existing allocations and as a result we ignore it.

Note there is a separate piece of work to surface events of interest to the end user so ideally they can make a manual decision.  That will pick up on the unknown case here.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/events/InboundEventsProcessor.kt
```diff
@@ -19,10 +19,11 @@ class InboundEventsProcessor(private val repository: AllocationRepository) {
...
       is OffenderReleasedEvent -> {
-        if (inboundEvent.isTemporary()) {
-          suspendOffenderAllocations(inboundEvent)
-        } else {
-          deallocateOffenderAllocations(inboundEvent)
+        when {
+          inboundEvent.isTemporary() -> suspendOffenderAllocations(inboundEvent)
+          inboundEvent.isPermanent() -> deallocateOffenderAllocations(inboundEvent)
+          // TODO pick up with the event of interest work.
+          else -> log.info("Ignoring event of potential interest $inboundEvent.")
```
matt-crowson-moj, line 26: Logging for now.  There is a separate piece of work to pick this up as an event or potential interest.

PR #285, opened 2023-04-24
SAA-593: Integration with case notes service

gerryallan-moj:


matt-crowson-moj:
Could do with a shell script for generating the model objects for the case notes API.  Similar to that of the prisoner API and prisoner search API.

src/main/resources/application-local.yml
```diff
@@ -43,6 +43,11 @@ prisoner-search:
     url: https://prisoner-offender-search-dev.prison.service.justice.gov.uk
     timeout: 10s
 
+case-notes:
+  api:
+    url: https://dev.offender-case-notes.service.justice.gov.uk
+    timeout: 10s
```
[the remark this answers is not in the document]
gerryallan-moj: Added to helm charts

gerryallan-moj:
> Could do with a shell script for generating the model objects for the case notes API. Similar to that of the prisoner API and prisoner search API.

see generate-casenotes-api-types.sh in root folder

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/AttendancesService.kt
```diff
@@ -40,6 +43,10 @@ class AttendancesService(
     val attendanceReasonsByCode = attendanceReasonRepository.findAll().associateBy { it.code.uppercase().trim() }
 
     val updatedAttendances = attendanceRepository.findAllById(attendanceUpdatesById.keys).mapNotNull {
+      var caseNoteDetails: CaseNote? = null
+      if (!attendanceUpdatesById[it.attendanceId]!!.caseNote.isNullOrEmpty()) {
```
matt-crowson-moj, line 47: This `if (!attendanceUpdatesById[it.attendanceId]!!` will not fail nicely if there is no match on the attendance id.  Ideally it should be handle it if it is null and not just let the !! throw a null pointer exception.

You can probably try something like this:
```
if (!attendanceUpdatesById.getOrElse(it.attendanceId) { throw IllegalArgumentException("Put meaningful message here") } ....
```

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/integration/wiremock/CaseNotesApiMockServer.kt
```diff
@@ -0,0 +1,22 @@
...
+
+  private val mapper = ObjectMapper()
+
+  fun stubGetCaseNote(prisonerNumber: String, caseNoteId: Long) {
+    stubFor(
+      WireMock.get(WireMock.urlEqualTo("/case-notes/$prisonerNumber/$caseNoteId"))
+        .willReturn(
+          WireMock.aResponse()
+            .withHeader("Content-Type", "application/json")
+            .withBodyFile("casenotesapi/case-note-1.json")
```
matt-crowson-moj, line 17: If we are going to pull in JSON files like this it might be useful to put some sort of interpolation pattern so we can use different files.

```
 fun stubGetCaseNote(prisonerNumber: String, caseNoteId: Long) {
    stubFor(
      WireMock.get(WireMock.urlEqualTo("/case-notes/$prisonerNumber/$caseNoteId"))
        .willReturn(
          WireMock.aResponse()
            .withHeader("Content-Type", "application/json")
            .withBodyFile("casenotesapi/case-note-$caseNoteId.json")
```

You may also want to use the prison number, up to you.

matt-crowson-moj (commented):
Looking good.  A coupe of things to look at.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/AttendancesService.kt
gerryallan-moj: added a check

matt-crowson-moj (approved):
LGTM

PR #287, opened 2023-04-25
SAA-317 replacing inbound event processor with concept of event handlers so each event can be worked on in isolation.

matt-crowson-moj:
This pulls out event handling from one component e.g. the processor, into a handler for each event.

The inbound service then decides which handler to use and delegates accordingly.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/events/InboundEventsService.kt
```diff
@@ -17,7 +20,12 @@ class InboundEventsService(
 
   fun process(inboundEvent: InboundEvent) {
     if (rolloutPrisonRepository.findByCode(inboundEvent.prisonCode())?.active == true) {
-      processor.process(inboundEvent)
+      when (inboundEvent) {
+        is OffenderReceivedEvent -> receivedEventHandler.handle(inboundEvent)
```
matt-crowson-moj, line 24: @timharrison-moj - with this change it means I can now work on the offender received event in isolation so in theory you can work on events of interest and we shouldn't clash.

PR #293, opened 2023-04-26
SAA-387 tweaks to localstack config and adding wiremock to simulate prison-api calls for inbound events when using localstack aws events.

matt-crowson-moj:
Provides a bit of template to test events with localstack.

When using you may need to tweak or add files to get the particular scenario you need to test.


wiremock-docker/__files/prison-api/inmates/A11111A.json
```diff
@@ -0,0 +1,56 @@
+{
+  "_comment_": "Contains inmate that has been released from custodial sentence",
```
matt-crowson-moj, line 2: The comments lines are not returned by prison-api.  I added these to aid with showing the purpose of the files.

wiremock-docker/mappings/prison_api_inmate.json
```diff
@@ -0,0 +1,16 @@
+{
```
matt-crowson-moj, line 1: This will match on any offender number and attempt to find a matching files in the inmates folder.

PR #297, opened 2023-04-27
SAA-422 small refactoring exercise to consolodate the allocations job. This is going to be extended to include additional tasks in upcoming work.

matt-crowson-moj:
There are some extra tasks around changing the status of allocations that we can piggy back off the existing deallocation job hence the renaming to be managing allocations.

This helps prevent conflicts with regards to CRON timings as well.  All changes to the allocations will capture in job.


timharrison-moj (approved):
LGTM

PR #299, opened 2023-04-28
SAA-528: Prisoner events are marked as suspended if their allocation is suspended

smcveigh941:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/AttendancesService.kt
```diff
@@ -113,7 +113,7 @@ class AttendancesService(
           issuePayment = false,
           status = AttendanceStatus.COMPLETED,
           recordedTime = LocalDateTime.now(),
-          recordedBy = "SYSTEM",
```
matt-crowson-moj, line 116: SYSTEM is used in other places ...

PR #301, opened 2023-04-28
SAA-634: Added support for creating "inCell" activities

leejacobson-moj:
## Overview

This adds support for the "inCell" field on the create an activity request. If set to "true" the activity entity "inCell" field will be set to true and the location fields of it's schedules will be set to null.

Note: I've not added validation to the request to check if the location field is empty when the "inCell" field is set to "true". Instead if the "inCell" field is set to "true" the location field is simply ignored.


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityServiceTest.kt
```diff
@@ -370,4 +370,45 @@ class ActivityServiceTest {
       .isInstanceOf(IllegalArgumentException::class.java)
       .hasMessage("The education level code '1' is not active in NOMIS")
   }
+
+  @Test
+  fun `createActivity - Create In-cell activity`() {
+    val createdBy = "SCH_ACTIVITY"
+
+    val createActivityRequest: ActivityCreateRequest = mapper.readValue(
```
matt-crowson-moj, line 378: Very minor point perhaps name variable in line with the test e.g. createInCellActivityRequest?

Would be good to get some of the boilerplate code with reading JSON files.  For example we could add a reusable extension function to the mapper e.g. 

```
  internal inline fun <reified T> ObjectMapper.read(pathToFile: String): T =
    this.readValue(this::class.java.getResource(pathToFile), object : TypeReference<T>() {})

```

This would requite a bit of refactoring though.

matt-crowson-moj (approved):
LGTM

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityServiceTest.kt
leejacobson-moj: Updated the variable name.

If it's okay I'll leave the second suggestion for now as it's blocking my UI work, but if I'm in this in this area of code again I'll take a look and refactoring those JSON reads.

matt-crowson-moj (approved):
LGTM

PR #305, opened 2023-05-02
SAA-614: Issue payment on session cancellation

smcveigh941:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Attendance.kt
```diff
@@ -91,6 +91,7 @@ data class Attendance(
   fun cancel(reason: AttendanceReason, payIncentiveCode: String?) {
     status = AttendanceStatus.COMPLETED
     payAmount = if (payIncentiveCode != null) getPay(payIncentiveCode)?.rate ?: 0 else 0
+    issuePayment = true
```
matt-crowson-moj, line 94: Is this checked in a unit test?

PR #310, opened 2023-05-04
SAA-643 updaing build and the service to generate the prison api types instead of pulling them into the actual service code base.

matt-crowson-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/client/prisonapi/overrides/ReferenceCode.kt
```diff
@@ -1,4 +1,4 @@
-package uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.client.prisonapi.model
+package uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.client.prisonapi.overrides
```
matt-crowson-moj, line 1: For the time being we still need these to stop things from breaking.  In an ideal world we would not have these.

PR #317, opened 2023-05-05
SAA-677: Fix for nullable fields in events

timharrison-moj:
On Incentives review events the `reason` and `prisonId` fields are sometimes not filled in, and are null.
This was causing an exception when creating the intermediate object for incentives review events.


matt-crowson-moj (approved):
LGTM

PR #318, opened 2023-05-05
SAA-654: Record other absence reason

leejacobson-moj:
## Overview

Enables `otherAbsenceReason` field to be set when marking attendance


timharrison-moj (approved):
LGTM. I didn't notice any tests changing here, so perhaps some verification that they other absence reason is being set?

timharrison-moj (approved):
Perfect. LGTM.

PR #319, opened 2023-05-08
SAA-606: Override the Education class to fix nullable fields.

timharrison-moj:
Small fix for generated prison API types being different to expected nullable fields for Education.


matt-crowson-moj (approved):
LGTM

PR #326, opened 2023-05-12
SAA-676: New patch endpoint to update activities

gerryallan-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityService.kt
```diff
@@ -116,7 +119,7 @@ class ActivityService(
     }
 
     activity.let {
-      val scheduleLocation = if (request.inCell) null else getLocationForSchedule(it, request)
+      val scheduleLocation = if (request.inCell) null else getLocationForSchedule(it, request.locationId!!)
       val prisonRegime = prisonRegimeService.getPrisonRegimeByPrisonCode(activity.prisonCode)
```
timharrison-moj, line 123: If not changing the location, does the PATCH request need to supply a `locationId`? The double-pling suggests this must be present in the request.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityService.kt
```diff
@@ -220,4 +223,233 @@ class ActivityService(
...
+    }
+    if (request.slots != null) {
+      applySlotsUpdate(request.slots, activity)
+    }
+
+    activity.updatedTime = now
+    activity.updatedBy = updatedBy
+
+    activity.schedules().forEach { it.markAsUpdated(now, updatedBy) }
+
```
timharrison-moj, line 259: This is good - thanks - it makes sure that the activity schedule is always updated on a patch, and emit the right sync event.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityService.kt
```diff
@@ -220,4 +223,233 @@ class ActivityService(
...
+    activityRepository.saveAndFlush(activity)
+
+    activity = activityRepository.findOrThrowNotFound(activityId)
+
```
timharrison-moj, line 263: The `saveAndFlush` should return the updated activity, so probably no need to call the `findOrThrowNotFound()` here?

timharrison-moj: `return transform(activityRepository.saveAndFlush(activity))` should do.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityService.kt
```diff
@@ -220,4 +223,233 @@ class ActivityService(
...
+    applyStartDateUpdate(request, activity)
+    applyEndDateUpdate(request, activity)
+    applyMinimumIncentiveNomisCodeUpdate(request, activity)
+    applyMinimumIncentiveLevelUpdate(request, activity)
+    applyRunsOnBankHolidayUpdate(request, activity)
+    applyCapacityUpdate(request, activity)
+    applyRiskLevelUpdate(request, activity)
+    applyLocationUpdate(request, activity)
+    applyInCellUpdate(request, activity)
+    applyAttendanceRequiredUpdate(request, activity)
```
timharrison-moj, line 244: For these apply methods - the values could be nullable in the request, and checked here with `req.category?.let {  apply... }` - so that only the supplied values are updated? It means the PATCH would work with even a single value provided. Could be an iteration though..

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityService.kt
```diff
@@ -220,4 +223,233 @@ class ActivityService(
...
+    this.updatedBy = updatedBy
+  }
+
+  private fun applyCategoryUpdate(
+    request: ActivityUpdateRequest,
+    activity: Activity,
+  ) {
+    request.categoryId?.apply {
+      activity.activityCategory = activityCategoryRepository.findOrThrowIllegalArgument(this)
+    }
```
timharrison-moj, line 281: Ignore above comment - you've got the null-checks in the apply methods! Sorry.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityService.kt
```diff
@@ -220,4 +223,233 @@ class ActivityService(
...
+    val timeSlots =
+      mapOf(
+        TimeSlot.AM to Pair(prisonRegime.amStart, prisonRegime.amFinish),
+        TimeSlot.PM to Pair(prisonRegime.pmStart, prisonRegime.pmFinish),
+        TimeSlot.ED to Pair(prisonRegime.edStart, prisonRegime.edFinish),
+      )
+    activity.schedules().forEach {
+      it.removeSlots()
+      it.addSlots(slots, timeSlots)
+    }
```
timharrison-moj, line 453: I think this is assuming that the slots will always be removed and recreated for every patch? Even if the slots are not changing?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Activity.kt
```diff
@@ -73,23 +73,27 @@ data class Activity(
...
 
   @OneToMany(mappedBy = "activity", fetch = FetchType.EAGER, cascade = [CascadeType.ALL], orphanRemoval = true)
   @Fetch(FetchMode.SUBSELECT)
-  private val activityMinimumEducationLevel: MutableList<ActivityMinimumEducationLevel> = mutableListOf()
+  private var activityMinimumEducationLevel: MutableList<ActivityMinimumEducationLevel> = mutableListOf()
 
   @OneToMany(mappedBy = "activity", fetch = FetchType.EAGER, cascade = [CascadeType.ALL], orphanRemoval = true)
   @Fetch(FetchMode.SUBSELECT)
-  private val activityPay: MutableList<ActivityPay> = mutableListOf()
+  private var activityPay: MutableList<ActivityPay> = mutableListOf()
```
matt-crowson-moj, line 88: Most if not all of the lists on our entities are already mutable so should be no need to make these '`vars`'

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityService.kt
```diff
@@ -220,4 +223,233 @@ class ActivityService(
...
+    pay.forEach {
+      activity.addPay(
+        incentiveNomisCode = it.incentiveNomisCode!!,
+        incentiveLevel = it.incentiveLevel!!,
+        payBand = prisonPayBands[it.payBandId]
+          ?: throw IllegalArgumentException("Pay band not found for prison '$prisonCode'"),
+        rate = it.rate,
+        pieceRate = it.pieceRate,
+        pieceRateItems = it.pieceRateItems,
+      )
```
timharrison-moj, line 435: Assuming pay will always be removed and recreated? If nullable in request, this could be more selective?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Activity.kt
```diff
@@ -160,6 +164,10 @@ data class Activity(
     )
   }
 
+  fun removePay() {
+    activityPay.removeAll(activityPay)
```
matt-crowson-moj, line 168: Can we not just use the clear() function on the mutable lists?

`mutableEntityList.clear()`

src/test/resources/__files/activity/activity-update-request-2.json
```diff
@@ -0,0 +1,3 @@
+{
+  "categoryId": 2
```
timharrison-moj, line 2: These requests do suggest that all of the request fields are all nullable... so maybe a couple of my questions above will be easy to answer!

timharrison-moj (commented):
LGTM. A few questions added to consider only.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/resource/ActivityController.kt
```diff
@@ -235,4 +237,63 @@ class ActivityController(
...
+  @PatchMapping(value = ["/{prisonCode}/activityId/{activityId}"])
+  @Operation(
+    summary = "Update an activity",
+    description = "Update an activity. Requires any one of the following roles ['ACTIVITY_HUB', 'ACTIVITY_HUB_LEAD', 'ACTIVITY_ADMIN'].",
+  )
+  @ApiResponses(
+    value = [
+      ApiResponse(
+        responseCode = "202",
+        description = "The activity was updated.",
```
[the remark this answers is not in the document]
gerryallan-moj: Agreed with Tim to leave as 202 for now

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityService.kt
gerryallan-moj: This is just for the create where the location is always supplied

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityService.kt
gerryallan-moj: :-)

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityService.kt
gerryallan-moj: See below

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityService.kt
gerryallan-moj: :-)

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityService.kt
gerryallan-moj: We only do this if slots are supplied in the patch

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Activity.kt
gerryallan-moj: Changed back to val

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityService.kt
gerryallan-moj: We only do this if pay is supplied in the patch

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Activity.kt
gerryallan-moj: Updated as suggested

src/test/resources/__files/activity/activity-update-request-2.json
gerryallan-moj: Yes, we only need to supply updated fields or lists to the patch.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityService.kt
matt-crowson-moj: Sorry I am curious on this one.  When you say unreliable, in what way?  We do use this pattern quite a bit.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Activity.kt
```diff
@@ -89,7 +93,7 @@ data class Activity(
 
   @OneToMany(mappedBy = "activity", fetch = FetchType.EAGER, cascade = [CascadeType.ALL], orphanRemoval = true)
   @Fetch(FetchMode.SUBSELECT)
-  private val schedules: MutableList<ActivitySchedule> = mutableListOf()
+  private var schedules: MutableList<ActivitySchedule> = mutableListOf()
```
matt-crowson-moj, line 96: Still a `var`?

PR #327, opened 2023-05-12
SAA-654: Bug fix + test

leejacobson-moj:
## Overview

Fixes bug causing `otherAbsenceReason` to be recorded as null in attendance history


matt-crowson-moj (approved):
LGTM

PR #328, opened 2023-05-12
SAA-529 refactoring handlers so they return an Outcome instead of a boolean.

matt-crowson-moj:
This change allows for a more fluent api approach when an event handler is called/used.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/events/InboundEventsService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/events/InboundEventsService.kt
@@ -17,24 +17,16 @@ class InboundEventsService(
     private val log: Logger = LoggerFactory.getLogger(this::class.java)
   }
 
-  fun process(inboundEvent: InboundEvent) {
-    when (inboundEvent) {
-      is OffenderReceivedEvent -> {
-        if (!receivedEventHandler.handle(inboundEvent)) {
-          interestingEventHandler.handle(inboundEvent)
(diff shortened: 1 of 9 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/events/InboundEventsService.kt
```diff
@@ -17,24 +17,16 @@ class InboundEventsService(
...
+    when (event) {
+      is OffenderReceivedEvent -> receivedEventHandler.handle(event).onFailure { interestingEventHandler.handle(event) }
+      is OffenderReleasedEvent -> releasedEventHandler.handle(event).onFailure { interestingEventHandler.handle(event) }
+      is IncentivesInsertedEvent -> interestingEventHandler.handle(event)
+      is IncentivesUpdatedEvent -> interestingEventHandler.handle(event)
+      is IncentivesDeletedEvent -> interestingEventHandler.handle(event)
+      is CellMoveEvent -> interestingEventHandler.handle(event)
+      is NonAssociationsChangedEvent -> interestingEventHandler.handle(event)
+      else -> log.warn("Unsupported event ${event.javaClass.name}")
     }
```
timharrison-moj, line 30: From this mornings chat I think it might be worth storing as interesting events whether they fail or not?

matt-crowson-moj: Yeah I noticed this mentioned at the meet.  I had already started this ....

PR #330, opened 2023-05-15
SAA-712 digital prison reporting user

timharrison-moj:
This PR creates a new read-only group in the DEV database, creates a `digital_prison_reporting` user and adds them into the group with the user and password supplied from K8s secrets and fed into Flyway as property placeholders. The configuration is a bit tricky but ...

* The migration should only run against the DEV DB - its in the `/dev` migration folder.
* The env vars are not required for running tests against the H2 database - the migration does not apply to test profiles.
* When running locally with progresql, we do want the migration to run and the env vars to be set.


iterative-it (approved):
LGTM. Great to understand the process

PR #333, opened 2023-05-16
SAA-529 enabling feature for activities changed event and adding localstack example change message.

matt-crowson-moj:


timharrison-moj (approved):
LGTM

PR #336, opened 2023-05-17
Ensure that prisoner search api calls are only made if at least one prison number is supplied

iterative-it:
The prisoner search `/prisoner-search/prisoner-numbers` will return 400 bad request if called with no prisoner numbers. That can happen in a few cases with appointments when occurrences are deleted. Rather than check prior to making the call, the logic to prevent that exception is in the client itself. This is to prevent code duplication and missed checks


matt-crowson-moj (approved):
LGTM

PR #338, opened 2023-05-18
SAA-668 updating the manage attendnaces job to check for attendance records that need locking down.

matt-crowson-moj:
This covers the work for the overnight JOB which manages attendances, their expiry period and whether they are LOCKED to further edits in the activities service.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Attendance.kt
```diff
@@ -47,7 +47,7 @@ data class Attendance(
   var recordedBy: String? = null,
 
   @Enumerated(EnumType.STRING)
-  var status: AttendanceStatus = AttendanceStatus.WAITING,
+  private var status: AttendanceStatus = AttendanceStatus.WAITING,
```
matt-crowson-moj, line 50: This should not be directly mutable.  We should be calling methods/functions on the entity to transition the status states.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/integration/ManageAttendanceRecordsJobIntegrationTest.kt
```diff
@@ -95,7 +97,7 @@ class ManageAttendanceRecordsJobIntegrationTest : IntegrationTestBase() {
 
   @Sql("classpath:test_data/seed-activity-id-5.sql")
   @Test
-  fun `No attendance records are created for gym induction AM`() {
+  fun `No attendance records are created for gym induction AM when attendance not required on activity`() {
```
matt-crowson-moj, line 100: Added a bit more context to the test description.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/AttendancesServiceTest.kt
```diff
@@ -161,16 +155,16 @@ class AttendancesServiceTest {
...
-    assertThat(attendance.status).isEqualTo(AttendanceStatus.COMPLETED)
+    assertThat(attendance.status()).isEqualTo(AttendanceStatus.COMPLETED)
     with(attendance.attendanceReason!!) {
       assertThat(code).isEqualTo(AttendanceReasonEnum.OTHER)
     }
     assertThat(attendance.otherAbsenceReason).isEqualTo("other absence reason")
   }
 
   @Test
-  fun `other absence reason is not set when attendance reason does not equal "OTHER"`() {
```
matt-crowson-moj, line 172: Double quotes in the test description was throwing a compilation warning.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ManageAttendancesService.kt
```diff
@@ -76,8 +83,51 @@ class ManageAttendancesService(
...
+            when {
+              dateOfAttendance.between(yesterday.minusWeeks(2), yesterday) -> {
+                unlockedAttendance.takeIf { it.status(AttendanceStatus.COMPLETED) }
+              }
+
+              else -> unlockedAttendance
+            }?.let {
+              runCatching {
+                attendanceRepository.saveAndFlush(unlockedAttendance.lock())
+              }.onFailure {
```
timharrison-moj, line 116: Are we going to run into the same issues I have had? i.e. This job is updating attendance to LOCKED, and will emit the same sync event as attendance-updated. Maybe that is ok? Mike will probably throw these away - as the only change will be the paylock flag in NOMIS attendance, and that should already have been set.

timharrison-moj (approved):
LGTM. Tests look to confirm the correct behaviour - but I wasn't sure from looking at the job code.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ManageAttendancesService.kt
matt-crowson-moj: The only entities we are touching here are attendances so we should only ever get the attendance updated event.  I'll double check this though.

PR #345, opened 2023-05-23
SAA-609: Preprod circle and helm values.

timharrison-moj:
Enabling the preprod deployment (API)

### DIFF
--- a/helm_deploy/values-preprod.yaml
+++ b/helm_deploy/values-preprod.yaml
@@ -10,8 +10,10 @@ generic-service:
     PRISON_API_URL: https://api-preprod.prison.service.justice.gov.uk
     PRISONER_SEARCH_API_URL: https://prisoner-offender-search-preprod.prison.service.justice.gov.uk
     CASE_NOTES_API_URL: https://preprod.offender-case-notes.service.justice.gov.uk
-    APPLICATIONINSIGHTS_CONFIGURATION_FILE: applicationinsights.dev.json
-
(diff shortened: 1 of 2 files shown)


helm_deploy/values-preprod.yaml
```diff
@@ -10,8 +10,10 @@ generic-service:
     PRISON_API_URL: https://api-preprod.prison.service.justice.gov.uk
     PRISONER_SEARCH_API_URL: https://prisoner-offender-search-preprod.prison.service.justice.gov.uk
     CASE_NOTES_API_URL: https://preprod.offender-case-notes.service.justice.gov.uk
-    APPLICATIONINSIGHTS_CONFIGURATION_FILE: applicationinsights.dev.json
-
+    APPLICATIONINSIGHTS_CONFIGURATION_FILE: applicationinsights.json
+    SCHEDULE_AHEAD_DAYS: 14
+    FEATURE_EVENTS_SNS_ENABLED: true
```
matt-crowson-moj, line 15: I wonder if lines 15 and 16 can actually just go in the default one?  We are overriding them in with the same values in dev as well so may as well perhaps?

matt-crowson-moj (approved):
LGTM

helm_deploy/values-preprod.yaml
timharrison-moj: Dunno. They will be different in prod, and might be different in preprod - when we start using it.

PR #348, opened 2023-05-24
SAA-433: Cater for cancellations when creating attendance records

leejacobson-moj:
## Overview

Adds support for creating attendance records for scheduled instances which have been cancelled

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ManageAttendancesService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ManageAttendancesService.kt
@@ -60,24 +60,33 @@ class ManageAttendancesService(
 
   private fun createAttendanceRecordIfNoPreExistingRecord(instance: ScheduledInstance, allocation: Allocation) {
     if (!attendanceAlreadyExistsFor(instance, allocation)) {
-      val attendance = allocation
-        .takeIf { it.status(PrisonerStatus.AUTO_SUSPENDED, PrisonerStatus.SUSPENDED) }
-        ?.let {
-          Attendance(
-            scheduledInstance = instance,
-            prisonerNumber = allocation.prisonerNumber,
-            attendanceReason = attendanceReasonRepository.findByCode(AttendanceReasonEnum.SUSPENDED),
-            payAmount = instance.rateFor(allocation.payBand),
-            issuePayment = false,
-            status = AttendanceStatus.COMPLETED,
-            recordedTime = LocalDateTime.now(),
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ManageAttendancesService.kt
```diff
@@ -73,10 +73,23 @@ class ManageAttendancesService(
...
+            attendanceReason = attendanceReasonRepository.findByCode(AttendanceReasonEnum.NOT_REQUIRED),
+            recordedTime = LocalDateTime.now(),
+            recordedBy = ServiceName.SERVICE_NAME.value,
+          )
         } ?: Attendance(
         scheduledInstance = instance,
         prisonerNumber = allocation.prisonerNumber,
         payAmount = instance.rateFor(allocation.payBand),
+        status = if (instance.cancelled) AttendanceStatus.COMPLETED else AttendanceStatus.WAITING,
       )
```
timharrison-moj, line 93: Is the `instance.takeIf { it.cancelled }` already filtering just to cancelled? So is the `status =  if ( cancelled) ` necessary?

leejacobson-moj: Yes, you're right – thanks for spotting this Tim. I was suppose to remove this line.

timharrison-moj (dismissed):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ManageAttendancesService.kt
```diff
@@ -73,6 +73,18 @@ class ManageAttendancesService(
             recordedTime = LocalDateTime.now(),
             recordedBy = ServiceName.SERVICE_NAME.value,
           )
+        } ?: instance.takeIf { it.cancelled }
```
matt-crowson-moj, line 76: I am wondering if this block of code (now it has grown) might be easier to read in a when statement? For example

```
      val attendance = when {
        allocation.status(PrisonerStatus.AUTO_SUSPENDED, PrisonerStatus.SUSPENDED) -> Attendance(...)
        instance.cancelled -> Attendance(...)
        else -> Attendance(...)
      }

```

timharrison-moj: I agree with this - despite just approving - it did take me a while to look through that particular block of code.

timharrison-moj (approved):
LGTM

PR #350, opened 2023-05-25
SAA-607: Sets an appropriate summary value for appointment events

leejacobson-moj:
## Overview

Minor change to return a more useful summary for appointment events


iterative-it (approved):
LGTM - nice test work!

PR #352, opened 2023-05-26
SAA-609: Updates to logback for app insights changes

timharrison-moj:
Recommended changes - from others with the same issues.


yasinmustafa (approved):
lgtm

PR #356, opened 2023-05-30
SAA-678: Include case note text in attendance history

gerryallan-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/util/TransformFunctions.kt
```diff
@@ -341,6 +342,7 @@ fun transform(attendanceHistory: EntityAttendanceHistory): ModelAttendanceHistor
         notes = it.notes,
       )
     },
+    caseNoteText = attendanceHistory.caseNoteId?.let { caseNotesApiClient.getCaseNote(prisonerNumber, attendanceHistory.caseNoteId)?.text },
     comment = attendanceHistory.comment,
```
timharrison-moj, line 346: I'm not sure about this change, Gerry..  feels a bit off to pass an API client down into the transform functions to make the call to get case notes from there?  For me, these calls feel better being made in a service method, and passed into the transform functions as pieces of data, but I'll ask Matt for a second opinion and get back to you.

matt-crowson-moj: Yes agreed.  Should really only be using the api clients in services/component stereotypes.

timharrison-moj: Yeah, Matt agrees with me - sorry!  Is there any way the calls to the case note service could stay within the service methods, and passed as data into the transforms instead?

gerryallan-moj: Reworked as suggested

matt-crowson-moj (approved):
LGTM

PR #358, opened 2023-05-31
SAA-763 planned allocation deallocations should be considered when the deallocation job runs.

matt-crowson-moj:
Updates to ensure deallocation job factors in planned deallocations in the job which deallocates and ends allocations that have an end date (or planned end date).


timharrison-moj (approved):
LGTM

PR #359, opened 2023-06-01
SAA-762 attendance locking and expiring

timharrison-moj:
This PR implements the following decisions agreed with Mike @ Syscon:
* Removal of our Attendance status for LOCKED - we no longer need it - NOMIS is responsible for deciding when to pay-lock.
* Adding a boolean `editable` flag to our attendance entity and model - and rules to set it.
* Adding a new event `activities.prisoner.attendance-expired` to mimic the NOMIS flush-schedules job for sync.
* Replacing the LOCK attendance job with an EXPIRE equivalent.
* Tests for editable attendance and expiry job
* Feature switches.


matt-crowson-moj (dismissed):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Attendance.kt
```diff
@@ -195,11 +211,11 @@ data class Attendance(
...
       .reversed()
       .map { attendanceHistoryRow -> transform(attendanceHistoryRow) },
+    editable = this.editable(),
   )
 }
 
 enum class AttendanceStatus {
   WAITING,
   COMPLETED,
-  LOCKED,
```
leejacobson-moj, line 204: Minor but I think the schema description in `AttendanceUpdateRequest` lists "LOCKED" as a status value. Might want to remove that.

timharrison-moj: Good spot - I did search for it, so wondering how I missed it! I'll take it out.

matt-crowson-moj (approved):

PR #362, opened 2023-06-01
SAA-780 the reason codes presented to the UI need to be restricted to a subset of codes.

matt-crowson-moj:
Some of the deallocation reason codes are only applicable to the API e.g. when jobs run.


timharrison-moj (approved):
LGTM

PR #363, opened 2023-06-01
SAA-601 - Update allocation status on start date

sioldham:
Alongside the unit tests I have also performed the following manual test

- Create and activity via the UI with a start date of today
- Allocate a prisoner to that activity
- Update the prisoner status in the DB to `PENDING` (it's not possible to set a future start date via the UI)
- Invoke the `job/manage-allocations` endpoint via swagger
- Verify that the prisoner status is now `ACTIVE`


timharrison-moj (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Allocation.kt
```diff
@@ -152,6 +151,9 @@ data class Allocation(
       status = prisonerStatus,
     )
 
+  fun activate() =
+    this.apply { prisonerStatus = PrisonerStatus.ACTIVE }
```
matt-crowson-moj, line 155: Could possibly add a check (IllegalStateException?) to this to make sure it can be activated?

You can only activate PENDING for example??

matt-crowson-moj (approved):
LGTM

PR #381, opened 2023-06-09
SAA-838 job was trying to end already ended allocations hence it was failing.

matt-crowson-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ManageAllocationsService.kt
```diff
@@ -70,7 +69,7 @@ class ManageAllocationsService(
         .flatMap { prison ->
           activityRepository.getAllForPrisonAndDate(prison.code, today).flatMap { activity ->
             if (activity.ends(today)) {
-              activity.schedules().flatMap { it.allocations() }
+              activity.schedules().flatMap { it.allocations().filterNot(Allocation::isEnded) }
```
matt-crowson-moj, line 72: This is the crux of the change.

This was including all allocation on an activity and trying to end them, even if they were already ended.

PR #382, opened 2023-06-09
Return activities with the option to exclude archived activities

smcveigh941:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Activity.kt
```diff
@@ -270,16 +272,16 @@ data class Activity(
...
 
   @Override
   override fun toString(): String {
     return this::class.simpleName + "(activityId = $activityId )"
   }
 
-  private fun getActivityState(endDate: LocalDate?): ActivityState {
-    return if (endDate != null && endDate < LocalDate.now()) {
+  private fun getActivityState(): ActivityState {
+    return if (endDate != null && endDate!! < LocalDate.now()) {
```
matt-crowson-moj, line 284: I think this should be <= e.g.

endDate!! <= LocalDate.now()

PR #385, opened 2023-06-12
SAA-773 - Do not overwrite status of suspended prisoners on cancel/uncancel

sioldham:
Alongside the unit tests I have also performed the following manual test

- Create an activity via the UI with a start date of today
- Allocate three prisoners to that activity
- Set the status of one of the prisoners to SUSPENDED.
- Generate the attendance records using swagger (one of them now has an attendance reason of 7 - SUSPENDED)
- Cancel the session. The attendance records for the unsuspended prisoners have their reason set to 8 - WAITING. The attendance record for the suspended prisoner still has its attendance reason set to 7 and the prisoner attendance and pay is recorded as **SUSPENDED NO PAY**
- Uncancel the session. The attendance records for the unsuspended prisoners have their reason set to null. The attendance record for the suspended prisoner still has its attendance reason set to 7

image


matt-crowson-moj (approved):
LGTM

PR #386, opened 2023-06-13
SAA-839 work to improve resilience of our jobs and record their success or failure in the DB should we need to recover/re-run.

matt-crowson-moj:
These changes do not guarantee jobs from failing but it does mean any jobs that run in the same call will get the opportunity to run should a previous one fail.  It also records the success/failure of the job in the DB.

This can be expanded on - thoughts are to provide some information to the Job entity that gets created with some details of the job outcome.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Job.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Job.kt
@@ -18,11 +18,12 @@ data class Job(
 
   @Enumerated(EnumType.STRING)
   val jobType: JobType,
+
+  val startedAt: LocalDateTime,
 ) {
-  var startedAt: LocalDateTime = LocalDateTime.now()
-    private set
 
   var endedAt: LocalDateTime? = null
+    private set
 
   var successful: Boolean = false
     private set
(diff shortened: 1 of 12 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/job/ManageAllocationsJob.kt
```diff
@@ -1,26 +1,34 @@
...
+          service.allocations(AllocationOperation.STARTING_TODAY)
+        },
+      )
     }
 
     if (withDeallocate) {
-      service.allocations(AllocationOperation.DEALLOCATE_ENDING)
-      service.allocations(AllocationOperation.DEALLOCATE_EXPIRING)
+      safeJobRunner.runSafe(
+        JobDefinition(jobType = JobType.DEALLOCATION) {
```
matt-crowson-moj, line 27: Should these be split into two calls?  If the first fails then the second won't run for example.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/job/SafeJobRunner.kt
```diff
@@ -0,0 +1,33 @@
...
+    val job = Job.start(jobDefinition.jobType)
+
+    runCatching {
+      jobDefinition.block()
+    }
+      .onSuccess { job.succeeded() }
+      .onFailure { log.error("Failed to run ${jobDefinition.jobType} job", it) }
+
+    jobRepository.saveAndFlush(job)
+  }
```
timharrison-moj, line 30: Very neat.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/job/SafeJobRunnerTest.kt
```diff
@@ -0,0 +1,85 @@
...
+  @Test
+  fun `runs safe job with error`() {
+    runner.runSafe(JobDefinition(JobType.ATTENDANCE) { throw RuntimeException("it failed") })
+
+    verify(jobRepository).saveAndFlush(jobEntityCaptor.capture())
+
+    with(jobEntityCaptor.firstValue) {
+      assertThat(endedAt).isNull()
+      assertThat(successful).isFalse
+    }
```
timharrison-moj, line 63: Aha - unsuccessful test here.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Job.kt
```diff
@@ -0,0 +1,62 @@
...
+
+  companion object {
+    fun successful(jobType: JobType, start: LocalDateTime) = Job(jobType = jobType, startedAt = start).succeeded()
+
+    fun failed(jobType: JobType, start: LocalDateTime) = Job(jobType = jobType, startedAt = start).failed()
+  }
+}
+
+enum class JobType {
+  ALLOCATION,
```
matt-crowson-moj, line 57: We may want a finer granularity here.  This does not go down to the different types of dellocations for example.

PR #392, opened 2023-06-15
SAA-839 allow dependent jobs to be grouped so following job(s) only run if the previous succeeds.

matt-crowson-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/job/MorningJobs.kt
```diff
@@ -0,0 +1,25 @@
...
+import org.springframework.scheduling.annotation.Async
+import org.springframework.stereotype.Component
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.entity.JobType
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.service.AllocationOperation
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.service.AttendanceOperation
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.service.ManageAllocationsService
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.service.ManageAttendancesService
+
+@Component
+class MorningJobs(
```
matt-crowson-moj, line 12: This for example purposes only and is not actually used anywhere.

The attendance job will only run if the previous allocation pending runs successfully.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/job/SafeJobRunner.kt
```diff
@@ -14,14 +15,36 @@ class SafeJobRunner(private val jobRepository: JobRepository) {
...
+    jobDefinitions.forEach { job ->
+      if (success.get()) {
+        runSafe(job)
+          .onFailure {
+            success.set(false)
+
+            log.warn("Failure occurred running job ${job.jobType}")
+          }
+      } else {
+        log.warn("Ignoring job ${job.jobType} due to failure in a dependent job.")
```
matt-crowson-moj, line 39: Would we still want to record the fact in the DB this job did not run?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/job/MorningJobs.kt
```diff
@@ -0,0 +1,25 @@
...
+  private val jobRunner: SafeJobRunner,
+) {
+
+  @Async("asyncExecutor")
+  fun execute() {
+    jobRunner.runDependentJobs(
+      JobDefinition(JobType.ALLOCATE) { service.allocations(AllocationOperation.STARTING_TODAY) },
+      JobDefinition(JobType.ATTENDANCE_CREATE) { attendancesService.attendances(AttendanceOperation.CREATE) },
+    )
+  }
```
matt-crowson-moj, line 24: A separate call to a non-dependent job for creation of scheduled instances could go here at the end ...

PR #393, opened 2023-06-15
SAA-736: Keep instances in sync with changes to activity start and end dates

gerryallan-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityService.kt
```diff
@@ -315,7 +317,14 @@ class ActivityService(
   ) {
     request.startDate?.apply {
       activity.startDate = this
-      activity.schedules().forEach { it.startDate = this }
+      activity.schedules().forEach {
+        if (it.startDate < this) {
```
matt-crowson-moj, line 321: I'd find it useful if would could put some of these conditions into functions with intention reveling names.

I am struggling a little to follow the rules.

timharrison-moj: Yes, I agree - either function or a comment to describe how the from/to dates, when not null, affect the instances being created?

timharrison-moj (commented):
LGTM, just a comment on clarifying the start/end date change actions?

matt-crowson-moj (approved):
LGTM - I still think we have some holes in the process but probably better to get this merged and discuss.

PR #396, opened 2023-06-16
SAA-846: Add get bulk appointment details endpoint for appointment set view

iterative-it:
- Adds additional columns to bulk_appointments table via a migration
- Adds properties to BulkAppointment JPA data class
- Switches all JPA entity relationships to lazy loading
- Added @Transaction attribute to services


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/BulkAppointment.kt
```diff
@@ -10,35 +10,99 @@ import jakarta.persistence.JoinColumn
...
+      } else {
+        locationMap[internalLocationId].toAppointmentLocationSummary(internalLocationId!!, prisonCode)
+      },
+      inCell,
+      startDate,
+      appointments().map { it.occurrenceDetails(prisonerMap, referenceCodeMap, locationMap, userMap) }.flatten(),
+      created,
+      userMap[createdBy].toSummary(createdBy),
+    )
+  }
```
timharrison-moj, line 107: So the idea has been to make collections lazily loaded and, depending on whether the caller wants summary, model or detail, the other collections are loaded as part of the transform? As most calls will be focussed on the summary or model transformations,  the overall number of JDBC statements should drop? About right? The worst case would be a long-running bulk appointment with lots of people allocated .. so the details model will always be an expensive operation.

iterative-it: Yep, spot on. The details views are the heaviest in terms of total number of queries however those are only loaded when users want to manage the appointments. For the various search and summary views, which is the largest percentage of daily use, this change prevents the entire object graph being eagerly loaded

timharrison-moj: Just added a comment on @NamedEntityGraph - which could significantly reduce the numbers of JDBC statements prepared and issued for detailed views - but could be a step if or when time allows - just tuning queries.

timharrison-moj (approved):
A lot of work! All LGTM.
I'll leave the decision about risk and size of changes to you - if you're comfortable and its tested locally, I don't see any issues.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/resource/BulkAppointmentDetailsController.kt
```diff
@@ -0,0 +1,69 @@
...
+
+@RestController
+@RequestMapping("/bulk-appointment-details", produces = [MediaType.APPLICATION_JSON_VALUE])
+class BulkAppointmentDetailsController(
+  private val bulkAppointmentDetailsService: BulkAppointmentDetailsService,
+) {
+  @GetMapping(value = ["/{bulkAppointmentId}"])
+  @ResponseBody
+  @Operation(
+    summary = "Bulk create a set of appointments",
```
matt-crowson-moj, line 26: Is the summary and and description correct here?  This looks like a GET endpoint and the text refers to creating ...

iterative-it: Damn. Didn't see this till after I'd merged! I'll fix this in another PR

PR #401, opened 2023-06-16
SAA-543 hibernate filters on large collections

timharrison-moj:
This PR is draft, for discussion.

It includes:

* Hibernate filters defined to limit the size of large collection `instances`
* LAZY loading strategy for instances collection requiring @Transactional service methods.
* The `allocations` collection has been left as `FetchType.EAGER` for now - perhaps a next step?
* The instances filter accept a parameter from the client -  `earliestSessionDate`  - which defaults to 1 month if not provided.
* An example endpoint providing the filter parameters - `/activity/{activityId}/filtered?earliestSessionDate=2023-03-03`
* Implements an `ActivityBasic` projection - for use where the full Activity is not required i.e. check exists, get schedule id.
* Changes to the create instances job to use `ActivityBasic` and filtered instances - does not need the full collection (tested)

Next steps:
* Analyse which client methods can be switched to use filtered instances or avoid the instances collection.
* Test the client changes - activities dashboard, attendance, allocation.
* Other jobs - ManageAllocations, ManageAttendance might be more efficient using the basic service calls.
* Ensure sync is getting the same events.

This has been updated to reflect latest thoughts:
https://dsdmoj.atlassian.net/wiki/spaces/SAA/pages/4426957457/JPA+Entities+with+potentially+large+collections


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/repository/ActivityRepositoryCustom.kt
```diff
@@ -0,0 +1,47 @@
...
+import jakarta.persistence.PersistenceException
+import jakarta.persistence.TypedQuery
+import org.hibernate.Session
+import org.slf4j.LoggerFactory
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.entity.Activity
+import java.time.LocalDate
+import java.util.Optional
+
+interface ActivityRepositoryCustom {
+  fun getActivityByIdWithFilters(activityId: Long, earliestSessionDate: LocalDate): Optional<Activity>
```
matt-crowson-moj, line 14: Can we avoid using Optional here and just use Kotlins nullable return type?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/repository/ActivityRepositoryCustom.kt
```diff
@@ -0,0 +1,47 @@
...
+    // Enable the session date filter to limit the scheduled instances returned
+    log.info("Enabling filter SessionDateFilter with earliestSessionDate: $earliestSessionDate")
+    val sessionDateFilter = session.enableFilter("SessionDateFilter")
+    sessionDateFilter.setParameter("earliestSessionDate", earliestSessionDate)
+
+    val hql = "select a from Activity a where a.activityId = :activityId"
+    val query: TypedQuery<Activity> = entityManager.createQuery(hql, Activity::class.java)
+    query.setParameter("activityId", activityId)
+
+    try {
```
matt-crowson-moj, line 38: Suggestion if return nullable type

```
    return runCatching {
      query.singleResult
    }.onFailure { log.error("Activity by ID failed ${it.message}") }
      .getOrNull()

```

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/resource/ActivityController.kt
```diff
@@ -90,6 +93,172 @@ class ActivityController(
   fun getActivityById(@PathVariable("activityId") activityId: Long): Activity =
     activityService.getActivityById(activityId)
```
matt-crowson-moj, line 94: I wonder if we should remove this, deprecate it or under the covers all the filtered method?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/ActivitySchedule.kt
```diff
@@ -26,6 +30,12 @@ import java.time.temporal.ChronoUnit
 @Entity
 @Table(name = "activity_schedule")
 @EntityListeners(ActivityScheduleEntityListener::class)
+@FilterDef(
+  name = "SessionDateFilter",
```
matt-crowson-moj, line 34: Suggest pulling the filter name out to a constant, the constanct can then be referenced in the repo e.g.

```
const val SESSION_DATE_FILTER = "SessionDateFilter"

@Entity
@Table(name = "activity_schedule")
@EntityListeners(ActivityScheduleEntityListener::class)
@FilterDef(
  name = SESSION_DATE_FILTER,
  parameters = [
    ParamDef(name = "earliestSessionDate", type = LocalDate::class),
  ],
)
data class ActivitySchedule(
```

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityService.kt
```diff
@@ -46,15 +51,30 @@ class ActivityService(
...
   @Value("\${online.create-scheduled-instances.days-in-advance}") private val daysInAdvance: Long = 14L,
 ) {
-
   companion object {
     val log: Logger = LoggerFactory.getLogger(this::class.java)
   }
 
-  fun getActivityById(activityId: Long) =
-    transform(
-      activityRepository.findOrThrowNotFound(activityId),
```
matt-crowson-moj, line 56: So do we not want to use the extension function anymore?  I ask because changed in some places but not others.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/ActivitySchedule.kt
```diff
@@ -35,12 +45,7 @@ data class ActivitySchedule(
   @JoinColumn(name = "activity_id", nullable = false)
   val activity: Activity,
 
-  @OneToMany(
-    mappedBy = "activitySchedule",
-    fetch = FetchType.EAGER,
-    cascade = [CascadeType.ALL],
-    orphanRemoval = true,
-  )
+  @OneToMany(mappedBy = "activitySchedule", cascade = [CascadeType.ALL], orphanRemoval = true)
```
iterative-it, line 48: While the default for OneToMany is lazy, it's not the default for all mappings. ManyToOne and OneToOne are eager. It might be more readable to add fetch lazy here to show that is the intent

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/ActivitySchedule.kt
```diff
@@ -74,30 +79,15 @@ data class ActivitySchedule(
...
 
-  @OneToMany(
-    mappedBy = "activitySchedule",
-    fetch = FetchType.EAGER,
-    cascade = [CascadeType.ALL],
-    orphanRemoval = true,
-  )
-  @Fetch(FetchMode.SUBSELECT)
+  @OneToMany(mappedBy = "activitySchedule", fetch = FetchType.LAZY, cascade = [CascadeType.ALL], orphanRemoval = true)
+  @Filters(Filter(name = "SessionDateFilter", condition = "session_date >= :earliestSessionDate"))
```
iterative-it, line 83: Especially as it's done here

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/projections/ActivityBasic.kt
```diff
@@ -0,0 +1,12 @@
+package uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.entity.projections
+
+import java.time.LocalDate
+
+data class ActivityBasic(
+  val prisonCode: String,
```
iterative-it, line 6: I've gone with the convention of EntitySummary for the non complete version of the entity and EntityDetails for the fully hydrated, UI focused version. The EntityModel sits between and is the full object graph but ids only. Don't know if it's worth aligning activities with that.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/repository/ActivityRepository.kt
```diff
@@ -31,4 +32,54 @@ interface ActivityRepository : JpaRepository<Activity, Long> {
...
+    """
+    SELECT 
+       new uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.entity.projections.ActivityBasic(
+            a.prisonCode, a.activityId, sched.activityScheduleId, a.summary, a.startDate, a.endDate
+       )
+    FROM Activity a JOIN a.schedules sched
+    WHERE a.activityId = :activityId
+    """,
+  )
+  fun getActivityBasicById(@Param("activityId") activityId: Long): ActivityBasic
```
iterative-it, line 48: Possibly change to using a view to reduce the duplication of the query in these repository methods?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/resource/ActivityController.kt
```diff
@@ -90,6 +93,172 @@ class ActivityController(
...
+    ],
+  )
+  fun getActivityByIdWithFilters(
+    @PathVariable("activityId") activityId: Long,
+    @RequestParam(value = "earliestSessionDate", required = false)
+    @Parameter(description = "The date of the earliest scheduled instances to include. Defaults to newer than 1 month ago.")
+    earliestSessionDate: LocalDate?,
+  ) = activityService.getActivityByIdWithFilters(activityId, earliestSessionDate)
+
+  @GetMapping(value = ["/{activityId}/basic"])
```
iterative-it, line 153: With appointments, I've tended to return the Summary/Lite/Basic version as the default then have /details as the specific endpoint. The idea is to encourage users to think about whether they need the more expensive operation. Don't know if that's worth considering here

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityService.kt
```diff
@@ -46,15 +51,30 @@ class ActivityService(
...
-    )
+  fun getActivityByIdWithFilters(activityId: Long, earliestSessionDate: LocalDate?): ModelActivity {
+    val earliestSession = earliestSessionDate ?: LocalDate.now().minusMonths(1)
+    val activity = activityRepository.getActivityByIdWithFilters(activityId, earliestSession)
+      .orElseThrow { EntityNotFoundException("Activity $activityId not found") }
+    return transform(activity)
```
iterative-it, line 62: I think every entity conversion function for appointments is on the entity itself. Might need a separate ticket to move those into transform functions if that's the recommended approach

iterative-it (commented):
Some really good examples in here of all possible optimisation approaches. Particularly interested in the named graphs. I've highlighted areas where the activities and appointments approaches differ simply to promote conversations as to whether there should be a move to align. No expectation that anything in this PR is changed as it all looks good to mem

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/resource/ActivityController.kt
timharrison-moj: It is the plan - but I wanted to have both versions available and then start to switch the different client calls over gradually, testing each use-case.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/repository/ActivityRepositoryCustom.kt
timharrison-moj: Yes, I agree - I'll take out the Optional.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/repository/ActivityRepositoryCustom.kt
timharrison-moj: Yes - will go along these lines - didn't like the Optional or try/catch stuff.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityService.kt
timharrison-moj: I'd changed a number of areas to use override functions but then pulled back from it. These extension function have a non-obvious (at first viewing) effect - they extend JpaRepository so it always calls the specific method, ignoring if there are @override by ActivityRepository. Whilst debugging it was much clearer to see what was happening by being explicit. Sometimes, if convenience functions mask or alter behaviour I'm 50/50 on the benefit. 1 or 2 lines saved, or explicit and clear?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/repository/ActivityRepository.kt
timharrison-moj: I'll have a look now.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/repository/ActivityRepositoryCustom.kt
timharrison-moj: Done, thanks.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/repository/ActivityRepositoryCustom.kt
timharrison-moj: Done, thanks.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/ActivitySchedule.kt
timharrison-moj: Done, thanks.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/ActivitySchedule.kt
timharrison-moj: Done, thanks.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/projections/ActivityBasic.kt
timharrison-moj: The next couple of UI PRs will be about exploring what data is actually needed by the client and supplying only that. Specifically, where there are long lists of activities these could get very expensive in DB processing - areas like the activities , attendance and allocation dashboards. I want to see if there are better ways to provide that through views.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/resource/ActivityController.kt
timharrison-moj: I'm going to explore from the client side next, then decide how to assemble that data more efficiently.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityService.kt
timharrison-moj: I'll replace the ones I took out for consistency, and maybe add a note that it will not override a function of the same name in the repository.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/repository/ActivityRepository.kt
timharrison-moj: Added a view to reduce the repeated SQL and align with other views in the service (avoids the NEW constructor in JPSQL).

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/ActivitySchedule.kt
```diff
@@ -69,30 +76,15 @@ data class ActivitySchedule(
...
```
matt-crowson-moj, line 80: Small thing. You could use the constant value here for the name of the filter, meh.

timharrison-moj: Next PR? There will be a couple more to flip to using the filtered endpoint in different places.

matt-crowson-moj (approved):
LGTM

PR #402, opened 2023-06-19
SAA-678: display case notes on attendance history

gerryallan-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Attendance.kt
```diff
@@ -175,7 +176,7 @@ data class Attendance(
       )
   }
 
-  fun toModel(caseNoteText: String?) = ModelAttendance(
+  fun toModel(caseNotesApiClient: CaseNotesApiClient?) = ModelAttendance(
```
matt-crowson-moj, line 179: Didn't we have this before passing through the client?  Not ideal really.

Not really something that should be use in the Domain layer.

If we really want to use it would could be a little sneaky about it and use a function type.

```
fun toModel(caseNoteLookup: (String) -> String?) = = ModelAttendance(
  ...
)
```

gerryallan-moj: toModel has been refactored to transform function

matt-crowson-moj (approved):
LGTM

Not sure I understand why the casenotes api is nullable really.  I understand the same was in in place for case note text.  Just seems a bit odd.  And yes it is used/called elsewhere.

PR #405, opened 2023-06-19
SAA-868 changes to clean up schedule instances when changes are made to the schedules slots.

matt-crowson-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/PrisonRegime.kt
```diff
@@ -37,4 +38,10 @@ data class PrisonRegime(
 
   fun hasExpired(predicate: () -> LocalDate?) =
     predicate()?.onOrBefore(LocalDate.now().minusDays(maxDaysToExpiry.toLong())) == true
+
+  fun timeSlots() = mapOf(
```
matt-crowson-moj, line 42: Seemed logical to move this into the prison regime to save having to keep working it out in the service layer.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityService.kt
```diff
@@ -207,7 +203,9 @@ class ActivityService(
           (fromDate == null || day >= fromDate) &&
           (toDate == null || day <= toDate)
         ) {
-          this.addInstance(sessionDate = day, slot = slot)
+          if (this.hasNoInstancesOnDate(day)) {
```
matt-crowson-moj, line 206: I think we should be checking before adding an instance.

PR #407, opened 2023-06-20
SAA-876 allocations should only have attendance records created when in effect i.e. not pending or ended.

matt-crowson-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ManageAttendancesService.kt
```diff
@@ -57,7 +57,7 @@ class ManageAttendancesService(
   private fun List<ScheduledInstance>.andAttendanceRequired() = filter { it.attendanceRequired() }
 
   private fun ScheduledInstance.forEachInFlightAllocation(f: (allocation: Allocation) -> Unit) {
-    activitySchedule.allocations().filterNot { it.status(PrisonerStatus.ENDED) }.forEach { f(it) }
+    activitySchedule.allocations().filterNot { it.status(PrisonerStatus.PENDING, PrisonerStatus.ENDED) }.forEach { f(it) }
   }
```
timharrison-moj, line 61: This, I think, is right as we still need attendances for AUTO_SUSPENDED or SUSPENDED.

matt-crowson-moj: Indeed.  You will see this further down in the service where it checks the various states.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ManageAttendancesServiceTest.kt
```diff
@@ -92,9 +93,23 @@ class ManageAttendancesServiceTest {
     )
   }
 
+  @Test
+  fun `attendance record is not created when allocation is pending`() {
+    instance.activitySchedule.activity.attendanceRequired = true
+
```
timharrison-moj, line 99: Test using global data? But you're right not to refactor, as discussed.

timharrison-moj (approved):
LGTM.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ManageAttendancesServiceTest.kt
matt-crowson-moj: Kept changes to a minimum to avoid conflicts.  I am comfortable this is sufficient for now.

PR #408, opened 2023-06-20
SAA-874 start date of allocation cannot be before activity start date or after activity end date.

matt-crowson-moj:
Test case

Activity starting and ending on 22 June 2023.

image

image

image


timharrison-moj (approved):
LGTM. 

As an aside, there is some `walking the tree of objects` here, but was already there - so if I tale out the large collections `instances` and `allocations`  from ActivitySchedule, this area will need to be on the rework list too.

PR #413, opened 2023-06-21
SAA-871: Suitability endpoint updates

leejacobson-moj:


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/CandidatesServiceTest.kt
```diff
@@ -449,5 +461,31 @@ class CandidatesServiceTest {
...
+      candidateSuitabilitySetup(activity, candidate)
+
+      val suitability = service.candidateSuitability(
+        activity.schedules().first().activityScheduleId,
+        candidate.prisonerNumber,
+      )
+
+      val candidateAllocation = allocation().copy(allocationId = 1, prisonerNumber = "A1234BC")
+
+      assertThat(suitability.allocations).isEqualTo(
```
matt-crowson-moj, line 481: This is no biggy really.  There is an alternative assertion method for lists.  It would look something like this:

```
assertThat(suitability.allocations).containsOnly(
  AllocationPayRate(
    allocation = candidateAllocation.toModel(),
    payRate = candidateAllocation.allocationPay("BAS")?.toModelLite(),
  ),
)

There are various other permutations of `contains` but the above should work fine in this example.
```

timharrison-moj (commented):
LGTM (too).

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ManageAttendancesServiceTest.kt
```diff
@@ -146,6 +167,15 @@ class ManageAttendancesServiceTest {
 
     whenever(scheduledInstanceRepository.findAllBySessionDate(today)).thenReturn(listOf(instance))
     whenever(attendanceReasonRepository.findByCode(AttendanceReasonEnum.SUSPENDED)).thenReturn(attendanceReasons()["SUSPENDED"])
+    val attendees = instance.attendances.map { it.prisonerNumber }
+    whenever(prisonerSearchApiClient.findByPrisonerNumbers(attendees))
+      .thenReturn(
+        Mono.just(
+          attendees.map {
```
timharrison-moj, line 174: Question using Mono here? Response type from the search API is `List<Prisoner>` so not sure of reason for wrapping this in a Mono?

leejacobson-moj: `findByPrisonerNumbers()` returns a `Mono<List<Prisoner>>` hence the wrapping in a Mono here.

We have `findByPrisonerNumbersAsync()` which returns `List<Prisoner>`, but that's an async call so would require refactoring to use coroutines. It's also only used in one other place in the codebase.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/CandidatesServiceTest.kt
leejacobson-moj: Good to know – thank you. Updated.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ManageAttendancesServiceTest.kt
timharrison-moj: yeah, its ok - if other places use the Mono, but strictly speaking its not necessary.  ok for now.

timharrison-moj (approved):
LGTM

PR #416, opened 2023-06-23
SAA-890: Adds validation to attendance marking

leejacobson-moj:
## Overview

Simple change to add validation to attendance marking. Marking attendance that's not editable will now throw an exception.


timharrison-moj (approved):
LGTM

PR #419, opened 2023-06-25
Use a more consistent paginated endpoint for fetching all prisoners for activity candidates

smcveigh941:
* The previously used endpoint would have occasionally returned the list in a different order, which would have messed up the pagination on the frontend. This endpoint is more consistent


matt-crowson-moj (approved):
LGTM

PR #420, opened 2023-06-26
SAA-912: Return a default rollout plan for prisons which are not yet planned for rollout

smcveigh941:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/RolloutPrisonService.kt
```diff
@@ -1,14 +1,19 @@
...
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.entity.RolloutPrison
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.repository.RolloutPrisonRepository
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.util.transform
 
 @Service
 class RolloutPrisonService(private val repository: RolloutPrisonRepository) {
 
   fun getByPrisonCode(code: String) = transform(
-    repository.findByCode(code) ?: throw EntityNotFoundException(code),
+    repository.findByCode(code) ?: RolloutPrison(
```
matt-crowson-moj, line 12: Not sure I agree with this and even if we do say this is OK then the controller docs will need updating.

404 still feels logical to me though.

matt-crowson-moj (approved):
LGTM

PR #423, opened 2023-06-27
SAA-909: Update appointment name for scheduled events

leejacobson-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/util/TransformFunctions.kt
```diff
@@ -149,7 +149,10 @@ fun transformAppointmentInstanceToScheduledEvents(
     suspended = false,
     categoryCode = it.categoryCode,
     categoryDescription = referenceCodesForAppointmentsMap[it.categoryCode]?.description ?: "Unknown",
-    summary = it.appointmentDescription ?: referenceCodesForAppointmentsMap[it.categoryCode]?.description,
+    summary = when (it.appointmentDescription.isNullOrEmpty()) {
+      true -> referenceCodesForAppointmentsMap[it.categoryCode]?.description
+      false -> "${it.appointmentDescription} (${referenceCodesForAppointmentsMap[it.categoryCode]?.description})"
+    },
```
leejacobson-moj, line 155: Is this the best way to do this conditional logic?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/util/TransformFunctions.kt
```diff
@@ -149,7 +149,10 @@ fun transformAppointmentInstanceToScheduledEvents(
     suspended = false,
     categoryCode = it.categoryCode,
     categoryDescription = referenceCodesForAppointmentsMap[it.categoryCode]?.description ?: "Unknown",
-    summary = it.appointmentDescription ?: referenceCodesForAppointmentsMap[it.categoryCode]?.description,
+    summary = when (it.appointmentDescription.isNullOrEmpty()) {
+      true -> referenceCodesForAppointmentsMap[it.categoryCode]?.description
+      false -> "${it.appointmentDescription} (${referenceCodesForAppointmentsMap[it.categoryCode]?.description})"
```
iterative-it, line 154: In the event that the reference code isn't mapped, it should ideally fall back to the code itself so add ?: it.categoryCode after .description (or do that at the start of the map to reduce code duplication)

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/util/TransformFunctions.kt
iterative-it: It's personal preference. You can also do
```
if (it.appointmentDescription.isNullOrEmpty()) referenceCodesForAppointmentsMap[it.categoryCode]?.description else "${it.appointmentDescription} (${referenceCodesForAppointmentsMap[it.categoryCode]?.description})"
```

Which I personally prefer but that's cause my background isn't functional programming. @matt-crowson-moj will have the best approach

leejacobson-moj: I considered that, but given it's length it doesn't read well as a single line and the syntax looked a bit ugly to me when splitting the if else across two lines.

The `when` looks neater to me, but I'm not convinced it's the best way to do a simple inline if else conditional.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/util/TransformFunctions.kt
leejacobson-moj: Updated

PR #424, opened 2023-06-27
SAA-918: Use default location prefix format where no explicit pattern exists

iterative-it:
Matches the existing Whereabouts approach used in https://github.com/ministryofjustice/whereabouts-api/blob/main/src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/LocationGroupServiceSelector.kt and https://github.com/ministryofjustice/whereabouts-api/blob/main/src/main/java/uk/gov/justice/digital/hmpps/whereabouts/services/LocationGroupFromPrisonApiService.kt

Tested with Leeds and Gartree


timharrison-moj (dismissed):
LGTM.

timharrison-moj (commented):
Sorry, tests failed.

timharrison-moj (approved):
LGTM

PR #427, opened 2023-06-28
SAA-907: Return empty list of scheduled events if prisoner number list is empty

smcveigh941:
* Prevents prisonAPI throwing a `400 Bad Request` when an empty list is passed to it
* Should not be stripping the regex away from the config


timharrison-moj (approved):
LGTM

PR #431, opened 2023-06-29
SAA-877 archived activities should not be updateable this change prevents that.

matt-crowson-moj:
The API was not guarding the updating of archived activities.

At time of writing the frontend still doesn't.  You can go directly to edit an (archived) activity bypassing the journey to get there.


timharrison-moj (approved):
LGTM

PR #434, opened 2023-06-29
SAA-877 checking the supplied prison code matches that of the allocation prior to updating.

matt-crowson-moj:
This change fixes an issue whereby even though we are supplying the prison code along with the allocation prior to update we were not checking the allocations activity is at the given prison code.


yasinmustafa (approved):
lgtm

PR #436, opened 2023-06-30
SAA-933 this change corrects the error message when creating an activity if the education study area description does not match NOMIS.

matt-crowson-moj:


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityServiceTest.kt
```diff
@@ -383,6 +383,38 @@ class ActivityServiceTest {
       .hasMessage("The education level code '1' is not active in NOMIS")
   }
 
+  @Test
+  fun `createActivity - study area description does not match NOMIS`() {
```
matt-crowson-moj, line 387: Add (missing) test coverage.

timharrison-moj (approved):
LGTM

PR #441, opened 2023-07-04
SAA-935: Adds on-wing activity location

leejacobson-moj:
## Overview

Adds support for on-wing activity flag

PR #442, opened 2023-07-04
SDIT-897 Attendance sync multiple ended allocations

mikehalmamoj:
We made an incorrect assumption that when we are looking for an attendance record to sync. there will be exactly one active allocation. This is not always true - we could receive a request to expire an attendance *after* the prisoner has been deallocated from the course. Therefore we just get the latest allocation rather than the active allocation.


src/main/resources/migrations/common/V2023.07.04.2__fix_attendance_sync_view.sql
```diff
@@ -0,0 +1,25 @@
...
+       a.comment,
+       a.status,
+       a.pay_amount,
+       a.bonus_amount,
+       a.issue_payment,
+       a2.start_date as allocation_start_date
+from attendance a
+         join scheduled_instance si on a.scheduled_instance_id = si.scheduled_instance_id
+         join allocation a2 on si.activity_schedule_id = a2.activity_schedule_id and
+                               a.prisoner_number = a2.prisoner_number
```
mikehalmamoj, line 24: The change to the SQL is that we used to also look for `a2.prisoner_status != 'ENDED'` when joining on allocations - now we just get them all.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/SynchronisationService.kt
```diff
@@ -1,14 +1,13 @@
...
 @Service
 class SynchronisationService(private val repository: AttendanceSyncRepository) {
-  fun getAttendanceSync(attendanceId: Long): AttendanceSync =
-    repository.findById(attendanceId)
-      .orElseThrow { EntityNotFoundException("Attendance sync not found: $attendanceId") }
-      .toModel()
+  fun findAttendanceSync(attendanceId: Long): AttendanceSync? =
+    repository.findAllByAttendanceId(attendanceId)
+      .maxByOrNull { it.allocationStartDate }
+      ?.toModel()
```
mikehalmamoj, line 12: As we now get all allocations we are only interested in the latest one.

src/test/resources/test_data/seed-activity-id-17.sql
```diff
@@ -38,7 +38,7 @@ insert into allocation(allocation_id, activity_schedule_id, prisoner_number, boo
 values (5, 2, 'A22222A', 10002, 3, '2022-10-10', null, '2022-10-10 10:00:00', 'MRS BLOGS', null, null, null, null, null, null, 'ACTIVE');
 
 insert into allocation(allocation_id, activity_schedule_id, prisoner_number, booking_id, prison_pay_band_id, start_date, end_date, allocated_time, allocated_by, deallocated_time, deallocated_by, deallocated_reason, suspended_time, suspended_by, suspended_reason, prisoner_status)
-values (6, 1, 'A33333A', 10003, 2, '2022-10-20', null, '2022-10-20 09:00:00', 'MRS BLOGS', null, null, null, null, null, null, 'SUSPENDED');
+values (6, 1, 'A33333A', 10003, 2, '2022-10-20', null, '2022-10-20 09:00:00', 'MRS BLOGS', null, null, null, null, null, null, 'ENDED');
```
mikehalmamoj, line 41: Test data changed to have multiple allocations where none is currently active - but we're only interested in the latest one.

src/main/resources/migrations/common/V2023.07.04.2__fix_attendance_sync_view.sql
```diff
@@ -0,0 +1,25 @@
+-- =============================================
+-- ATTENDANCE SYNC VIEW
+-- =============================================
+
+CREATE OR REPLACE VIEW v_attendance_sync AS
```
matt-crowson-moj, line 5: So @leejacobson-moj also has some migrations and I think this clashes.

So perhaps check with him first - this is the [PR](https://github.com/ministryofjustice/hmpps-activities-management-api/pull/441)

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/SynchronisationService.kt
timharrison-moj: So one attendance row has a specific scheduled_instance as its parent, and that has a specific activity_schedule as its parent, but there could potentially be a few allocations of the same prisoner to that activity_schedule (if they've been allocated, deallocated and and allocated again)?  So this change will always take the latest allocation. Sounds reasonable.

timharrison-moj (dismissed):
LGTM

matt-crowson-moj (approved):
lgtm

PR #446, opened 2023-07-06
SAA-803: Added IRSA service account and removing access keys for queues and topics

timharrison-moj:
Part of the Cloud Platform push to remove long-lived credentials.


matt-crowson-moj (approved):
LGTM

PR #449, opened 2023-07-10
SAA-979 changes to add/remove instances when the run on bank holiday flag is updated to true/false.

matt-crowson-moj:
This change removes or adds future scheduled instances for an activity when the runs on bank holiday flag is updated to true or false.

If an activity was initially set up to run on bank holidays but was subsequently updated to no longer run on bank holidays then any future instances that exist which run on bank holidays will be removed.

The opposite is true if the activity was initially set up to not run on bank holidays, future instances for those bank holidays will be added.


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/helpers/Assertions.kt
```diff
@@ -0,0 +1,11 @@
+package uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.helpers
+
+import org.assertj.core.api.Assertions.assertThat
+
+internal infix fun <T> T.isEqualTo(value: T) {
```
matt-crowson-moj, line 5: Happy to remove these if peeps don't like ...

PR #450, opened 2023-07-11
SAA-979 changing forward creation of instances to 14 days in dev/preprod and 60 days in prod when create/edit activities.

matt-crowson-moj:
Creation of schedule instances on activity creation and  update needs to be the same as the job that creates scheduled instances in the morning.

This will help provide a consistent view when looking at activities in the service that did not exist when the job ran in the early hours of the morning.


timharrison-moj (approved):
LGTM

PR #451, opened 2023-07-11
SAA-964: Disallow duplicate pay band and iep level combinations for pay

smcveigh941:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Activity.kt
```diff
@@ -160,6 +160,14 @@ data class Activity(
     pieceRate: Int?,
     pieceRateItems: Int?,
   ) {
+    require(
```
matt-crowson-moj, line 163: Perhaps a little cleaner to do this?

```
    require(activityPay().none { it.payBand == payBand && it.incentiveNomisCode == incentiveNomisCode }) {
      "The pay band and incentive level combination must be unique for each pay rate"
    }

```

Either way double '==' instead of triple '==='

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Activity.kt
```diff
@@ -160,6 +160,14 @@ data class Activity(
...
+      activityPay().find {
```
matt-crowson-moj, line 164: Enforce this at the DB level or do we do this already?

matt-crowson-moj (approved):
LGTM - shame we can't do this in the DB as well.

PR #452, opened 2023-07-12
SAA-911-3 - reset time should be the new time

gerryallan-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Attendance.kt
```diff
@@ -155,6 +155,7 @@ data class Attendance(
     pieces = null
     caseNoteId = null
     otherAbsenceReason = null
+    recordedTime = LocalDateTime.now()
```
matt-crowson-moj, line 158: Does this have test coverage?  I know it is a one line change but it should be covered by a test really.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Attendance.kt
```diff
@@ -147,7 +147,7 @@ data class Attendance(
     )
   }
 
-  private fun resetAttendance() {
+  fun resetAttendance() {
```
matt-crowson-moj, line 150: By making this pubic you can potentially reset an attendance which is no longer editable.

I'll pick this up.

matt-crowson-moj (approved):
Approved but does need another change which I will pick up.

PR #455, opened 2023-07-13
SAA-998: DB and API updates for multi-week schedules

leejacobson-moj:
## Overview

This PR adds necessary DB columns and API response model updates in preparation for supporting multi-week schedules.

Changes include:
- New `schedule_weeks` column on the `ACTIVITY_SCHEDULE` table for recording the number of weeks in an activity schedule
- New `week_number` column on the `ACTIVITY_SCHEDULE_SLOT` table for recording the week of the activity schedule which the slot details relate to
- Updates `ActivitySchedule` and `ActivityScheduleLite` models to include `scheduleWeeks` field
- Updates `ActivityScheduleSlot` model to include `weekNumber`

Note: The ability to create / edit multi-week schedules will be added in a later PR

### Screenshots

Screenshot at Jul 13 14-29-11


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/ActivitySchedule.kt
```diff
@@ -70,6 +70,8 @@ data class ActivitySchedule(
   var updatedBy: String? = null,
 
   var instancesLastUpdatedTime: LocalDateTime? = null,
+
+  val scheduleWeeks: Int,
```
matt-crowson-moj, line 74: Does this need to be a `var` to support updates?

matt-crowson-moj, line 74: Could add some validation at the domain level to make sure it always a positive value.

PR #457, opened 2023-07-13
SAA-929 - Spiking use of ResponseBodyAdvice

sioldham:
I have added the `ActivityControllerResponseFilterAdvice` class which will apply to any Controller method returning an activity. It checks the `Caseload-Id` header and, if it doesn't match the prison, returns a 404 (via the usual `EntityNotFoundException` route).

If you run the `ActivityIntegrationTest`, specifically `attempting to a get an activity from a different prison returns a 404`  you can observe the behaviour (I've updated the `WebTestClient.getActivityById` function to send the correct header by default.

Example log line:

`ERROR 55835 --- [o-auto-1-exec-1] ControllerResponseFilterAdvice$Companion : Cannot return Activity [2] from method [getActivityById] because the Activity is for prison [PVI] and the user's case load is for prison [MDI]. |`


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/resource/filter/ActivityControllerResponseFilterAdvice.kt
```diff
@@ -0,0 +1,39 @@
...
+import org.springframework.web.servlet.mvc.method.annotation.ResponseBodyAdvice
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.model.Activity
+
+@ControllerAdvice
+class ActivityControllerResponseFilterAdvice : ResponseBodyAdvice<Activity> {
+
+  companion object {
+    val log = LoggerFactory.getLogger(this::class.java)
+  }
+
```
timharrison-moj, line 20: This is quite nice as it's "outside" the real code and adding an extra bit of functionality. The challenge is how to do this for different return types? Might get very repetitive, as we have Activity, ActivityLite, ActivityBasic, ActivitySchedule, ActivityScheduleLite, ScheduledInstances ... etc... the list goes on.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/resource/filter/ActivityControllerResponseFilterAdvice.kt
```diff
@@ -0,0 +1,39 @@
...
+    if (activity?.prisonCode != request.headers.getFirst("Caseload-Id")) {
+      throw EntityNotFoundException()
+    } else {
+      return activity
+    }
+  }
+
+  override fun supports(returnType: MethodParameter, converterType: Class<out HttpMessageConverter<*>>): Boolean {
+    return returnType.parameterType.equals(Activity::class.java) && returnType.method.name.startsWith("get")
+  }
```
timharrison-moj, line 38: The issue is going to be with all the different return types. Because of this, I wonder whether a util function, called directly from the service layer might be a) clearer, and b) something we can apply optionally to different calls, as some already accept the prison as a parameter.  

Something like:
```
fun serviceLayerMethod(req: Request, caseload: String?): ActivityTypeThing {
  val myThing = respository.getTheThing(req)
  util.checkCaseloadAccess(caseload, myThing.prisonCode)  // would throw if not accessible & controller advice catch
  return transform(myThing)  
}
```

timharrison-moj: Many ways to skin this, so just thinking of a clear, flexible, optional way to do it?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/resource/filter/ActivityControllerResponseFilterAdvice.kt
sioldham: Good point. I've changed it to use and interface instead. That would solve the multiple return type problem

PR #459, opened 2023-07-14
SAA-998: Validation

leejacobson-moj:
## Overview

Adds validation for new entity fields.


iterative-it (approved):
LGTM. In line with the ddd pattern used for these entities

timharrison-moj (approved):
LGTM

PR #460, opened 2023-07-17
SAA-929 - Spiking applying checks in the service layer

sioldham:
Applied CaseLoad ID check to the following endpoints:

```
GET /activities/{activityId}
GET /schedules/{scheduleId}
GET /scheduled-instances/{instanceId}
GET /schedules/{scheduleId}/allocations
GET /schedules/{scheduleId}/candidates
GET /allocations/id/{allocationId}
```


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/config/ControllerAdvice.kt
```diff
@@ -83,6 +84,20 @@ class ControllerAdvice(private val mapper: ObjectMapper) : ResponseEntityExcepti
       )
   }
 
+  @ExceptionHandler(CaseLoadAccessException::class)
+  fun handleCaseLoadAccessException(e: CaseLoadAccessException): ResponseEntity<ErrorResponse> {
+    log.info("Case load access exception: {}", e.message)
+    return ResponseEntity
```
timharrison-moj, line 90: Method is fine - I wonder if 403 is more appropriate? It is there, but YOU can see it due to caseload permissions.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/util/CheckCaseLoadAccess.kt
```diff
@@ -0,0 +1,7 @@
+package uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.util
+
+fun checkCaseLoadAccess(prisonCode: String, caseLoadId: String?) {
+  if (prisonCode != caseLoadId) throw CaseLoadAccessException()
+}
```
timharrison-moj, line 5: I think we might need to allow here if the call was made using a client token rather than user token? This was the very last part of the conversation in the thread I linked you into. If no caseload header, and a client token was used or had role ACTIVITY_ADMIN, allow it..

This would allow for dev use using personal tokens, sync service use, and other clients who might call our service.

timharrison-moj: No users will have the ROLE_ACTIVITY_ADMIN.

timharrison-moj (commented):
I like the method - its clear, simple and concentrated in one place. Couple of comments to help make it fit with both user requests and client requests.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/resource/ActivityController.kt
```diff
@@ -90,8 +91,8 @@ class ActivityController(
       ),
     ],
   )
-  fun getActivityById(@PathVariable("activityId") activityId: Long): Activity =
-    activityService.getActivityById(activityId)
+  fun getActivityById(@PathVariable("activityId") activityId: Long, @RequestHeader(CASELOAD_ID) caseLoadId: String?): Activity =
```
matt-crowson-moj, line 94: Question.

Given the header is optional how will we know when to enforce it i.e. head required/not required?

For example if the activity was being pulled back as part of data sync needs they won't have an active case load id.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/util/CheckCaseLoadAccess.kt
matt-crowson-moj: Ah I see you have already asked this.  @sioldham this is where I was going with my other comment.

sioldham: I think it's woth moving the salient parts of that thread over here. 
@matt-crowson-moj  Is the code below how we distinguish between a "client token" and a "user token"?

`private fun findPrincipal(claims: Map<String, Any?>): String {
  return if (claims.containsKey("user_name")) {
    claims["user_name"] as String
  } else {
    claims["client_id"] as String
  }
}`

sioldham: Might be worth packaging all this into some sort of `ClientDetails` class to pass to the check-and-throw function?

```
data class ClientDetails (
    val caseLoadId: String?,
    val tokenType: TokenType,
    val roles: Set<String>
)

enum class TokenType (
    USER,
    CLIENT
)
```

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/resource/ActivityController.kt
```diff
@@ -90,8 +96,14 @@ class ActivityController(
       ),
     ],
   )
-  fun getActivityById(@PathVariable("activityId") activityId: Long): Activity =
-    activityService.getActivityById(activityId)
+  fun getActivityById(
+    @PathVariable("activityId") activityId: Long,
+    @RequestHeader(CASELOAD_ID) caseLoadId: String?,
+    @RequestHeader(AUTHORIZATION) authToken: String,
```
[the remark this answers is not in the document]
sioldham: I thought that too. I hate having to strip the bearer prefix off and manually parse the token. This is what I was alluding to on the standup this morning - I need to access the JWT to examine the claims but I couldn't immediately find a nice utility.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/util/CheckCaseLoadAccess.kt
```diff
@@ -0,0 +1,26 @@
...
+import org.springframework.security.core.context.SecurityContextHolder
+import org.springframework.security.oauth2.jwt.Jwt
+import org.springframework.web.context.request.RequestContextHolder
+import org.springframework.web.context.request.ServletRequestAttributes
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.resource.CASELOAD_ID
+
+fun checkCaseLoadAccess(prisonCode: String) {
+  val httpRequest = if (RequestContextHolder.getRequestAttributes() != null) (RequestContextHolder.getRequestAttributes() as ServletRequestAttributes).request else null
+  val jwt = if (SecurityContextHolder.getContext()?.authentication?.credentials != null) SecurityContextHolder.getContext().authentication.credentials as Jwt else null
+  val caseLoadId = httpRequest?.getHeader(CASELOAD_ID)
```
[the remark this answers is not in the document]
sioldham: I didn't intend to re-add it because then it has to be passed through to the service layer and it just feels a bit less tidy. I guess we can get around this by adding a Swagger/OpenAPI annotation. I'll check.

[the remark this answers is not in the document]
timharrison-moj: I think there are already built-in spring security methods which expose the principal and their claims, so probably don't need to manually parse the token? I'll look in other services for examples, but I'm sure calculate-release-dates-api used something like this .. I'll send you the link if I find something similar.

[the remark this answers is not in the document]
timharrison-moj: `AuthAwareTokenConverter` deconstructs the JWT on each request to get claims and principal name, so a new method here to determine if a `user_name` or `client_id` claim is present looks possible, or something like that?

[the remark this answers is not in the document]
sioldham: I've updated the `AuthAwareAuthenticationToken`  to store the `roles` and `isClientToken`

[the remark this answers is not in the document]
sioldham: Also sorted the swagger issue using the `@Parameter` annotation @matt-crowson-moj

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/AllocationServiceTest.kt
```diff
@@ -74,6 +73,8 @@ class AllocationServiceTest {
   fun `transformed allocation returned when find by id`() {
     val expected = allocation()
 
+    addCaseLoadIdToRequestHeader("123")
+    whenever(scheduleRepository.findById(expected.activitySchedule.activityScheduleId)).thenReturn(Optional.of(expected.activitySchedule))
```
timharrison-moj, line 77: Not sure why a caseload ID of 123 is succeeding?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/AllocationServiceTest.kt
```diff
@@ -23,17 +20,19 @@ import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.helpers.lowPayB
...
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.helpers.moorlandPrisonCode
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.model.request.AllocationUpdateRequest
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.repository.ActivityScheduleRepository
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.repository.AllocationRepository
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.repository.PrisonPayBandRepository
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.util.addCaseLoadIdToRequestHeader
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.util.toModelPrisonerAllocations
 import java.time.LocalDate
-import java.util.Optional
+import java.util.*
```
timharrison-moj, line 30: All util.* packages included?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityServiceTest.kt
```diff
@@ -252,11 +261,19 @@ class ActivityServiceTest {
 
   @Test
   fun `getActivityById returns an activity for known activity ID`() {
+    addCaseLoadIdToRequestHeader("123")
     whenever(activityRepository.findById(1)).thenReturn(Optional.of(activityEntity()))
```
timharrison-moj, line 265: Should fail with a caseload id of "123"?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ScheduledInstanceServiceTest.kt
```diff
@@ -45,6 +46,7 @@ class ScheduledInstanceServiceTest {
 
     @Test
     fun `scheduled instance found - success`() {
+      addCaseLoadIdToRequestHeader("MDI")
       whenever(repository.findById(1))
```
timharrison-moj, line 50: Small one ... Caseload is generally one word, so the camel case should be `addCaseloadIdToRequestHeader`?

timharrison-moj (commented):
LGTM.
Only a few small questions and I think its there!

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityServiceTest.kt
matt-crowson-moj: This is the default prison code in the factory method @timharrison-moj on `activityEntiity()`.

The factory `fun` should probably be changed to use a real default prison code.

timharrison-moj (approved):
LGTM
Approved - just needs branch updating and utils.* being a bit more specific I think.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/AllocationServiceTest.kt
sioldham: Done

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ScheduledInstanceServiceTest.kt
sioldham: Done

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityServiceTest.kt
sioldham: It's the default test one. I've made it clearer

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/util/CheckCaseloadAccess.kt
```diff
@@ -0,0 +1,29 @@
+package uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.util
+
+import org.springframework.security.core.context.SecurityContextHolder
+import org.springframework.web.context.request.RequestContextHolder
+import org.springframework.web.context.request.ServletRequestAttributes
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.config.AuthAwareAuthenticationToken
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.resource.ACTIVITY_ADMIN
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.resource.CASELOAD_ID
+
+fun checkCaseloadAccess(prisonCode: String?) {
```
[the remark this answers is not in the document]
sioldham: I think on at least one of the entities where you need to make a few hops `entity->parent->something->prisonId` there is a nullable in there

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ScheduledInstanceService.kt
```diff
@@ -30,6 +31,7 @@ class ScheduledInstanceService(
   fun getActivityScheduleInstanceById(id: Long): ActivityScheduleInstance {
     val activityScheduleInstance = repository.findById(id)
       .orElseThrow { EntityNotFoundException("Scheduled Instance $id not found") }
+    checkCaseloadAccess(activityScheduleInstance?.activitySchedule?.activity?.prisonCode)
```
matt-crowson-moj, line 34: Not sure this is right Si.  Goes back to my question on allowing a nullable prison code in the case load util check.

The instance cannot be null here as you have just thrown an exception if nothing found.  As a result prison code can't be null here.

I think this should be `checkCaseloadAccess(activityScheduleInstance.activitySchedule.activity.prisonCode)`

PR #462, opened 2023-07-17
SAA-996: Multi-week schedule model updates

leejacobson-moj:
## Overview

This is a small change to add the helper function needed to retrieve the week number for a given date of a multi-week schedule.

It implements this in the `ActivitySchedule.addInstances()` method in the `ActivityService` class.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/ActivitySchedule.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/ActivitySchedule.kt
@@ -152,6 +152,22 @@ data class ActivitySchedule(
 
   fun isSuspendedOn(date: LocalDate) = suspensions.any { it.isSuspendedOn(date) }
 
+  fun getWeekNumber(date: LocalDate): Int {
+    require(!date.isBefore(startDate) && (endDate == null || !date.isAfter(endDate))) {
+      "Date must be within activity schedule range."
+    }
+    // To calculate the week number for a given date:
+    // - Calculate first Monday of schedule
+    // - Calculate the total number of days into the schedule from the first Monday
+    // - Calculate the number of days into the schedule period for the given date
+    // - Finally, calculate the week number from the number of days into this schedule period
+    val daysInWeek = 7
+    val scheduleFirstMonday = startDate.minusDays(startDate.dayOfWeek.value - 1L)
+    val daysIntoSchedule = ChronoUnit.DAYS.between(scheduleFirstMonday, date)
+    val daysIntoThisSchedulePeriod = daysIntoSchedule % (daysInWeek * scheduleWeeks)
+    return (daysIntoThisSchedulePeriod / daysInWeek).toInt() + 1
+  }
+
   fun hasNoInstancesOnDate(day: LocalDate) =
     instances.none { instance -> instance.sessionDate == day }
 
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityService.kt
@@ -181,7 +181,7 @@ class ActivityService(
         scheduleWeeks = 1,
       ).let { schedule ->
         schedule.addSlots(request.slots!!, timeSlots)
-        schedule.addInstances(activity, schedule.slots(), null, null)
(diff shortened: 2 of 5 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityService.kt
```diff
@@ -229,8 +229,10 @@ class ActivityService(
     val endDay = today.plusDays(daysInAdvance)
     val listOfDatesToSchedule = today.datesUntil(endDay).toList()
 
-    listOfDatesToSchedule.forEach { day ->
-      slots.forEach { slot ->
+    listOfDatesToSchedule.filter { activity.isActive(it) }.forEach { day ->
+      val weekNumber = this.getWeekNumber(day)
+
+      slots.filter { it.weekNumber == weekNumber }.forEach { slot ->
```
leejacobson-moj, line 235: We only want to add instances for slots with a matching week number

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityService.kt
```diff
@@ -229,34 +228,34 @@ class ActivityService(
...
+              thursday = slot.thursdayFlag,
+              friday = slot.fridayFlag,
+              saturday = slot.saturdayFlag,
+              sunday = slot.sundayFlag,
+            ),
+          )
+
+          if (day.dayOfWeek in daysOfWeek &&
+            this.hasNoInstancesOnDate(day, slot.startTime to slot.endTime) &&
+            (runsOnBankHoliday || !bankHolidayService.isEnglishBankHoliday(day))
```
leejacobson-moj, line 253: I've refactored the logic a little here to try to improve readability.

I've replace the `Activity.isActive()` check with the `ActivitySchedule.isActiveOn()` method. I believe this is more correct since in theory an activity could have multiple schedules with differing active states.

I've also moved the `isEnglishBankHoliday()` call be the last check before adding the new instance, because presumably this is the most expensive check to make.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityService.kt
```diff
@@ -392,14 +391,16 @@ class ActivityService(
 
       activity.startDate = newStartDate
       activity.schedules().forEach {
-        if (it.startDate < newStartDate) {
+        val currentScheduleStartDate = it.startDate
+        it.startDate = newStartDate
```
leejacobson-moj, line 395: I've moved setting the schedule `startDate` here because it's now needed by `.addInstances()`.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityService.kt
timharrison-moj: Agreed, looks good.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/ActivitySchedule.kt
```diff
@@ -152,6 +152,25 @@ data class ActivitySchedule(
...
+    // start date (schedule "start week").
+    // To calculate the week number for a given date:
+    // - Calculate schedule start week.
+    // - Calculate the number of times the schedule has run for the given date.
+    // - Calculate the start week for schedule running on the given date.
+    // - Finally, calculate the week number for the schedule running on the given date
+    val daysInWeek = 7
+    val daysInSchedule = (daysInWeek * scheduleWeeks)
+    val scheduleStartWeek = startDate.minusDays(startDate.dayOfWeek.value - 1L)
+    val scheduleOccurrenceNo = ChronoUnit.DAYS.between(scheduleStartWeek, date) / daysInSchedule
```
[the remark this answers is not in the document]
leejacobson-moj: Thanks Tim, basically I was calculating the remainder slightly differently. I was finding the first Monday of the schedule running at the given date, then the difference between this Monday and given date would be your remainder, but your suggestion is a more elegant way to do that.

Although, instead of the doing `remainder` < `daysInWeek` a division there would return the week number for an arbitrary number of weeks in a schedule (eg, 4 week schedules).

I'll update.

timharrison-moj (approved):
LGTM

PR #467, opened 2023-07-18
SAA-1012: Only remove redundant future instances when updating slots

leejacobson-moj:
## Overview

Small change to the create / remove instance logic. This fixes the bug where all past instances are removed during a slot update and implements the following rules for activity create / edit requests:

- All past instances should be kept (regardless of attendance or previous slot settings)
- Any instances already scheduled for today should remain active
- Any future instances (tomorrow or after) which are no longer valid should be removed (regardless of existing attendance records)
- Any future instances (tomorrow or after) which are still valid should be kept
- Where new instances need to be created these should be created for tomorrow onwards

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityService.kt
@@ -224,9 +224,9 @@ class ActivityService(
     fromDate: LocalDate? = null,
     toDate: LocalDate? = null,
   ) {
-    val today = LocalDate.now()
-    val endDay = today.plusDays(daysInAdvance)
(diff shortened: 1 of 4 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityService.kt
```diff
@@ -224,9 +224,9 @@ class ActivityService(
     fromDate: LocalDate? = null,
     toDate: LocalDate? = null,
   ) {
-    val today = LocalDate.now()
-    val endDay = today.plusDays(daysInAdvance)
-    val listOfDatesToSchedule = today.datesUntil(endDay).toList()
+    val tomorrow = LocalDate.now().plusDays(1)
```
leejacobson-moj, line 227: We should only add new instances from tomorrow onward.

PR #472, opened 2023-07-20
SAA-996: Multi-week schedule updates

leejacobson-moj:
## Overview

This change primarily updates the `addSlot()` method to support adding slots for multi-week schedules, but also includes some improves around creating and removing instances.

Note: Because `createActivity()` and `updateActivity()` use the request models I will add tests for creating and updating multi-week schedules once those models have been updated.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityService.kt
@@ -12,7 +12,6 @@ import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.client.prisonap
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.common.TimeSlot
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.entity.Activity
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.entity.ActivitySchedule
-import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.entity.ActivityScheduleSlot
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.entity.ActivityState
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.entity.toModelLite
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.model.ActivityScheduleLite
@@ -162,13 +161,6 @@ class ActivityService(
 
     activity.let {
       val scheduleLocation = if (request.inCell || request.onWing) null else getLocationForSchedule(it, request.locationId!!)
-      val prisonRegime = prisonRegimeService.getPrisonRegimeByPrisonCode(activity.prisonCode)
-      val timeSlots =
-        mapOf(
-          TimeSlot.AM to Pair(prisonRegime.amStart, prisonRegime.amFinish),
-          TimeSlot.PM to Pair(prisonRegime.pmStart, prisonRegime.pmFinish),
-          TimeSlot.ED to Pair(prisonRegime.edStart, prisonRegime.edFinish),
-        )
 
       activity.addSchedule(
         description = request.description!!,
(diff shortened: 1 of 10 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityService.kt
```diff
@@ -206,47 +200,32 @@ class ActivityService(
...
 
   /*
    * Note: we add instances even if the activity hasn't started for unlock list purposes.
    */
-  private fun ActivitySchedule.addFutureInstances(
-    slots: List<ActivityScheduleSlot>,
-    fromDate: LocalDate? = null,
-    toDate: LocalDate? = null,
-  ) {
+  private fun ActivitySchedule.addInstances() {
```
leejacobson-moj, line 216: This method can be called to create new activity schedule instances whenever this may be required (eg, update requests).

It no longer requires params and will instead use the activity schedule entity to check if whether new instances need to be created, and if so create them.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityService.kt
```diff
@@ -466,20 +419,32 @@ class ActivityService(
...
 
-          futureSessionDatesToRemove.forEach { schedule.removeInstances(it, it) }
-        }
-      }
+  private fun ActivitySchedule.updateInstances() {
+    this.removeRedundantInstances()
+    this.addInstances()
+  }
+
+  private fun ActivitySchedule.removeRedundantInstances() {
```
leejacobson-moj, line 431: Similar to `addInstances()` this method can be called whenever instances might need to be removed.

It will only remove future instances and will consider: start date, end date, slots (supports multi-week) and bank holidays.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityService.kt
```diff
@@ -323,7 +302,10 @@ class ActivityService(
     activity.updatedTime = now
     activity.updatedBy = updatedBy
 
-    activity.schedules().forEach { it.markAsUpdated(now, updatedBy) }
+    activity.schedules().forEach {
+      it.updateInstances()
```
leejacobson-moj, line 306: Previously the ActivitySchedule entity was responsible for updating it's instances with the exception of bank holiday flag updates because this required a call to the bank holiday service.

For consistency I've removed the logic for creating new instances from `ActivitySchedule` and replaced with this single `updateInstances()` service call which can be safely called at any time and for any update to the `ActivitySchedule`.

I think this is a better approach since a call to update a schedule's start time or slots will no longer come with the potential unintended side effect of creating / removing instances. Instead, if instances need to be updated `updateInstances()` can be used.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityService.kt
```diff
@@ -593,20 +558,14 @@ class ActivityService(
...
   }
 
-  private fun List<Slot>.toMap(regimeTimeSlots: Map<TimeSlot, Pair<LocalTime, LocalTime>>) =
-    associate { regimeTimeSlots[it.timeSlot()]!! to it.getDaysOfWeek() }
+  private fun List<Slot>.toMap(regimeTimeSlots: Map<TimeSlot, Pair<LocalTime, LocalTime>>):
+    Map<Pair<Int, Pair<LocalTime, LocalTime>>, Set<DayOfWeek>> {
+    // TODO: Add support for multi-week schedules
+    val weekNumber = 1
+    return this.associate { Pair(weekNumber, regimeTimeSlots[it.timeSlot()]!!) to it.getDaysOfWeek() }
+  }
```
timharrison-moj, line 570: It's look good, and clearer. However, there are a lot of changes to the fundamental scheduling part of the service here so can I ask the obvious question please? How much local testing have you put this through? And should somebody else help you by trying this with a range of scenarios before we approve/merge and give it a go on DEV?

leejacobson-moj: I did some local testing this morning. Everything seems to be working so far as I can tell... Creating activities, adding slots, updating start times, allocation, etc.

I'd argue the changes are less scary then they look because it's mostly replacing the duplicated logic around create / removing instances with two reusable methods. I spent quite a bit of time looking at the logic there and I'm quite sure both are doing what they should be doing.

It's possible there are things I've missed though. If someone has time to run through a few scenario locally that would be appreciated. Hopefully there shouldn't be any major issues though.

timharrison-moj (commented):
LGTM. Would like to try this locally, and perhaps get Matt's eyes on it too?

timharrison-moj (approved):
LGTM. I checked this out and ran locally to confirm that it still operates in the same way for a 1-week schedule. And I cheekily added a 2-week activity schedule with different slots for week 1 and week 2, and it updated the instances to correctly schedule the week 1 slots this week and the week 2 slots next week.

PR #476, opened 2023-07-21
SAA-1016 changes to audit deallocation of a prisoner upon being permanently released.

matt-crowson-moj:
Deallocations are now audited upon a prisoner being permanently released from prison.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/AuditableEntityListener.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/AuditableEntityListener.kt
@@ -10,11 +10,13 @@ import org.springframework.stereotype.Component
 import toActivityCreatedEvent
 import toActivityUpdatedEvent
 import toPrisonerAllocatedEvent
+import toPrisonerDeallocatedEvent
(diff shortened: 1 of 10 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/AuditableEntityListener.kt
```diff
@@ -35,17 +37,29 @@ class AuditableListener {
   fun onUpdate(entity: Any) {
     when (entity) {
       is Activity -> audit(entity.toActivityUpdatedEvent(), "Failed to audit activity update event for activity id ${entity.activityId}")
+      is Allocation -> audit(entity)
     }
   }
 
+  private fun audit(allocation: Allocation) {
```
timharrison-moj, line 44: General question - do you think this will get cumbersome as we decide to add more things to be audited? Wondering whether we try this for now and see how it goes, but have a suspicion that we might end up with specific audit function calls in the services. The problem is in identifying the `reason` for the update.

matt-crowson-moj: Yes agree. It does feel like perhaps nearer to the point at which it happens should possibly be responsible.  Not 100% sure though.

PR #481, opened 2023-07-25
SAA-1026 3rd party dependency updates to resolve security warnings and also use latest versions where available.

matt-crowson-moj:


timharrison-moj (approved):
LGTM

PR #485, opened 2023-07-27
Record interesting event if allocation is pending

smcveigh941:


matt-crowson-moj (approved):
LGTM

timharrison-moj (approved):
LGTM. Yes, agree that active or pending, we should record the interesting events, as it might alter their decisions.

PR #486, opened 2023-07-27
SAA-849: Updated create daily instances job and request models

leejacobson-moj:
## Overview

This completes the changes required to support multi-week schedules. Specifically this change includes:
- Updated create & update request models with multi-week schedule fields
- Updates to activity to support new request fields
- Updates to `CreateScheduledInstancesJob` to support multi-week schedules


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/model/request/ActivityCreateRequest.kt
```diff
@@ -122,6 +124,11 @@ data class ActivityCreateRequest(
   )
   val capacity: Int?,
 
+  @field:Min(value = 1, message = "Schedule weeks must be either 1 or 2")
+  @field:Max(value = 2, message = "Schedule weeks must be either 1 or 2")
```
leejacobson-moj, line 128: This is an arbitrary restriction given our requirement for just 1 and 2 week schedules. The API should support any number of schedule weeks.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/repository/ActivityRepositoryCustom.kt
```diff
@@ -38,4 +44,27 @@ class ActivityRepositoryImpl : ActivityRepositoryCustom {
     }.onFailure { log.error("Activity by ID with filters ${it.message}") }
       .getOrNull()
   }
+
+  @Override
+  override fun findByActivityIdAndPrisonCodeWithFilters(
```
leejacobson-moj, line 49: Keen to understand if this is a good approach (or even required).

My concern here was that previously when updating an activity all of its instances were being fetched from the DB then filtered to find the ones which needed to be removed.

Since we never want to update or remove past instances and a long running activity could have many hundreds of past instances I thought adding this filter to fetch only the future instances might be worthwhile.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ManageScheduledInstancesService.kt
```diff
@@ -74,38 +72,30 @@ class ManageScheduledInstancesService(
...
-            log.info("This instance was suspended ${schedule.description} on $day at ${slot.startTime}")
+            log.info("This instance was suspended ${schedule.description} on $day")
           }
 
-          instancesCreated = true
-        }
-      }
-    }
+          val lastUpdate = schedule.instancesLastUpdatedTime
+          activityService.addScheduleInstances(schedule, days)
```
leejacobson-moj, line 89: To remove duplicate logic I've updated this job to use the existing method in `activityService` to add new instances.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/repository/ActivityRepositoryCustom.kt
matt-crowson-moj: Agree. We could use the existing custom method and check the prison code against that but meh.

I think perhaps also we need to start using the caseload checking in the service here?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/model/request/ActivityCreateRequest.kt
```diff
@@ -122,6 +124,11 @@ data class ActivityCreateRequest(
   )
   val capacity: Int?,
 
+  @field:Min(value = 1, message = "Schedule weeks must be either 1 or 2")
```
matt-crowson-moj, line 127: I'd probably change the messages. Min 1 and Max 2 and not have them the same, meh

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/model/request/ActivityCreateRequest.kt
```diff
@@ -139,6 +146,9 @@ data class ActivityCreateRequest(
 )
 
 data class Slot(
+  @field:Positive(message = "The week number must be a positive integer")
+  @Schema(description = "The week of the schedule this slot relates to", example = "1")
+  val weekNumber: Int,
```
matt-crowson-moj, line 151: Given the arbitrary restriction above should this be restricted as well?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/model/request/ActivityUpdateRequest.kt
```diff
@@ -87,6 +89,11 @@ data class ActivityUpdateRequest(
   @Schema(description = "The list of pay rates that can apply to this activity")
   val pay: List<ActivityPayCreateRequest>? = null,
 
+  @field:Min(value = 1, message = "Schedule weeks must be either 1 or 2")
```
matt-crowson-moj, line 92: Same as my comments on the create request.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/model/request/ActivityCreateRequest.kt
leejacobson-moj: I could add extra validation to these request models, but this is covered by the entity validation in `ActivityScheduleSlot.kt`

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityService.kt
```diff
@@ -202,41 +198,47 @@ class ActivityService(
...
-        this.slots().filter { it.weekNumber == weekNumber }.forEach { slot ->
-          val daysOfWeek = slot.getDaysOfWeek()
-
-          if (day.dayOfWeek in daysOfWeek &&
-            this.hasNoInstancesOnDate(day, slot.startTime to slot.endTime) &&
-            (runsOnBankHoliday || !bankHolidayService.isEnglishBankHoliday(day))
-          ) {
-            this.addInstance(sessionDate = day, slot = slot)
-          }
+  private fun ActivitySchedule.addInstances(daysToSchedule: List<LocalDate>? = null) {
```
matt-crowson-moj, line 219: I have tried to make this a bit easier to follow though still think it is not ideal

```
  private fun ActivitySchedule.addInstances(daysToSchedule: List<LocalDate>? = null) {
    val possibleDaysToSchedule = daysToSchedule ?: LocalDate.now().let { today -> today.datesUntil(today.plusDays(daysInAdvance)).toList() }

    possibleDaysToSchedule.filter(::isActiveOn).forEach { activeDay ->
        val scheduleWeekNumber = this.getWeekNumber(activeDay)

        this.slots().filter { slot -> slot.weekNumber == scheduleWeekNumber }.forEach { slot ->
          if (activeDay.dayOfWeek in slot.getDaysOfWeek() &&
            this.hasNoInstancesOnDate(activeDay, slot.startTime to slot.endTime) &&
            (runsOnBankHoliday || !bankHolidayService.isEnglishBankHoliday(activeDay))
          ) {
            this.addInstance(sessionDate = activeDay, slot = slot)
            this.instancesLastUpdatedTime = LocalDateTime.now()
          }
        }
      }
  }

```

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityService.kt
```diff
@@ -202,41 +198,47 @@ class ActivityService(
...
+
+      this.slots().filter { it.weekNumber == weekNumber }.forEach { slot ->
+        val daysOfWeek = slot.getDaysOfWeek()
+
+        if (day.dayOfWeek in daysOfWeek &&
+          this.hasNoInstancesOnDate(day, slot.startTime to slot.endTime) &&
+          (runsOnBankHoliday || !bankHolidayService.isEnglishBankHoliday(day))
+        ) {
+          this.addInstance(sessionDate = day, slot = slot)
+          this.instancesLastUpdatedTime = LocalDateTime.now()
```
matt-crowson-moj, line 238: I wonder if the timestamp change can go into the addInstance method?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityService.kt
```diff
@@ -446,7 +441,10 @@ class ActivityService(
         !this.runsOnBankHoliday && bankHolidayService.isEnglishBankHoliday(it.sessionDate)
     }
 
-    this.removeInstances(instancesToRemove)
+    if (instancesToRemove.isNotEmpty()) {
+      this.removeInstances(instancesToRemove)
+      this.instancesLastUpdatedTime = LocalDateTime.now()
```
matt-crowson-moj, line 446: Similar here.  Move the update the timestamp into the removeInstances function?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityService.kt
```diff
@@ -291,15 +291,12 @@ class ActivityService(
     applyInCellUpdate(request, activity)
     applyOnWingUpdate(request, activity)
     applyAttendanceRequiredUpdate(request, activity)
-    if (request.minimumEducationLevel != null) {
```
matt-crowson-moj, line 294: Good to see this moved into the respective functions.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityService.kt
matt-crowson-moj: I know/realise we are doing this in several places setting timestamps in services but it kinda feels wrong.  The entities know when they are updated.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ManageScheduledInstancesService.kt
```diff
@@ -74,38 +72,30 @@ class ManageScheduledInstancesService(
...
 
-    if (instancesCreated) {
-      // This triggers the sync event, adds new instances but does not delete the earlier ones
-      schedule.instancesLastUpdatedTime = LocalDateTime.now()
-      activityScheduleRepository.saveAndFlush(schedule)
+          // If there are been updates, save
+          if (schedule.instancesLastUpdatedTime != lastUpdate) {
+            activityScheduleRepository.saveAndFlush(schedule)
```
matt-crowson-moj, line 93: Not sure the save and flush is actually required now.  This service is transactional i.e. changes are automatically saved.  I appreciate it was here before.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ManageScheduledInstancesService.kt
matt-crowson-moj: Has this been tested locally to make sure all is working as before?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/model/request/ActivityCreateRequest.kt
leejacobson-moj: My argument against this would be that an error message saying that the schedule weeks needs to be greater or equal to 1 isn't very helpful when the only valid values are 1 or 2.

I wanted a between validator really, but that doesn't exist.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ManageScheduledInstancesService.kt
leejacobson-moj: Yeah, I've tested via the `/job/create-scheduled-instances` endpoint. As far as I can see it seems to be working correctly.

The underlying code is used during create and edit which I've tested quite a bit. I know Tim has ran some tests there too.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityService.kt
leejacobson-moj: Thanks Matt, this looks good. I'll add this now and run some tests.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ManageScheduledInstancesService.kt
leejacobson-moj: I'm going to keep this in for now since it's doing no harm and removing this will require updates to several tests which expect `.saveAndFlush()` to be called.

Agree it would be good to remove if unneeded, but probably better done in a separate ticket.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/repository/ActivityRepositoryCustom.kt
leejacobson-moj: Discussed this with Matt and have now added Simon's case load check to the create and update endpoints.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ManageScheduledInstancesService.kt
timharrison-moj: Yes, I ran this locally with a number of scenarios and this worked as expected.

timharrison-moj (commented):
LGTM. Quick update of branch, and regenerate the API types, then should be good to go.

timharrison-moj (approved):
LGTM

PR #487, opened 2023-07-27
SAA-1019 - Add caseload checks for appts

sioldham:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/AppointmentOccurrenceSearchService.kt
```diff
@@ -70,6 +71,7 @@ class AppointmentOccurrenceSearchService(
     }
 
     val results = appointmentOccurrenceSearchRepository.findAll(spec)
+    checkCaseloadAccess(results.firstOrNull()?.prisonCode)
```
matt-crowson-moj, line 74: Just wondering if there is anything else we can do before running a search for something someone might not actually have access to e.g. some sort of mini pre-query which we then check before doing the actual search?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/AppointmentOccurrenceSearchServiceTest.kt
```diff
@@ -198,4 +215,21 @@ class AppointmentOccurrenceSearchServiceTest {
...
+    val result = appointmentOccurrenceSearchEntity()
+
+    whenever(appointmentOccurrenceSearchRepository.findAll(any())).thenReturn(listOf(result))
+    whenever(appointmentOccurrenceAllocationSearchRepository.findByAppointmentOccurrenceIds(listOf(result.appointmentOccurrenceId))).thenReturn(result.allocations)
+    whenever(referenceCodeService.getReferenceCodesMap(ReferenceCodeDomain.APPOINTMENT_CATEGORY))
+      .thenReturn(mapOf(result.categoryCode to appointmentCategoryReferenceCode(result.categoryCode)))
+    whenever(locationService.getLocationsForAppointmentsMap(result.prisonCode))
+      .thenReturn(mapOf(result.internalLocationId!! to appointmentLocation(result.internalLocationId!!, "TPR")))
+
+    Assertions.assertThatThrownBy { service.searchAppointmentOccurrences("TPR", request, principal) }
```
matt-crowson-moj, line 232: Opps `Assertions.`

matt-crowson-moj (dismissed):
LGTM I do have one question though ...

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/AppointmentOccurrenceSearchService.kt
iterative-it: Actually yes. This check can be moved to the start of the function and use the `prisonCode` parameter rather than checking the results of the search. That would prevent the search being run at all if an invalid prison code header was supplied I think.

iterative-it: In fact having tested this, I've realised that you get a 403 if there are no search results so definitely move this check to the top of the method. I'd also change the signature for checkCaseloadAccess to make prisonCode non nullable as the check fails if the passed in prison code is null

iterative-it (changes requested):
See the comment around appointment occurrence search

matt-crowson-moj (approved):
LGTM

PR #492, opened 2023-08-01
SAA-690 - deallocation end date not updating

gerryallan-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/AllocationsService.kt
```diff
@@ -109,6 +109,8 @@ class AllocationsService(
       }
       if (allocation.endDate == null) {
         allocation.activitySchedule.deallocatePrisonerOn(allocation.prisonerNumber, this, request.reasonCode.toDeallocationReason(), updatedBy)
+      } else {
+        allocation.activitySchedule.deallocatePrisonerOn(allocation.prisonerNumber, this, allocation.plannedDeallocation!!.plannedReason, updatedBy)
```
matt-crowson-moj, line 113: Can do with some test coverage please if there isn't any.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/AllocationsService.kt
```diff
@@ -109,6 +109,8 @@ class AllocationsService(
...
       }
```
timharrison-moj, line 114: Not sure about this change, Gerry. Can we always assume a `plannedDeallocation` exists and assert it is not null when changing the end date?

gerryallan-moj: if endDate has already been set then there is always a plannedDeallocation

timharrison-moj (approved):
LGTM

PR #495, opened 2023-08-02
SAA-997: Limit appointment occurrence allocations

leejacobson-moj:
## Overview

Adds validation to prevent creation of extremely large repeat appointments

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/AppointmentService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/AppointmentService.kt
@@ -2,7 +2,6 @@ package uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.service
 
 import org.springframework.stereotype.Service
 import org.springframework.transaction.annotation.Transactional
-import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.client.prisonapi.api.PrisonApiUserClient
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.client.prisonersearchapi.api.PrisonerSearchApiClient
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.entity.AppointmentRepeatPeriod
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.entity.AppointmentSchedule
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/AppointmentService.kt
```diff
@@ -141,6 +139,14 @@ class AppointmentService(
     }
   }
 
+  private fun failIfMaximumOccurrencesExceeded(prisonerNumbers: List<String>, repeat: AppointmentRepeat?) {
+    val maxOccurrenceAllocations = 20000
+    val repeatCount = repeat?.count ?: 1
+    require(prisonerNumbers.size * repeatCount <= maxOccurrenceAllocations) {
+      "You cannot schedule more than ${maxOccurrenceAllocations / prisonerNumbers.size} appointments for this number of attendees."
+    }
+  }
```
timharrison-moj, line 148: Has 20,000 just been picked arbitrarily or is there some limit being reached, or performance issue?  Could maybe define as a constant at class level? e.g. `const val MAX_APPOINTMENT_OCCURRENCE_ALLOCATIONS = 20000`

timharrison-moj (approved):
LGTM. Couple of questions.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/AppointmentService.kt
leejacobson-moj: The ticket was created to mitigate performance issues related to creating large repeat appointments.

I created an appointment with just under 20,000 occurrence allocations to test this validation rule and it took a few seconds to run locally. I'm guessing 20,000 was seen as a reasonable limit since it's quite unlikely appointments of this size would need to be created. I can check if there was any reasoning for 20,000 with Emily though if you like?

I did think about making it a class const, but since it's only used in this one method and isn't likely to be needed elsewhere I thought it might be clearer if it was initialised in the method. I don't have a strong opinion though, so happy to move it if you prefer a class const?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/AppointmentService.kt
```diff
@@ -162,6 +168,8 @@ class AppointmentService(
     isCancelled: Boolean = false,
     isMigrated: Boolean = false,
   ): AppointmentEntity {
+    failIfMaximumOccurrencesExceeded(prisonerNumbers, repeat)
+
     if (!isMigrated) {
```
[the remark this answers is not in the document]
leejacobson-moj: In theory the user shouldn't ever see it because the same validation rule will be added to the create appointment journey, https://github.com/ministryofjustice/hmpps-activities-management/pull/472

If the UI validation wasn't there though it would throw an error when creating the appointment. So they'd see this on the check answers page:
Screenshot at Aug 02 13-47-30

I added the validation to the API since this could be abused if there was no limit.

PR #501, opened 2023-08-07
SAA-1065 - Duplicate activity name validation in "Edit activity" jour…

gerryallan-moj:
…ney compares and activity to itself


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityService.kt
```diff
@@ -341,7 +348,7 @@ class ActivityService(
     activity: Activity,
   ) {
     request.summary?.apply {
-      failDuplicateActivity(prisonCode, this)
+      failDuplicateActivityOnEdit(prisonCode, this, activityId = activity.activityId)
```
matt-crowson-moj, line 351: Sorry Gerry but this doesn't actually feel necessary to add another function, we have all we need.

If the summary passed in is (already) the same as the activity then this is effectively a no-op, there is no change and really nothing to save.

If they are not the same then you can use the existing method and check for a duplicate name.

gerryallan-moj: updated as suggested

matt-crowson-moj (approved):
LGTM

PR #504, opened 2023-08-08
SAA-1027: Filter out candidates on the Waiting List from the candidates list

smcveigh941:
Candidates who have a PENDING or APPROVED waiting list application should not appear in the list of suitable candidates


matt-crowson-moj (approved):
LGTM

PR #505, opened 2023-08-08
Fix bug with the workplace risk assessment filter

smcveigh941:
The previous build assumed that any alert code with type `R` must be one of `RLO, RME or RHI`. This is not the case, and there are other non workplace risk assessment alert codes with type R


timharrison-moj (approved):
LGTM

PR #506, opened 2023-08-09
SAA-727 - Handle alerts updated event

sioldham:
Tested via localstack


matt-crowson-moj:
A couple of things:

- Can you add integration test coverage in the InboundEventsIntegrationTest please.
- Is there a corresponding cloud platform PR to subscribe to the new inbound domain event?  Without this it won't work as we won't actually be listening for it.

matt-crowson-moj (approved):
LGTM

PR #511, opened 2023-08-10
SAA-1027: Update APPROVED waitlist applications when the person is allocated

smcveigh941:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityScheduleService.kt
```diff
@@ -128,7 +131,21 @@ class ActivityScheduleService(
       allocatedBy = allocatedBy,
     )
 
-    repository.saveAndFlush(schedule)
+    val allocation = repository.saveAndFlush(schedule).allocations().last()
```
matt-crowson-moj, line 134: Strictly speaking we shouldn't need to do saveAndFlush as this is now withing the context of a `@Transactional `annotation.  Same goes for the waiting lists below.

matt-crowson-moj (approved):
LGTM

PR #512, opened 2023-08-11
Bugfix/saa 958

gerryallan-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Allocation.kt
```diff
@@ -127,7 +127,6 @@ data class Allocation(
 
   private fun maybeEndDate() =
     when {
-      endDate != null -> endDate
```
matt-crowson-moj, line 130: How is this fixing things?  Sorry, I just don't understand.  This is now no longer checking against the allocation end date.

gerryallan-moj: There is no requirement to validate against the existing allocation end date - as long as we don't go past the activity end date, then it is valid for the allocation to be extended.

matt-crowson-moj (approved):
LGTM

PR #514, opened 2023-08-11
SAA-1027: Set declined waitlist applications to removed if person is allocated

smcveigh941:


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityScheduleServiceTest.kt
```diff
@@ -324,4 +324,31 @@ class ActivityScheduleServiceTest {
       assertThat(allocation?.prisonerNumber).isEqualTo("123456")
     }
   }
+
+  @Test
+  fun `allocate updates any DECLINED waitlist applications to REMOVED status`() {
+    val schedule = activitySchedule(activityEntity())
```
timharrison-moj, line 330: Is that right, that `allocate` updates any DECLINED waitlist applications to REMOVED? Not sure I follow why.

PR #516, opened 2023-08-11
SAA-864 Log key events

gerryallan-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityService.kt
```diff
@@ -123,6 +125,15 @@ class ActivityService(
...
+    val propertiesMap = mapOf(
+      "prisonName" to request.prisonCode,
+      "activityName" to request.summary,
+    )
+    val metricsMap = mapOf(
+      "numberOfResults" to 1.0,
+    )
+    telemetryClient.trackEvent("SAA-CreateActivity", propertiesMap, metricsMap)
+
     val activity = Activity(
```
sioldham, line 137: I created a set of audit event classes a while back (they're in the **model.audit** package and theres is an **ActivityCreatedEvent**). They're used by our own audit service to push events to an audit queue and/or store in the database.

I wonder if we could reuse/extend these audit classes and then extend the current audit service to turn them into telemetry events. Then, in this example, the createActivity method could just instantiate an audit object and send it to the audit service. 

Just a thought.

sioldham (approved):

PR #518, opened 2023-08-14
SAA-1081 auditing the associated waiting list identifier with the prisoner allocated audit event if there is one.

matt-crowson-moj:
This change moves away from using the audit event listener. The main reason being it is getting too difficult to determine the nature of the audit event.

Note: there is no integration test change here as part of this PR, the events raised stays the same the prisoner allocated event.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/WaitingList.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/WaitingList.kt
@@ -62,10 +62,23 @@ data class WaitingList(
 
   fun isStatus(vararg s: WaitingListStatus) = s.any { it == status }
 
-  fun allocated(allocation: Allocation) {
-    this.status = WaitingListStatus.ALLOCATED
-    this.allocation = allocation
-  }
+  fun allocated(allocation: Allocation) =
+    apply {
+      require(allocation.prisonCode() == prisonCode) {
+        "Allocation ${allocation.allocationId} prison does not match with waiting list $waitingListId"
+      }
+
+      require(allocation.prisonerNumber == prisonerNumber) {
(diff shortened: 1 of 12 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/WaitingList.kt
```diff
@@ -62,10 +62,23 @@ data class WaitingList(
 
   fun isStatus(vararg s: WaitingListStatus) = s.any { it == status }
 
-  fun allocated(allocation: Allocation) {
-    this.status = WaitingListStatus.ALLOCATED
-    this.allocation = allocation
-  }
+  fun allocated(allocation: Allocation) =
+    apply {
+      require(allocation.prisonCode() == prisonCode) {
```
matt-crowson-moj, line 67: Adding some checks against the supplied allocation for the waiting list to ensure they marry up.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityScheduleService.kt
```diff
@@ -130,13 +128,22 @@ class ActivityScheduleService(
...
+        .also {
+          require(it.size <= 1) {
+            "" +
+              "Prisoner has more than one APPROVED waiting list. A prisoner can only have one approved waiting list"
+          }
+        }
+        .singleOrNull()?.allocated(allocation)
+
+      repository.saveAndFlush(schedule)
+      auditService.logEvent(allocation.toPrisonerAllocatedEvent(maybeWaitingList?.waitingListId))
```
matt-crowson-moj, line 146: Moved auditing of this event out of the audit listener and into the server.

It is getting too tricky/difficult to determine specific changes within the audit listener.

PR #523, opened 2023-08-15
SAA-1030 intial commit for supporting declining waiting lists.

matt-crowson-moj:
Providing a reusable service layer function to decline PENDING and/or APPROVED waiting list applications.

This will be used by background jobs e.g. activities ending jobs and event handlers e.g. prisoner release event.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/WaitingListService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/WaitingListService.kt
@@ -72,7 +72,7 @@ class WaitingListService(
         requestedBy = request.requestedBy!!,
         comments = request.comments,
         createdBy = createdBy,
-        status = request.status!!,
+        initialStatus = request.status!!,
       ),
     )
       .also { log.info("Added ${request.status} waiting list application for prisoner ${request.prisonerNumber} to activity ${schedule.description}") }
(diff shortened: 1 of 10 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/WaitingListService.kt
```diff
@@ -72,7 +72,7 @@ class WaitingListService(
         requestedBy = request.requestedBy!!,
         comments = request.comments,
         createdBy = createdBy,
-        status = request.status!!,
+        initialStatus = request.status!!,
       ),
```
timharrison-moj, line 76: Status required in the request? i.e. cannot possibly be null?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/WaitingListService.kt
```diff
@@ -240,4 +240,24 @@ class WaitingListService(
 
     telemetryClient.trackEvent(event.value, propertiesMap, metricsMap())
   }
+
+  @Transactional
+  fun declinePendingOrApprovedApplicationsFor(prisonCode: String, prisonerNumbers: Set<String>, reason: String, declinedBy: String) {
+    waitingListRepository.findByPrisonCodeAndPrisonerNumberIn(prisonCode, prisonerNumbers)
```
timharrison-moj, line 246: Will we ever get a list of prisoners in events? Assuming this method is to handle the release/transfer type of events?

matt-crowson-moj: No not so.  We also have jobs.  E.g. prisoners expiring today one or more?

timharrison-moj (approved):
LGTM. Minor questions only.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/WaitingListService.kt
matt-crowson-moj: This change is really to allow some validation on the actual status when it is changed/updated.

PR #524, opened 2023-08-16
SAA-1108 declining affected waiting list records on receipt of prisoner release event.

matt-crowson-moj:
…


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/WaitingListService.kt
```diff
@@ -28,7 +27,6 @@ import java.time.LocalDateTime
 
 @Service
 @Transactional(readOnly = true)
-@PreAuthorize("hasAnyRole('ACTIVITY_HUB', 'ACTIVITY_HUB_LEAD', 'ACTIVITY_ADMIN')")
```
matt-crowson-moj, line 31: Pre-auth is handled in the controller i.e. one place so don't have to keep two in sync.

PR #525, opened 2023-08-16
SAA-1109 declining affected waiting list records on receipt of activities changed event on end.

matt-crowson-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/events/handlers/ActivitiesChangedEventHandler.kt
```diff
@@ -80,10 +83,18 @@ class ActivitiesChangedEventHandler(
     }
   }
 
-  private fun deallocatePrisonerAndRemoveFutureAttendances(event: ActivitiesChangedEvent) =
+  private fun deallocatePrisonerAndRemoveFutureAttendances(event: ActivitiesChangedEvent) {
+    waitingListService.declinePendingOrApprovedApplicationsFor(
+      event.prisonCode(),
+      setOf(event.prisonerNumber()),
```
matt-crowson-moj, line 89: @timharrison-moj I think you were right on a previous PR.  At the service layer in this scenario we will only ever have one prisoner number.

When ending an activity, we will be declining all for that activity in a pending or approved state, we don't care about the prisoner numbers.

Will be refactoring this in a later PR though.  This should be OK for now.

PR #527, opened 2023-08-17
SAA-1055: Create occurrences asynchronously when total instances > 500

iterative-it:
There is a need by chaplains in prisons to create very large, repeating group appointments. These are for large worship or prayer meetings e.g. Friday prayers and Sunday services.

These can be up to 360 people and repeat weekly for a year. Such an appointment would result in 18,720 appointment allocation records and instances synched to NOMIS.

Creating these large appointments can take up to a minute. Efforts were made to optimise this via JPA bulk inserts but these were not successful. As a result, a new approach was taken where for these very large appointments, only the first occurrence is created. The remaining occurrences are created via an asynchronous job. This PR adds this job.

Once merged, this pattern will need to be replicated for editing and cancelling these very large appointments.


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/integration/AppointmentIntegrationTest.kt
```diff
@@ -232,6 +233,99 @@ class AppointmentIntegrationTest : IntegrationTestBase() {
...
+
+    // Synchronous creation. First occurrence and allocations only
+    val appointment = webTestClient.createAppointment(request)!!
+    var allocationIds = appointment.occurrences.flatMap { it.allocations.map { allocation -> allocation.id } }
+    appointment.occurrences hasSize 1
+    allocationIds hasSize 3
+
+    // Wait for remaining occurrences to be created
+    Thread.sleep(1000)
+    val appointmentDetails = webTestClient.getAppointmentById(appointment.id)!!
```
timharrison-moj, line 314: Might this timeout be unreliable?

timharrison-moj (approved):
LGTM

PR #534, opened 2023-08-21
SAA-1123 delete cascade activity

timharrison-moj:
Will completely remove an activity and all child entities - schedules, pay, allocations, slots, instances and attendance.
Protected by ROLE_NOMIS_ACTIVITIES and intended to be called by the migration service to clean up a failed migration if that becomes necessary.


matt-crowson-moj (approved):
LGTM

PR #535, opened 2023-08-22
SDIT-1039 Workaround Hibernate bug which returned duplicate records

mikehalmamoj:
Fixes a problem with attendance sync. that we were experiencing in dev - an offender that was allocated, released, returned to prison on a new booking and reallocated to the same activity always failed attendance sync. because we tried to attend against the old allocation. (A4744DZ).


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/SynchronisationService.kt
```diff
@@ -10,6 +10,5 @@ import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.repository.Atte
 class SynchronisationService(private val repository: AttendanceSyncRepository) {
   fun findAttendanceSync(attendanceId: Long): AttendanceSync? =
     repository.findAllByAttendanceId(attendanceId)
-      .maxByOrNull { it.allocationStartDate }
       ?.toModel()
```
mikehalmamoj, line 13: Previously we were grabbing all allocations and just returning the latest. However there was a bug in hibernate that was returning the same allocation in all restults - very confident it was a bug because running the Hibernate generated SQL didn't result in this problem.

PR #542, opened 2023-08-22
SAA-1096 - Added appointments metrics

sioldham:


iterative-it (approved):
LGTM - looking forward to seeing these coming through in app insights!

PR #544, opened 2023-08-23
SAA-1133 fix to prevent activating future pending allocations.

matt-crowson-moj:
This fixes an issue whereby the job for activating pending allocations was activating future pending allocations.

The knock effect of this was attendance records being created for future allocations.  We now only look at past or present pending allocations.

We are looking at past pending allocations to handle the situation of a failure and we need to rerun the job at a later point in time.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ManageAllocationsService.kt
```diff
@@ -57,12 +57,16 @@ class ManageAllocationsService(
...
-  private fun activatePendingAllocations() =
+  /**
+   * We consider pending allocations before today in the event we need to (re)run due to something out of a control e.g.
+   * a job fails to run due to a cloud platform issue.
+   */
+  private fun activatePendingAllocationsOnOrBeforeToday() =
     LocalDate.now().let { today ->
       forEachRolledOutPrison()
         .forEach { prison ->
           activityRepository.getAllForPrisonAndDate(prison.code, today).forEach { activity ->
```
matt-crowson-moj, line 68: Interested to hear thoughts on whether we think it would be worthwhile revisiting the query used here.

This could be expensive/slow as data builds up.  We are applying filters across all allocations when we could limit the number of rows in a repo query.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/integration/ManageAllocationsJobIntegrationTest.kt
```diff
@@ -115,15 +115,21 @@ class ManageAllocationsJobIntegrationTest : IntegrationTestBase() {
 
   @Sql("classpath:test_data/seed-allocations-pending.sql")
   @Test
-  fun `pending allocation is activated`() {
+  fun `pending allocations on or before today are activated`() {
     with(allocationRepository.findAll()) {
-      size isEqualTo 1
-      prisoner("A11111A") isStatus PENDING
+      size isEqualTo 3
+      prisoner("PAST") isStatus PENDING
```
matt-crowson-moj, line 121: Extended so we now have past, present and future pending allocations.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ManageAllocationsService.kt
timharrison-moj: Think we want to ignore ended activities, and possibly also avoid reactivating old allocations if they are already ended by also checking the end date is null or >= today? Most efficient way would be a view, but maybe not before rollout.

timharrison-moj (approved):
LGTM. One suggestion.

PR #545, opened 2023-08-23
Set Time-To-Live on cronjobs, to keep a 4 day history of jobs

smcveigh941:
4 days is the value CP have gone with for other jobs e.g. preprod database restore


matt-crowson-moj (approved):
LGTM

PR #549, opened 2023-08-25
SAA-1142: Keep description and summary the same.

timharrison-moj:
For migrated activities.
This keeps them consistent with other activities and prevents an odd description showing on forms that use this field.


matt-crowson-moj (approved):
LGTM

PR #552, opened 2023-08-29
SAA-991: Add option for setting location to Off Wing

smcveigh941:


timharrison-moj (approved):
LGTM. Branch needs an update though.

PR #556, opened 2023-08-30
SAA-1055: Partial async update and cancel

iterative-it:
Adds partial async support for update and cancel appointment actions


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/job/CancelAppointmentOccurrencesJob.kt
```diff
@@ -0,0 +1,69 @@
...
+    cancelled: LocalDateTime,
+    cancelledBy: String,
+    cancelOccurrencesCount: Int,
+    cancelInstancesCount: Int,
+    startTimeInMs: Long,
+  ) {
+    jobRunner.runJob(
+      JobDefinition(JobType.CANCEL_APPOINTMENT_OCCURRENCES) {
+        log.info("Cancelling remaining occurrences for appointment with id $appointmentId")
+        val elapsed = measureTimeMillis {
```
matt-crowson-moj, line 52: I like the `measureTimeMillis { ... }` we should probably do this in most if not all jobs (long running processes).

matt-crowson-moj (approved):
LGTM

PR #559, opened 2023-08-31
SAA-1150 removes pending allocations prior to deallocation on prisoner release event.

matt-crowson-moj:
The is the first of several tickets related to removal of pending allocations prior to deallocation of prisoners.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/ActivitySchedule.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/ActivitySchedule.kt
@@ -341,6 +341,18 @@ data class ActivitySchedule(
       addSlot(it.first, it.second.first, it.second.second, updates[it]!!)
     }
   }
+
+  fun removePending(allocation: Allocation) {
+    require(allocations.contains(allocation)) {
+      "Allocation ${allocation.allocationId} cannot be removed. It is not associated with the schedule."
+    }
+
+    require(allocation.status(PrisonerStatus.PENDING)) {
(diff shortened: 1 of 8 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Allocation.kt
```diff
@@ -252,7 +252,11 @@ data class Allocation(
 }
 
 enum class PrisonerStatus {
-  ACTIVE, PENDING, SUSPENDED, AUTO_SUSPENDED, ENDED
+  ACTIVE, PENDING, SUSPENDED, AUTO_SUSPENDED, ENDED;
+
+  companion object {
+    fun allExcuding(vararg status: PrisonerStatus) = entries.filterNot { status.contains(it) }.toTypedArray()
```
matt-crowson-moj, line 258: Not 100% sure on this, may change in a future PR.  I have to more areas to cover activities change event and the allocation job.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/ActivitySchedule.kt
```diff
@@ -341,6 +341,14 @@ data class ActivitySchedule(
       addSlot(it.first, it.second.first, it.second.second, updates[it]!!)
     }
   }
+
+  fun remove(allocation: Allocation) {
+    require(allocations.contains(allocation)) {
+      "The supplied allocation cannot be removed because it is not allocated to this schedule"
+    }
+
+    allocations.remove(allocation)
```
matt-crowson-moj, line 350: So this may need some stricter rules applying e.g. can only remove pending for example.

sioldham (approved):

PR #567, opened 2023-09-05
SAA-1159 split transactions within a potentially long-running job

timharrison-moj:
Problem: The createScheduledInstances job was running as a single, long transaction, and immediately after a prison migration this is potentially a long job.  It was creating sync events and the sync service was detecting them and calling back for data before it had been committed.

This change alters the transaction boundaries such that each activity is treated as its own transaction and committed immediately after raising an event.

It also fixes some issues with the job runner - this job was recorded two job results every time it ran, so this is also now resolved.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ManageScheduledInstancesService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ManageScheduledInstancesService.kt
@@ -3,100 +3,47 @@ package uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.service
 import jakarta.persistence.EntityNotFoundException
 import org.slf4j.LoggerFactory
 import org.springframework.beans.factory.annotation.Value
+import org.springframework.stereotype.Component
 import org.springframework.stereotype.Service
+import org.springframework.transaction.annotation.Propagation
 import org.springframework.transaction.annotation.Transactional
-import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.entity.JobType
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.entity.RolloutPrison
-import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.job.JobDefinition
-import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.job.SafeJobRunner
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.repository.ActivityRepository
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.repository.ActivityScheduleRepository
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ManageScheduledInstancesService.kt
```diff
@@ -3,100 +3,47 @@ package uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.service
...
-  * bank holidays, except for those activities that are marked as Ok to run on bank holidays.
-  *
-  * It will still create scheduled instances where an activity schedule has been suspended between two dates
-  * i.e. where a row exists in the ACTIVITY_SCHEDULE_SUSPENSION table because this table should always be
-  * empty - there is no current way in the UI to populate it.
-  */
+  @Transactional(propagation = Propagation.REQUIRED)
   fun create() {
-    jobRunner.runJob(
-      JobDefinition(
```
timharrison-moj, line 49: This was an additional runJob() - there is already one of these in the call chain above it, so removed. It was recording the job as running twice.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ManageScheduledInstancesService.kt
```diff
@@ -3,100 +3,47 @@ package uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.service
...
-              block = { createInstancesForActivitySchedule(prison, basic.activityScheduleId, listOfDatesToSchedule) },
-              success = "Scheduling sessions of ${prison.code} ${basic.summary}",
-              failure = "Failed to schedule ${prison.code} ${basic.summary}",
-            )
-          }
+    rolloutPrisonRepository.findAll()
+      .filter { it.isActivitiesRolledOut() }
+      .forEach { prison ->
+        log.info("Scheduling activities for prison ${prison.description} until $endDay")
+        val activities = activityRepository.getBasicForPrisonBetweenDates(prison.code, today, endDay)
```
timharrison-moj, line 38: This needs to be transactional - because it uses Hibernate filters to limit to the instances to the future.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ManageScheduledInstancesService.kt
```diff
@@ -3,100 +3,47 @@ package uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.service
...
+        activities.forEach { basic ->
+          continueToRunOnFailure(
+            block = { transactionHandler.createInstancesForActivitySchedule(prison, basic.activityScheduleId, listOfDatesToSchedule) },
+            success = "Scheduled instances for ${prison.code} ${basic.summary}",
```
timharrison-moj, line 42: This also needs to be transactional - and we want to commit after each activity.

iterative-it (approved):
This looks really good and answers some questions I had around the pattern for this. A good reference for future similar work

PR #570, opened 2023-09-07
Fix/saa 1150 relax allocation date validation on deallocate

matt-crowson-moj:
This change relaxes the rules around setting the allocation end date on the back of a prisoner release event.  Note this cannot be done via an API call.

A prisoner can be released before a (pending) allocation starts.  This means the end date can technically be before the start date in this situation only and so should be allowed.

This also mirrors the behaviour of NOMIS.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Appointment.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Appointment.kt
@@ -1,6 +1,7 @@
 package uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.entity
 
 import jakarta.persistence.CascadeType
+import jakarta.persistence.Column
 import jakarta.persistence.Entity
 import jakarta.persistence.EnumType
 import jakarta.persistence.Enumerated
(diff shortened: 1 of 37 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/events/handlers/PrisonerAllocationHandler.kt
```diff
@@ -0,0 +1,78 @@
...
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.entity.PrisonerStatus
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.enumeration.ServiceName
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.repository.AllocationRepository
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.repository.AttendanceRepository
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.service.WaitingListService
+import java.time.LocalDate
+import java.time.LocalDateTime
+
+@Component
+class PrisonerAllocationHandler(
```
matt-crowson-moj, line 16: Pulled out duplicate deallocation logic into a separate component.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Allocation.kt
```diff
@@ -52,7 +52,7 @@ data class Allocation(
 
   var endDate: LocalDate? = null
     set(value) {
-      require(value == null || value >= startDate) {
+      require(value == null || value >= startDate || prisonerStatus == PrisonerStatus.ENDED) {
```
matt-crowson-moj, line 55: This is where the rule is relaxed.

iterative-it (approved):
LGTM - nice refactor

PR #571, opened 2023-09-07
SAA-1162: Rename entities and services

iterative-it:
This PR follows on from the renaming of the appointments database tables. It renames the entities, repositories and services in the code base. It does not rename the models or the resources as that would break compatibility with the UI. That will be done in the next PR.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Appointment.kt
```diff
@@ -65,128 +57,136 @@ data class Appointment(
   @JoinColumn(name = "appointment_host_id")
   var appointmentHost: AppointmentHost? = null,
 
-  val internalLocationId: Long?,
+  var internalLocationId: Long?,
```
matt-crowson-moj, line 60: I presume making things `var's` is for future changes?

matt-crowson-moj (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Appointment.kt
iterative-it: Ah this is due to the rename of Appointment -> Appointment Series, an entity that should be immutable and Appointment Occurrence -> Appointment. GitHub understandably picks that up as a change to the original Appointment entity making it mutable.

PR #572, opened 2023-09-08
SAA-1178 fix activity summary view to join on correct activity and activity schedule columns.

matt-crowson-moj:


src/main/resources/migrations/common/V2023.09.08__v_activity_summary_fix_join_columns.sql
```diff
@@ -0,0 +1,26 @@
+CREATE OR REPLACE VIEW v_activity_summary
+AS
+select a.activity_id                                           id,
+       a.activity_category_id,
+       a.prison_code,
+       a.description                                           activity_name,
+       s.capacity,
+       (select count(*)
+        from allocation alloc
+        join activity_schedule s on s.activity_schedule_id = alloc.activity_schedule_id
```
matt-crowson-moj, line 10: Fix 1 here.

src/main/resources/migrations/common/V2023.09.08__v_activity_summary_fix_join_columns.sql
```diff
@@ -0,0 +1,26 @@
...
+        where s.activity_id = a.activity_id
+          and alloc.prisoner_status != 'ENDED')                allocated,
+       (select count(*)
+        from waiting_list w
+        where w.activity_id = a.activity_id
```
matt-crowson-moj, line 15: Fix 2 here.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/integration/PrisonIntegrationTest.kt
```diff
@@ -75,7 +75,7 @@ class PrisonIntegrationTest : IntegrationTestBase() {
         activityName = "Maths Level 1",
         category = educationCategory,
         capacity = 10,
-        allocated = 2,
+        allocated = 4,
```
matt-crowson-moj, line 78: You would need to look at the associated seed file to verify this.

From what I can tell looking at the seed file the number allocated count should have been 4 and not 2.

The seed file dates back to when you could have multiple schedules for an activity which turns out to be a good candidate for verifying the query changes i.e. we have a schedule where the id does not match that of the activity.  This was previously hidden due to the join on activity id being the same as the schedule id.

iterative-it (approved):
LGTM

PR #574, opened 2023-09-08
SAA-1162: Rename model and resources

iterative-it:
**Please don't merge this until the UI and Syscon PRs are ready.**

Final appointments rename work. Requires the related UI PR to be merged at the same time. Applies the following renaming and consolidation to the API endpoints and model:

- Renames all models in line with UI nomenclature
- The appointment set model (formerly bulk appointment) now shows a direct relationship between appointment set -> appointment without the appointment series in between. This is due to appointment sets being another way of aggregating appointments comparable to an appointment series.
- The appointment model now has either an appointment series summary or an appointment set summary depending on what the parent aggregation is. The schedule from the appointment series is now more correctly a child of the new appointment set summary
- Combines controllers and services based on entities
- Uses /entity/details for detail endpoints rather than /entity-details
- Updates all documentation and tests based on new taxonomy and relationships


matt-crowson-moj (approved):
LGTM

PR #577, opened 2023-09-11
SAA-1122: Added attendance summary endpoint

leejacobson-moj:
## Overview

- Adds new endpoint, `/scheduled-instances/attendance-summary` for retrieving details of all schedule instances running on a given date.

- Adds `date` filter to, `/schedules/{scheduleId}/allocations`. This provides an efficient way to retrieve an activity's allocations for a given date.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/resource/ActivityScheduleController.kt
```diff
@@ -108,7 +110,14 @@ class ActivityScheduleController(
...
     @Parameter(description = "If true will only return active allocations. Defaults to true.")
     activeOnly: Boolean?,
-  ) = scheduleService.getAllocationsBy(scheduleId, activeOnly ?: true)
+    @RequestParam(
+      value = "date",
+      required = false,
+    )
+    @DateTimeFormat(iso = DateTimeFormat.ISO.DATE)
+    @Parameter(description = "If provided will filter allocations by the given date. Format YYYY-MM-DD.")
+    date: LocalDate?,
```
leejacobson-moj, line 119: I think we should consider replacing the use of the `activeOnly` param here with a date param.

`activityOnly` can be achieved by passing today's date and fetching allocations active today, but a date param adds additional flexibility when fetching passed / future scheduled instances.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/model/AttendanceSummaryDetails.kt
```diff
@@ -0,0 +1,24 @@
...
+  val notRecorded: Long? = null,
+
+  @Schema(description = "The number of attendance recorded marked as attended", example = "2")
+  val attended: Long? = null,
+
+  @Schema(description = "The number of attendance recorded marked as absence", example = "1")
+  val absences: Long? = null,
+
+  @Schema(description = "The number of attendance recorded marked as paid", example = "2")
+  val paid: Long? = null,
```
leejacobson-moj, line 23: These are nullable because attendance records may not exist (for future instances). In these cases the UI needs to know if the difference between a 0 and null attendance count.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityScheduleService.kt
```diff
@@ -89,13 +90,18 @@ class ActivityScheduleService(
...
+  fun getAllocationsBy(scheduleId: Long, activeOnly: Boolean = true, activeOn: LocalDate? = null): List<Allocation> {
+    val activitySchedule = repository.getActivityScheduleByIdWithFilters(
+      scheduleId,
+      allocationsActiveOnDate = activeOn,
+    ) ?: throw EntityNotFoundException("Activity Schedule $scheduleId not found")
+
+    return activitySchedule
       .checkCaseloadAccess()
       .allocations()
       .filter { !activeOnly || !it.status(PrisonerStatus.ENDED) }
```
leejacobson-moj, line 102: I've left it as to not break existing implementations, but I don't like this filter.

For a long running activities most allocations are likely to have ended. Pulling these records from the DB then filtering the results is could be done more efficiently with an entity filter.

Again, I think replacing the allocate status filter here for a date filter would be more flexible and allow us to filter results for retrieving past / future instances.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/model/ScheduledInstanceAttendanceSummary.kt
```diff
@@ -0,0 +1,73 @@
...
+    example = "2023-03-30",
+  )
+  @JsonFormat(pattern = "yyyy-MM-dd")
+  val sessionDate: LocalDate,
+
+  @Schema(description = "The start time of the scheduled instance", example = "09:00")
+  @JsonFormat(pattern = "HH:mm")
+  val startTime: LocalTime,
+
+  @Schema(description = "The start time of the scheduled instance", example = "09:00")
```
matt-crowson-moj, line 52: Docs not quite right on the end time.  Should say `The end time ...`  Also be good to make the end time example different to that of the start time.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/model/ScheduledInstanceAttendanceSummary.kt
```diff
@@ -0,0 +1,73 @@
...
+  @Schema(description = "Flag to indicate if the location of the activity is on wing", example = "false")
+  var onWing: Boolean,
+
+  @Schema(description = "Flag to indicate if the location of the activity is off wing and not in a listed location", example = "false")
+  var offWing: Boolean,
+
+  @Schema(description = "The NOMIS internal location for this schedule")
+  var internalLocation: InternalLocation? = null,
+
+  @Schema(description = "Flag to indicate if the scheduled instance has been cancelled")
```
matt-crowson-moj, line 68: Perhaps add `example = "false"` to cancelled? Meh.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/resource/ScheduledInstanceController.kt
```diff
@@ -263,4 +268,50 @@ class ScheduledInstanceController(
     )
     scheduleInstanceCancelRequest: ScheduleInstanceCancelRequest,
   ): ResponseEntity<Any> = scheduledInstanceService.cancelScheduledInstance(instanceId, scheduleInstanceCancelRequest).let { ResponseEntity.noContent().build() }
+
+  @GetMapping(value = ["/attendance-summary"])
+  @ResponseBody
+  @Operation(
+    summary = "Attendance summary of activity sessions for a given date",
+    description = "Attendance summary of activity sessions for a given date",
```
matt-crowson-moj, line 276: Is it worth having a both summary and description if they are identical?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/resource/ScheduledInstanceController.kt
```diff
@@ -263,4 +268,50 @@ class ScheduledInstanceController(
...
+    ],
+  )
+  @CaseloadHeader
+  @PreAuthorize("hasAnyRole('PRISON', 'ACTIVITY_ADMIN')")
+  fun attendanceSummary(
+    @RequestParam(value = "date", required = true)
+    @DateTimeFormat(iso = DateTimeFormat.ISO.DATE)
+    @Parameter(description = "The date of the attendance summary. Format YYYY-MM-DD.")
+    date: LocalDate,
+    @RequestHeader(CASELOAD_ID, required = true)
```
matt-crowson-moj, line 314: Do you need the `RequestHeader` line?  I can't see this used anywhere else in the API?  Note I am only questioning the inclusion of the line 314 and not the `prisonCode` param.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityScheduleService.kt
matt-crowson-moj: Agreed.

This was done quite a while ago before the filter work.  There are quite a few places where improvements could/should be made now we know more/better.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/resource/ScheduledInstanceController.kt
matt-crowson-moj: I am asking because I wonder if we are missing in other places perhaps.

src/main/resources/migrations/common/V2023.09.11__scheduled_instance_attendance_summary.sql
```diff
@@ -0,0 +1,49 @@
...
+	acts.internal_location_code,
+	acts.internal_location_description,
+	si.cancelled,
+	COALESCE((
+	    SELECT COUNT(DISTINCT al.allocation_id)
+		FROM allocation al
+		WHERE al.activity_schedule_id = si.activity_schedule_id
+		AND al.start_date <= si.session_date
+		AND (
+		    (al.end_date IS NULL AND al.prisoner_status != 'ENDED')
```
matt-crowson-moj, line 25: Do you need the `AND al.prisoner_status != ENDED` check?  i.e. `end_date is null or end_date >= si.session_date`

ENDED will only be present when an end date is present.  One row exists in dev like this at the mo due to a bug.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/resource/ScheduledInstanceController.kt
leejacobson-moj: Initially I made it a request param, but I questioned if that was necessary since we're sending the active caseload in the header anyway. But thinking about it again I we probably assume client tokens will send that header.

I've updated to use a request param.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/model/AttendanceSummaryDetails.kt
```diff
@@ -0,0 +1,24 @@
+package uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.model
+
+import io.swagger.v3.oas.annotations.media.Schema
+
+@Schema(description = "Attendance summary details for a scheduled instance")
```
matt-crowson-moj, line 5: Sorry one more question.

Given this says it is the summary for an instance should this have the instance id included?  I realise it is used in another mddel.

matt-crowson-moj (approved):
LGTM - one more small question, not essential.

matt-crowson-moj (approved):
LGTM

PR #579, opened 2023-09-13
SAA-1162: Indexes and roles

iterative-it:
- Add indexes on columns used in soft delete where clauses
- Add NOMIS_ACTIVITIES role to endpoint used by sync

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/resource/AppointmentInstanceController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/resource/AppointmentInstanceController.kt
@@ -62,7 +62,7 @@ class AppointmentInstanceController(
     ],
   )
   @CaseloadHeader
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/resource/AppointmentInstanceController.kt
```diff
@@ -62,7 +62,7 @@ class AppointmentInstanceController(
     ],
   )
   @CaseloadHeader
-  @PreAuthorize("hasAnyRole('PRISON', 'ACTIVITY_ADMIN')")
+  @PreAuthorize("hasAnyRole('PRISON', 'ACTIVITY_ADMIN', 'NOMIS_ACTIVITIES')")
```
matt-crowson-moj, line 65: Could perhaps add a role check to the controller test?

matt-crowson-moj (approved):
LGTM

PR #590, opened 2023-09-20
SAA-204 update risley pay bands

timharrison-moj:
This changes Risley's pay bands from Low, Medium, High to the default 1-10 used by NOMIS in all emvironments.
(similar to how LEI is setup in our service).

It also updates any Activities and Allocations as part of the migration.


yasinmustafa (approved):
LGTM

PR #593, opened 2023-09-20
SAA-1208 adding a new transaction handler around attendance changes to guarantee attendances are updated before events are emitting attendance amended events.

matt-crowson-moj:
A problem has shown itself whereby (bulk) attendance amendment events are being processed by the sync service before the changes had fully committed in our service DB.  This resulted in the activities service and NOMIS being out of sync with each.

The crux of the changes here involve wrapping the attendance changes in a completely independent new transaction which has to complete after which we publish the events in the service/component and not use the attendance entity listener.

Hopefully this is a short term solution with regards to raising attendance amended events until we have time to revisit and address  a more complete solution towards the raising of events in general.  Ideally we should be looking to move away from entity listeners e.g. use of `TransactionalEventListener` which can determine when a transaction has committed so could replace the use of entity listeners perhaps.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/AttendanceEntityListener.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/AttendanceEntityListener.kt
@@ -1,7 +1,6 @@
 package uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.entity
 
 import jakarta.persistence.PostPersist
-import jakarta.persistence.PostUpdate
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory
 import org.springframework.beans.factory.annotation.Autowired
@@ -25,15 +24,6 @@ class AttendanceEntityListener {
       OutboundEvent.PRISONER_ATTENDANCE_CREATED,
       attendance.attendanceId,
       "Failed to send prisoner attendance created event for attendance ID ${attendance.attendanceId}",
-    )
-  }
-
-  @PostUpdate
-  fun onUpdate(attendance: Attendance) {
-    send(
-      OutboundEvent.PRISONER_ATTENDANCE_AMENDED,
-      attendance.attendanceId,
-      "Failed to send prisoner attendance amended event for attendance ID ${attendance.attendanceId}",
     )
   }
 
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/AttendancesService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/AttendancesService.kt
@@ -12,6 +12,8 @@ import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.repository.Atte
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.repository.AttendanceRepository
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.repository.ScheduledInstanceRepository
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.repository.findOrThrowNotFound
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.service.events.OutboundEvent
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.service.events.OutboundEventsService
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.util.transform
 import java.time.LocalDate
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.model.AllAttendance as ModelAllAttendance
@@ -23,6 +25,8 @@ class AttendancesService(
   private val attendanceRepository: AttendanceRepository,
   private val attendanceReasonRepository: AttendanceReasonRepository,
   private val caseNotesApiClient: CaseNotesApiClient,
+  private val transactionHandler: TransactionHandler,
+  private val outboundEventsService: OutboundEventsService,
 ) {
 
   companion object {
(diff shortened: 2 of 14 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/AttendanceEntityListener.kt
```diff
@@ -28,15 +27,6 @@ class AttendanceEntityListener {
     )
   }
 
-  @PostUpdate
-  fun onUpdate(attendance: Attendance) {
-    send(
-      OutboundEvent.PRISONER_ATTENDANCE_AMENDED,
```
matt-crowson-moj, line 34: We now have to raise this manually when we make any changes to existing attendance records.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ScheduledInstanceService.kt
```diff
@@ -62,16 +63,24 @@ class ScheduledInstanceService(
...
     log.info("Uncancelling scheduled instance $id")
 
-    val scheduledInstance = repository.findById(id)
-      .orElseThrow { EntityNotFoundException("Scheduled Instance $id not found") }
+    val uncancelledInstance = transactionHandler.new {
+      val scheduledInstance = repository.findById(id)
+        .orElseThrow { EntityNotFoundException("Scheduled Instance $id not found") }
 
-    scheduledInstance.uncancelSessionAndAttendances()
+      scheduledInstance.uncancelSessionAndAttendances()
```
matt-crowson-moj, line 70: This isn't quite right.  Need to return the attendances that have been uncancelled as well.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ScheduledInstanceService.kt
```diff
@@ -80,21 +89,29 @@ class ScheduledInstanceService(
...
+      val scheduledInstance = repository.findById(instanceId)
+        .orElseThrow { EntityNotFoundException("Scheduled Instance $instanceId not found") }
 
-    scheduledInstance.cancelSessionAndAttendances(
-      reason = scheduleInstanceCancelRequest.reason,
-      by = scheduleInstanceCancelRequest.username,
-      cancelComment = scheduleInstanceCancelRequest.comment,
-      cancellationReason = attendanceReasonRepository.findByCode(AttendanceReasonEnum.CANCELLED),
-    )
+      scheduledInstance.cancelSessionAndAttendances(
```
matt-crowson-moj, line 96: This isn't quite right.  Need to return the attendances that have been cancelled as well.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/TransactionHandler.kt
```diff
@@ -0,0 +1,15 @@
...
+import org.springframework.transaction.annotation.Transactional
+
+/**
+ * To be used in services or components where a new transaction is required to ensure any DB changes are committed prior
+ * to taking any further action e.g. emitting new/update/delete events.
+ */
+@Component
+class TransactionHandler {
+  @Transactional(propagation = Propagation.REQUIRES_NEW)
+  fun <T> new(block: () -> T): Result<T> = runCatching { block() }
```
timharrison-moj, line 14: I think the `runCatching` block will always prevent the transaction being rolled back?

timharrison-moj: The thing which causes the rollback is a run time exception.

matt-crowson-moj: 🤔 you might be right there will take out to be safe, easy enough to do.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/AttendancesService.kt
```diff
@@ -35,25 +39,33 @@ class AttendancesService(
...
+          newIssuePayment = updateRequest.issuePayment,
+          newIncentiveLevelWarningIssued = updateRequest.incentiveLevelWarningIssued,
+          newCaseNoteId = updateRequest.mayBeCaseNote(attendance.prisonerNumber)?.caseNoteId,
+          newOtherAbsenceReason = updateRequest.otherAbsenceReason,
+        )
+
+        attendanceRepository.saveAndFlush(attendance)
+      }.map { it.attendanceId }
+    }.getOrThrow()
+
```
timharrison-moj, line 63: Not sure what the `getOrThrow()` does here?

timharrison-moj: I like the idea of this block being transactional - but its hard to follow.  If it works though, happy to live with it.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/events/handlers/ActivitiesChangedEventHandler.kt
```diff
@@ -56,9 +62,15 @@ class ActivitiesChangedEventHandler(
...
+        transactionHandler.new {
+          allocationRepository.findByPrisonCodeAndPrisonerNumber(event.prisonCode(), event.prisonerNumber())
+            .suspendPrisonersAllocations(now, event)
+            .suspendPrisonersFutureAttendances(now, event)
+        }.let { updatedAttendances ->
+          updatedAttendances.forEach {
+            outboundEventsService.send(OutboundEvent.PRISONER_ATTENDANCE_AMENDED, it.attendanceId)
+          }.also { log.info("Sending attendance amended events.") }
+        }
       }
```
timharrison-moj, line 74: Did this bit need to change? Only dealing with a single prisoner's allocations and attendances.  For tomorrow!

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/AttendancesService.kt
matt-crowson-moj: There is no getOrThrow now as taken the result out.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/integration/AttendanceIntegrationTest.kt
```diff
@@ -46,9 +45,6 @@ class AttendanceIntegrationTest : IntegrationTestBase() {
   @Autowired
   private lateinit var attendanceRepository: AttendanceRepository
 
-  @Autowired
-  private lateinit var attendancesService: ManageAttendancesService
-
```
timharrison-moj, line 51: Unused?

timharrison-moj (commented):
LGTM. Would like to test it locally (as discussed).

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/integration/AttendanceIntegrationTest.kt
matt-crowson-moj: Yes.  I suspect a hangover from something else and probably not removed.

timharrison-moj (approved):
LGTM. Matt demonstrated locally.

PR #609, opened 2023-09-30
SAA-1252 reducing the event logging noice in the API.

matt-crowson-moj:
Changing logging calls around event handling to use debug log level to reduce the amount of unwanted noise in the production logs.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/events/InboundEventsListener.kt
```diff
@@ -31,7 +31,7 @@ class InboundEventsListener(
   @SqsListener("activities", factory = "hmppsQueueContainerFactoryProxy")
   @WithSpan(value = "Digital-Prison-Services-hmpps_activities_management_queue", kind = SpanKind.SERVER)
   internal fun onMessage(rawMessage: String) {
-    log.info("Inbound event raw message $rawMessage")
+    log.debug("Inbound event raw message $rawMessage")
```
matt-crowson-moj, line 34: This one is particularly noisy

PR #610, opened 2023-09-30
SAA-1251: Remove appointment attendees for released prisoners

iterative-it:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/AppointmentAttendeeEntityListener.kt
```diff
@@ -53,7 +53,15 @@ class AppointmentAttendeeEntityListener {
   @PostUpdate
   fun onUpdate(entity: AppointmentAttendee) {
     runCatching {
-      outboundEventsService.send(OutboundEvent.APPOINTMENT_INSTANCE_UPDATED, entity.appointmentAttendeeId)
+      outboundEventsService.send(
+        when {
+          entity.isRemoved() -> OutboundEvent.APPOINTMENT_INSTANCE_CANCELLED
+          entity.isDeleted -> OutboundEvent.APPOINTMENT_INSTANCE_DELETED
+
```
timharrison-moj, line 60: Does the sync service need to change to listen/act on the cancelled and deleted events? Or does it already do so?

iterative-it: Already does it. The onDelete in this listener already published APPOINTMENT_INSTANCE_DELETED and AppointmentEntityListener published APPOINTMENT_INSTANCE_CANCELLED for each attendee. This is simply another way both events can be published. I confirmed both types of sync were working in prod on Friday.

In terms of the issue with these being published prior to the transaction completing, for cancellations and deletions that's OK as the sync service doesn't call any of our endpoints, it simply acts on the sync instruction using the existing mapping from our id to the NOMIS one. It _should_ all work ;)

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/AppointmentSearchService.kt
```diff
@@ -97,7 +98,8 @@ class AppointmentSearchService(
     val locationMap = locationService.getLocationsForAppointmentsMap(prisonCode)
 
     logAppointmentSearchMetric(principal, prisonCode, request, results.size, startTime)
-    return results.toResults(attendeeMap, referenceCodeMap, locationMap)
+
+    return results.filter { it.appointmentType == AppointmentType.GROUP || attendeeMap.containsKey(it.appointmentId) }.toResults(attendeeMap, referenceCodeMap, locationMap)
   }
```
timharrison-moj, line 103: Not sure I understand this change, Dave? Is it related to the manage job or an extra tweak?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ManageAppointmentService.kt
```diff
@@ -0,0 +1,46 @@
...
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.common.onOrBefore
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.repository.AppointmentAttendeeRemovalReasonRepository
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.repository.AppointmentRepository
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.repository.PRISONER_STATUS_RELEASED_APPOINTMENT_ATTENDEE_REMOVAL_REASON_ID
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.repository.findOrThrowNotFound
+import java.time.LocalDateTime
+
+@Service
+@Transactional
+class ManageAppointmentService(
```
timharrison-moj, line 19: With a @Transactional service, this job will do everything within a single transaction. If the sync service is relying on the data being committed for any events raised, it might well call back before the transaction has ended.  May need similar treatment as per the manageScheduledInstance job, with a sub-transaction?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/AppointmentSearchService.kt
iterative-it: That's fair. It's not really explained. The clean up job soft deletes attendees. It's OK for group appointments to have no attendees and to be returned as search results but migrated appointments (type == INDIVIDUAL) should not show on search results if their attendee has been soft deleted. This does that

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ManageAppointmentService.kt
iterative-it: Mentioned in a previous comment but helpful here too. In terms of the issue with these being published prior to the transaction completing, for cancellations and deletions that's OK as the sync service doesn't call any of our endpoints, it simply acts on the sync instruction using the existing mapping from our id to the NOMIS one. It should all work ;)

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/integration/AppointmentJobIntegrationTest.kt
```diff
@@ -0,0 +1,109 @@
...
+    lastMovementType = MovementType.RELEASE,
+    confirmedReleaseDate = LocalDate.now(),
+  )
+
+  private val prisonerReleasedOneWeekAgo = PrisonerSearchPrisonerFixture.instance(
+    prisonerNumber = "C3456DE",
+    inOutStatus = Prisoner.InOutStatus.OUT,
+    status = "INACTIVE OUT",
+    lastMovementType = MovementType.RELEASE,
+    confirmedReleaseDate = LocalDate.now().minusWeeks(1),
```
timharrison-moj, line 64: If ACTIVE IN/OUT are there any checks about which prison this person is the responsibility of now? i.e. is it different to the expected prison?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ManageAppointmentServiceTest.kt
```diff
@@ -0,0 +1,113 @@
...
+      lastMovementType = null,
+      confirmedReleaseDate = LocalDate.now().plusDays(2),
+    )
+
+    private val prisonerReleasedToday = PrisonerSearchPrisonerFixture.instance(
+      prisonerNumber = "A1234BC",
+      inOutStatus = Prisoner.InOutStatus.OUT,
+      status = "INACTIVE OUT",
+      lastMovementType = MovementType.RELEASE,
+      confirmedReleaseDate = LocalDate.now(),
```
timharrison-moj, line 52: Same here - could they be ACTIVE but at a different prison?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ManageAppointmentServiceTest.kt
```diff
@@ -0,0 +1,113 @@
...
+
+    @Test
+    fun `do not remove attendee records for appointments with start dates prior to prisoner confirmed release date`() {
+      val appointmentSeries = appointmentSeriesEntity(startDate = LocalDate.now().minusDays(1))
+      val appointment = appointmentSeries.appointments().first()
+
+      whenever(appointmentRepository.findAllByStartDate(LocalDate.now().minusDays(1))).thenReturn(appointmentSeries.appointments())
+      whenever(prisonerSearch.findByPrisonerNumbers(appointment.prisonerNumbers())).thenReturn(Mono.just(listOf(prisonerReleasedToday)))
+
+      service.manageAppointmentAttendees(LocalDateRange(LocalDate.now().minusDays(1), LocalDate.now().minusDays(1)))
```
timharrison-moj, line 91: Did anything come of looking into the `actualReleaseDate` attribute? I am pretty sure its set on release for people. The `confirmedReleaseDate` is a planned date .. and it may not actually happen. During CVL we saw this date change several times just prior to release, as if they planned it but then couldn't manage it so shifted it forward a day or two.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/integration/AppointmentJobIntegrationTest.kt
iterative-it: That's expected as another check soon but we couldn't fully agree on the logic yesterday so it's been left out for now. Once we verify these checks are working, it's a fairly small addition.

timharrison-moj (commented):
LGTM. Couple of questions only.
Very quick work - but should we have a break for Sunday and look at this again on Monday? Bit worried we'd be rushing this in to do it now?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ManageAppointmentServiceTest.kt
iterative-it: See above

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ManageAppointmentServiceTest.kt
iterative-it: Looking into it, there's no property called `actualReleaseDate`. `releaseDate` is described as "Actual of most likely Release Date" and `confirmedReleaseDate` as "Release Date Confirmed". `confirmedReleaseDate` is also what `isPermanentlyReleased` used so I thought I'd maintain consistency with activities and we'd change it across the board later if we decide to

timharrison-moj (dismissed):
Thanks for responses. LGTM.
Be nice to let Matt take a look too?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/AppointmentAttendee.kt
```diff
@@ -40,11 +43,26 @@ data class AppointmentAttendee(
...
+  var removalReason: AppointmentAttendeeRemovalReason? = null
+
+  var removedBy: String? = null
+
+  fun isRemoved() = removedTime != null && !isDeleted
+
+  var isDeleted: Boolean = false
+
+  fun remove(removedTime: LocalDateTime = LocalDateTime.now(), removalReason: AppointmentAttendeeRemovalReason, removedBy: String?) {
+    this.removedTime = removedTime
```
matt-crowson-moj, line 60: Not essential but could tighten up the removal of this with a little validation e.g. fail if already removed, fail if date in removed time in the past?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ManageAppointmentService.kt
```diff
@@ -0,0 +1,46 @@
...
+    val now = LocalDateTime.now()
+    val releasedAppointmentAttendeeRemovalReason = appointmentAttendeeRemovalReasonRepository.findOrThrowNotFound(PRISONER_STATUS_RELEASED_APPOINTMENT_ATTENDEE_REMOVAL_REASON_ID)
+    val user = "MANAGE_APPOINTMENT_SERVICE"
+
+    localDateRange.forEach { startDate ->
+      val appointments = appointmentRepository.findAllByStartDate(startDate)
+
+      val prisoners = prisonerSearch.findByPrisonerNumbers(appointments.flatMap { it.prisonerNumbers() }.distinct()).block()!!
+
+      val prisonersReleasedBeforeStartDate = prisoners.filter { it.isPermanentlyReleased() && it.confirmedReleaseDate?.onOrBefore(startDate) == true }.associateBy { it.prisonerNumber }
```
matt-crowson-moj, line 38: Don't think you need the `associateBy { it.prisonerNumber }` can use `map { it.prisonerNumber }` instead as you are only interested in the prisoner number looking at the code.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ManageAppointmentService.kt
```diff
@@ -0,0 +1,46 @@
...
+
+      appointments.flatMap { it.attendees() }.filter { prisonersReleasedBeforeStartDate.containsKey(it.prisonerNumber) }.forEach {
+        it.remove(now, releasedAppointmentAttendeeRemovalReason, user)
+        log.info("Removed appointment attendee '${it.appointmentAttendeeId}' from appointment '${it.appointment.appointmentId} as associated prisoner was released on or before '$startDate'")
+      }
+    }
+  }
+}
```
matt-crowson-moj, line 46: Perhaps add some extension functions on the filters to aid readability?

```
  fun manageAppointmentAttendees(localDateRange: LocalDateRange) {
    val now = LocalDateTime.now()
    val releasedAppointmentAttendeeRemovalReason = appointmentAttendeeRemovalReasonRepository.findOrThrowNotFound(PRISONER_STATUS_RELEASED_APPOINTMENT_ATTENDEE_REMOVAL_REASON_ID)
    val user = "MANAGE_APPOINTMENT_SERVICE"

    localDateRange.forEach { startDate ->
      val appointments = appointmentRepository.findAllByStartDate(startDate)

      val prisoners = prisonerSearch.findByPrisonerNumbers(appointments.flatMap { it.prisonerNumbers() }.distinct()).block()!!

      val prisonersReleasedBeforeStartDate = prisoners.permanentlyReleaseOnOrBefore(startDate).map { it.prisonerNumber }

      appointments.flatMap { it.attendees() }.attendeesMatching(prisonersReleasedBeforeStartDate).forEach {
        it.remove(now, releasedAppointmentAttendeeRemovalReason, user)
        log.info("Removed appointment attendee '${it.appointmentAttendeeId}' from appointment '${it.appointment.appointmentId} as associated prisoner was released on or before '$startDate'")
      }
    }
  }
  
  private fun List<Prisoner>.permanentlyReleaseOnOrBefore(date: LocalDate) = filter { it.isPermanentlyReleased() && it.confirmedReleaseDate?.onOrBefore(date) == true }
  
  private fun List<AppointmentAttendee>.attendeesMatching(releasedPrisoners: List<String>) = filter { releasedPrisoners.contains(it.prisonerNumber) }

```

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/AppointmentAttendeeTest.kt
```diff
@@ -17,4 +22,28 @@ class AppointmentAttendeeTest {
     val expectedModel = listOf(appointmentAttendeeModel())
     assertThat(appointmentSeriesEntity().appointments().first().attendees().toModel()).isEqualTo(expectedModel)
   }
+
+  @Test
+  fun `removal with soft delete reason`() {
+    val entity = appointmentSeriesEntity().appointments().first().attendees().first().apply {
+      remove(removalReason = appointmentAttendeeDeletedReason(), removedBy = "TEST.USER")
+    }
+    assertThat(entity.removedTime).isCloseTo(LocalDateTime.now(), Assertions.within(1, ChronoUnit.SECONDS))
```
matt-crowson-moj, line 31: Minor, missing import for `Assertions.` with `Assertions.within` meh.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/AppointmentSetTest.kt
```diff
@@ -196,6 +197,28 @@ class AppointmentSetTest {
...
+      "CREATE.USER" to userDetail(1, "CREATE.USER", "CREATE", "USER"),
+      "UPDATE.USER" to userDetail(2, "UPDATE.USER", "UPDATE", "USER"),
+    )
+    val prisonerMap = getPrisonerMap()
+
+    with(entity.toDetails(prisonerMap, referenceCodeMap, locationMap, userMap)) {
+      appointments hasSize 3
+    }
+
+    entity.appointments().first().apply { this.removeAttendee(this.attendees().first()) }
```
matt-crowson-moj, line 215: This could possibly do with an assertion to say the attendees is empty or use `single()` instead of `first()` when removing attendee.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/AppointmentSetTest.kt
```diff
@@ -196,6 +197,28 @@ class AppointmentSetTest {
...
+    val entity = appointmentSetEntity()
+    val referenceCodeMap = mapOf(entity.categoryCode to appointmentCategoryReferenceCode(entity.categoryCode))
+    val locationMap = mapOf(entity.internalLocationId!! to appointmentLocation(entity.internalLocationId!!, "TPR"))
+    val userMap = mapOf(
+      "CREATE.USER" to userDetail(1, "CREATE.USER", "CREATE", "USER"),
+      "UPDATE.USER" to userDetail(2, "UPDATE.USER", "UPDATE", "USER"),
+    )
+    val prisonerMap = getPrisonerMap()
+
+    with(entity.toDetails(prisonerMap, referenceCodeMap, locationMap, userMap)) {
```
matt-crowson-moj, line 211: Using `with` possibly overkill for just one assertion meh.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/integration/AppointmentJobIntegrationTest.kt
```diff
@@ -0,0 +1,109 @@
...
+  fun setUp() {
+    prisonerSearchApiMockServer.stubSearchByPrisonerNumbers(listOf("A1234BC", "B2345CD", "C3456DE"), listOf(activeInPrisoner, prisonerReleasedToday, prisonerReleasedOneWeekAgo))
+  }
+
+  @Sql("classpath:test_data/seed-manage-appointments-job.sql")
+  @Test
+  fun `remove attendees for released prisoners`() {
+    with(appointmentAttendeeRepository.findAll()) {
+      size isEqualTo 18
+      onEach { it.isRemoved() isEqualTo false }
```
matt-crowson-moj, line 77: There is an isBool infix function as well. e.g. 

`it.isRemoved() isBool false`

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/integration/AppointmentJobIntegrationTest.kt
```diff
@@ -0,0 +1,109 @@
...
+      size isEqualTo 11
+      onEach { it.isRemoved() isEqualTo false }
+      onEach { it.isDeleted isEqualTo false }
+    }
+
+    verify(eventsPublisher, times(7)).send(eventCaptor.capture())
+
+    with(eventCaptor.allValues.filter { it.eventType == "appointments.appointment-instance.deleted" }) {
+      size isEqualTo 7
+      Assertions.assertThat(map { it.additionalInformation }).containsExactlyElementsOf(
```
matt-crowson-moj, line 93: Missing import for `Assertions.`

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ManageAppointmentService.kt
```diff
@@ -0,0 +1,46 @@
...
+    val user = "MANAGE_APPOINTMENT_SERVICE"
+
+    localDateRange.forEach { startDate ->
+      val appointments = appointmentRepository.findAllByStartDate(startDate)
+
+      val prisoners = prisonerSearch.findByPrisonerNumbers(appointments.flatMap { it.prisonerNumbers() }.distinct()).block()!!
+
+      val prisonersReleasedBeforeStartDate = prisoners.filter { it.isPermanentlyReleased() && it.confirmedReleaseDate?.onOrBefore(startDate) == true }.associateBy { it.prisonerNumber }
+
+      appointments.flatMap { it.attendees() }.filter { prisonersReleasedBeforeStartDate.containsKey(it.prisonerNumber) }.forEach {
```
matt-crowson-moj, line 40: I don't know the numbers here but could this result in a large transaction if the numbers are high?

matt-crowson-moj, line 40: Perhaps do this one appointment at a time instead of flatten across all?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/AppointmentAttendee.kt
iterative-it: I've not added any date validation as we're thinking we may use this to remove historic appointments that should have been removed by NOMIS. In terms of overriding the removal reason, that's intentional. If the reason caused a soft delete, it wouldn't be available to remove again. If it was a removal similar to a suspension, it can either become permanent or be reversed.

There's an upcoming PR which also improves the logic around adding and removing attendees.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ManageAppointmentService.kt
iterative-it: Done with extension method covering point below

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ManageAppointmentService.kt
iterative-it: Good idea - done

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/AppointmentAttendeeTest.kt
iterative-it: Done

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/AppointmentSetTest.kt
iterative-it: This was cause I'd broken the arrange act assert pattern by checking my test data first, changing it and checking again. I've removed that now

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/AppointmentSetTest.kt
iterative-it: Done

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/integration/AppointmentJobIntegrationTest.kt
iterative-it: Done

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/integration/AppointmentJobIntegrationTest.kt
iterative-it: Done

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/job/ManageAppointmentAttendeesJob.kt
```diff
@@ -0,0 +1,23 @@
...
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.service.ManageAppointmentService
+import java.time.LocalDate
+
+@Component
+class ManageAppointmentAttendeesJob(
+  private val jobRunner: SafeJobRunner,
+  private val service: ManageAppointmentService,
+) {
+  @Async("asyncExecutor")
+  fun execute(daysBeforeNow: Long, daysAfterNow: Long) {
```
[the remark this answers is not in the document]
iterative-it: Added a range check in the service limiting it to max 60 days

matt-crowson-moj (approved):
LGTM

PR #615, opened 2023-10-04
SAA-1278 record activities change event as event of interest when can not determine the release. Should not throw an exception.

matt-crowson-moj:
Because an exception was being thrown the event is not being recorded as an event of interest.  It should be recorded as an event of interest so the prison can deal with it instead.

Note this is partial fix.  There is a problem with the logic for determining the actual release reason by the looks of it.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/events/handlers/ActivitiesChangedEventHandler.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/events/handlers/ActivitiesChangedEventHandler.kt
@@ -114,7 +114,10 @@ class ActivitiesChangedEventHandler(
       when {
         prisoner.isTemporarilyReleased() -> DeallocationReason.TEMPORARILY_RELEASED
         prisoner.isPermanentlyReleased() -> DeallocationReason.RELEASED
-        else -> throw IllegalStateException("Unable to determine release reason for prisoner ${event.prisonerNumber()}")
+        else -> {
+          log.warn("Unable to determine release reason for prisoner ${event.prisonerNumber()}, prisoner status ${prisoner.status}, at prison ${prisoner.prisonId} for event $event")
+          null
+        }
       }
     }
 }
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/events/handlers/ActivitiesChangedEventHandler.kt
```diff
@@ -114,7 +114,10 @@ class ActivitiesChangedEventHandler(
       when {
         prisoner.isTemporarilyReleased() -> DeallocationReason.TEMPORARILY_RELEASED
         prisoner.isPermanentlyReleased() -> DeallocationReason.RELEASED
-        else -> throw IllegalStateException("Unable to determine release reason for prisoner ${event.prisonerNumber()}")
+        else -> {
+          log.warn("Unable to determine release reason for prisoner ${event.prisonerNumber()} for event $event")
```
matt-crowson-moj, line 118: Now this doesn't throw an exception it will be recorded as an event of interest.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/events/handlers/ActivitiesChangedEventHandlerTest.kt
```diff
@@ -284,4 +284,19 @@ class ActivitiesChangedEventHandlerTest {
...
+  fun `released prisoner is not deallocated on 'END' when cannot determine release reason`() {
+    mock<Prisoner> {
+      on { status } doReturn "INACTIVE OUT"
+      on { confirmedReleaseDate } doReturn TimeSource.today()
+      on { lastMovementTypeCode } doReturn "XXX"
+    }.also { permanentlyReleasedPrisoner ->
+      whenever(prisonerSearchApiClient.findByPrisonerNumber("123456")) doReturn permanentlyReleasedPrisoner
+    }
+
+    handler.handle(activitiesChangedEvent("123456", Action.END, moorlandPrisonCode)).also { it.isSuccess() isBool false }
```
matt-crowson-moj, line 298: Because we have an outcome of false this should get picked up by the interesting event handler.

PR #617, opened 2023-10-04
SAA-1278 increasing logging for the activities change event as we think there may be timing issues when calling prisoner search.

matt-crowson-moj:
We (think) we are seeing activities changed events succeeding on retries.

This would suggest there may be a timing issue e.g. seeing the event before the data retrieved prisoner search is updated.


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/events/handlers/ActivitiesChangedEventHandlerTest.kt
```diff
@@ -284,4 +286,21 @@ class ActivitiesChangedEventHandlerTest {
 
     verify(prisonerAllocationHandler).deallocate(moorlandPrisonCode, "123456", DeallocationReason.RELEASED)
   }
+
+  @Test
+  fun `released prisoner is not deallocated on 'END' when cannot determine release reason`() {
```
matt-crowson-moj, line 291: This was missing test coverage.

PR #618, opened 2023-10-04
SAA-1280: Adding a landing 3 to Wing R in the location mapping file.

timharrison-moj:
This was not present in Whereabouts location mappings either, so is either a new set of cells or the wing was not previously using the Whereabouts unlock lists.


matt-crowson-moj (approved):
LGTM

PR #620, opened 2023-10-06
SAA-1282 spike work to replace the way in which events are raised and published.

matt-crowson-moj:
Spike replacing the way events are raised to use domain events and transactional event listener.  This ensures the events are raised after commits and not before.

Attendance expired is the one activities event that is not using the event listener.

Appointments still to be done.

**Note: Good integration test coverage is a must for these.**


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/DomainEntityListener.kt
```diff
@@ -0,0 +1,25 @@
+package uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.entity
+
+import jakarta.persistence.PostPersist
+import jakarta.persistence.PostUpdate
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+import org.springframework.stereotype.Component
+
+@Component
+class DomainEntityListener {
```
matt-crowson-moj, line 10: One generic listener would be required with this approach.

Don't worry I have been checking to see where the various commit points are versus events being raised.  All looks good.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Allocation.kt
```diff
@@ -21,7 +21,7 @@ import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.model.Deallocat
 
 @Entity
 @Table(name = "allocation")
-@EntityListeners(AllocationEntityListener::class, AuditableEntityListener::class)
+@EntityListeners(DomainEntityListener::class, AuditableEntityListener::class)
```
matt-crowson-moj, line 24: We use the generic `DomainEntityListener` here and extend the `DomainEventEntity`.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityScheduleService.kt
```diff
@@ -176,6 +178,7 @@ class ActivityScheduleService(
         }
         .singleOrNull()?.allocated(allocation)
 
+      allocationRepository.saveAndFlush(allocation)
```
matt-crowson-moj, line 181: The two key changes in the service are needing to call save and flush on each allocation for the application event listener to work.  Without these it won't work.  This is the trade off.  You have to go down to the actual entity you need the event to be raised for and call save on the repository.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/events/OutboundEventsService.kt
```diff
@@ -23,7 +28,25 @@ class OutboundEventsService(private val publisher: OutboundEventsPublisher, priv
     private val log: Logger = LoggerFactory.getLogger(this::class.java)
   }
 
+  @TransactionalEventListener(phase = TransactionPhase.AFTER_COMMIT)
+  fun <T> transactionalEventListener(event: DomainEntityCreatedEvent<T>) {
```
matt-crowson-moj, line 32: Adding these to the outbound event service keeps it nice and clean we don't need to inject it anywhere.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/events/OutboundEventsService.kt
```diff
@@ -18,17 +26,45 @@ import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.service.events.
...
-class OutboundEventsService(private val publisher: OutboundEventsPublisher, private val featureSwitches: FeatureSwitches) {
+class OutboundEventsService(
+  private val publisher: OutboundEventsPublisher,
+  private val featureSwitches: FeatureSwitches,
+) {
   companion object {
     private val log: Logger = LoggerFactory.getLogger(this::class.java)
   }
 
+  @TransactionalEventListener(phase = TransactionPhase.AFTER_COMMIT)
```
matt-crowson-moj, line 37: Note the use of the AFTER_COMMIT.  This is uber important.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityService.kt
```diff
@@ -195,7 +195,7 @@ class ActivityService(
         schedule.addSlots(request.slots!!)
         schedule.addInstances()
 
-        return transform(activityRepository.saveAndFlush(activity))
+        return transform(activityRepository.saveAndFlush(activity).also { activityScheduleRepository.saveAndFlush(schedule) })
       }
```
timharrison-moj, line 199: Good or bad thing that `save()` or `saveAndFlush()` can no longer implicit when raising events? Good - its clear to devs, bad - must remember to do it, or events will be missed.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/MigrateActivityService.kt
```diff
@@ -450,7 +454,7 @@ class MigrateActivityService(
       log.info("SUSPENDED prisoner ${request.prisonerNumber} being allocated to ${activity.activityId} ${activity.summary} as PENDING")
     }
 
-    val savedSchedule = activityScheduleRepository.saveAndFlush(schedule)
+    val savedSchedule = activityScheduleRepository.saveAndFlush(schedule).also { allocationRepository.saveAllAndFlush(it.allocations()) }
```
timharrison-moj, line 458: I'll make a note that we'll need to re-test migrations after this change. Probably will anyway.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ScheduledInstanceService.kt
```diff
@@ -66,83 +64,53 @@ class ScheduledInstanceService(
...
 
-      send(OutboundEvent.ACTIVITY_SCHEDULED_INSTANCE_AMENDED, uncancelledInstance.scheduledInstanceId)
+    val uncancelledAttendances = scheduledInstance.uncancelSessionAndAttendances()
 
-      uncancelledAttendances.forEach { attendance ->
-        send(OutboundEvent.PRISONER_ATTENDANCE_AMENDED, attendance.attendanceId)
-      }
-    }
+    repository.saveAndFlush(scheduledInstance)
+    attendanceRepository.saveAllAndFlush(uncancelledAttendances)
```
timharrison-moj, line 78: Do we get the right parent / child IDs? Again, part of testing.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/integration/ActivityScheduleIntegrationTest.kt
```diff
@@ -479,7 +479,7 @@ class ActivityScheduleIntegrationTest : IntegrationTestBase() {
   fun `allocation followed by a deallocation of the same prisoner`() {
     prisonApiMockServer.stubGetPrisonerDetails("G4793VF", fullInfo = false)
 
-    repository.findById(1).orElseThrow().also { assertThat(it.allocations()).isEmpty() }
+//    repository.findById(1).orElseThrow().also { assertThat(it.allocations()).isEmpty() }
```
timharrison-moj, line 483: Can be removed? Duplicate line.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/integration/InboundEventsIntegrationTest.kt
```diff
@@ -112,7 +114,7 @@ class InboundEventsIntegrationTest : IntegrationTestBase() {
     assertThatAllocationsAreActiveFor("A11111A")
 
     assertThat(eventReviewRepository.count()).isEqualTo(1L)
-    verifyNoInteractions(outboundEventsService)
+    verifyNoInteractions(eventsPublisher)
   }
```
timharrison-moj, line 118: Wonder if this is as clear - `eventsPublisher` v. `outboundEventsService` ? Seemed to make sense to me seeing an `InboundEventsService` and an `OutboundEventsService` - clear terms.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/AttendancesServiceTest.kt
```diff
@@ -173,7 +165,6 @@ class AttendancesServiceTest {
     )
 
     verify(attendanceRepository).saveAndFlush(attendance)
-    verify(outboundEventsService).send(OutboundEvent.PRISONER_ATTENDANCE_AMENDED, attendance.attendanceId)
     assertThat(attendance.status()).isEqualTo(AttendanceStatus.COMPLETED)
```
timharrison-moj, line 168: I think this was clear before ... you can see and test that the correct outbound events are generated as a result of actions taken.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ScheduledInstanceServiceTest.kt
```diff
@@ -210,9 +205,6 @@ class ScheduledInstanceServiceTest {
       }
 
       verify(telemetryClient).trackEvent(TelemetryEvent.RECORD_ATTENDANCE.value, instance.attendances.first().toTelemetryPropertiesMap())
-      verify(outboundEventsService)
-        .send(OutboundEvent.ACTIVITY_SCHEDULED_INSTANCE_AMENDED, instance.scheduledInstanceId)
-      verify(outboundEventsService).send(OutboundEvent.PRISONER_ATTENDANCE_AMENDED, instance.attendances.first().attendanceId)
     }
```
timharrison-moj, line 208: Similarly, not able to test the result of the actions taken in producing the required events. Less clear?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/events/handlers/OffenderReceivedEventHandlerTest.kt
```diff
@@ -181,7 +174,6 @@ class OffenderReceivedEventHandlerTest {
     assertThat(historicAttendance.hasReason(AttendanceReasonEnum.SUSPENDED)).isTrue
     assertThat(todaysFutureAttendance.hasReason(AttendanceReasonEnum.SUSPENDED)).isFalse
     assertThat(tomorrowsFutureAttendance.hasReason(AttendanceReasonEnum.SUSPENDED)).isFalse
-    verify(outboundEventsService, times(2)).send(eq(OutboundEvent.PRISONER_ATTENDANCE_AMENDED), any())
   }
```
timharrison-moj, line 177: Ditto - breaking the link between something happening and an outbound event being generated?

src/test/resources/application-test.yml
```diff
@@ -9,6 +9,12 @@ spring:
         jwt:
           public-key-location: classpath:local-public-key.pub
 
+  jpa:
+    show-sql: true
+    properties:
+      hibernate:
+        generate_statistics: true
```
timharrison-moj, line 16: Do we need this in test profile?  Good for local profile, not sure about test?

timharrison-moj (commented):
It's a really neat and clever approach, and has some nice aspects, but I do have some reservations. to discuss. 
* Its a big change at the very core of our service - so carries a testing burden and a bit of risk.
* Are we decoupling the outbound events from the things which cause them too much?
* Are we losing the ability to clearly test the cause and effect when generating events?
* Is it as clear from the code what is happening with outbound events?
* We lose the ability to rely on implicit entity saves.
* Would it be clearer still entirely remove entity listeners and raise events specifically in the code, in the places which cause them?

PR #621, opened 2023-10-09
SAA-1284: Record appointment attendance

iterative-it:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Appointment.kt
```diff
@@ -84,7 +85,7 @@ data class Appointment(
   var updatedTime: LocalDateTime? = null,
 
   var updatedBy: String? = null,
-) {
+) : AbstractAggregateRoot<Appointment>() {
```
iterative-it, line 88: Matt's change uses a root DomainEventEntity to register standard CRUD events so that they are handled after committing. The new attendance marked event published by this entity is domain specific and is therefore registered manually. This generic AbstractAggregateRoot can be replaced with the DomainEventEntity when that change is approved.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/AppointmentAttendeeEntityListener.kt
```diff
@@ -53,15 +53,12 @@ class AppointmentAttendeeEntityListener {
   @PostUpdate
   fun onUpdate(entity: AppointmentAttendee) {
     runCatching {
-      outboundEventsService.send(
-        when {
-          entity.isRemoved() -> OutboundEvent.APPOINTMENT_INSTANCE_CANCELLED
-          entity.isDeleted -> OutboundEvent.APPOINTMENT_INSTANCE_DELETED
-
-          else -> { OutboundEvent.APPOINTMENT_INSTANCE_UPDATED }
```
iterative-it, line 61: Instance updated event publishing is triggered by a change to the parent appointment. There is currently no attendee level change that should result in publishing an appointment instance updated event so this has been removed. If this wasn't removed, marking attendance would result in publishing an appointment instance updated event and an unnecessary attempt to one way sync an unchanged appointment instance to NOMIS

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Appointment.kt
```diff
@@ -202,6 +236,8 @@ data class Appointment(
       },
     )
 
+  internal fun publishedDomainEvents() = this.domainEvents()
```
iterative-it, line 239: Couldn't find another way to expose the domain events published by this aggregate root. If this is agreed as a reasonable approach, it could be added to the root DomainEventEntity

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/events/handlers/AppointmentTransactionalEventListener.kt
```diff
@@ -0,0 +1,35 @@
...
+class AppointmentTransactionalEventListener(
+  private val telemetryClient: TelemetryClient,
+) {
+  companion object {
+    private val log = LoggerFactory.getLogger(this::class.java)
+  }
+
+  @TransactionalEventListener(phase = TransactionPhase.AFTER_COMMIT)
+  fun handleAppointmentAttendanceMarkedEvent(event: AppointmentAttendanceMarkedEvent) {
+    log.info(
```
iterative-it, line 23: I feel there's enough complexity in this log message that it would be worth unit testing it but I don't know whether it's possible to mock the logger companion object or capture the log message another way.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/events/handlers/AppointmentTransactionalEventListener.kt
```diff
@@ -0,0 +1,35 @@
...
+import org.springframework.stereotype.Service
+import org.springframework.transaction.event.TransactionPhase
+import org.springframework.transaction.event.TransactionalEventListener
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.entity.AppointmentAttendanceMarkedEvent
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.telemetry.TelemetryEvent
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.telemetry.toTelemetryMetricsMap
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.telemetry.toTelemetryPropertiesMap
+
+@Service
+class AppointmentTransactionalEventListener(
```
matt-crowson-moj, line 14: I would possibly include telemetry in the naming of the service as this is what it is really doing, capturing telemetry.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/events/handlers/AppointmentTransactionalEventListener.kt
```diff
@@ -0,0 +1,35 @@
...
+  private val telemetryClient: TelemetryClient,
```
matt-crowson-moj, line 15: I do like being able to pull out telemetry with this pattern.  I really don't like the pattern we have prior to this.

matt-crowson-moj (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Appointment.kt
matt-crowson-moj: So just had a though on this one.  If we want to keep something like this there is an annotation called VisbleForTesting.  If might be worth add this.  You will need to add a dependency to the gradle build file for this hopefully won't conflict with anything. org.assertj:assertj-core

matt-crowson-moj (approved):
LGTM

PR #623, opened 2023-10-11
SAA-1294 fixes issue whereby checking we have any alloctions of interest before taking any further action in events, if no allocations then should ignore event.

matt-crowson-moj:
The activities changed event and prisoner release event both have the same issue of looking up prisoner details in external API's before checking we actually have any allocations of interest.

If there are no allocations of interest then it should not even bother looking up in the externals API's.

This fix resolves that issue.


timharrison-moj (approved):
LGTM.

PR #625, opened 2023-10-11
SAA-655: Edit pay where someone is allocated to it

smcveigh941:
When a pay band is changed, if someone is allocated using that pay band, the allocation should be changed to use the new pay band

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/AllocationsService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/AllocationsService.kt
@@ -1,6 +1,8 @@
 package uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.service
 
 import jakarta.persistence.EntityNotFoundException
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
 import org.springframework.stereotype.Service
 import org.springframework.transaction.annotation.Transactional
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.common.between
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityService.kt
```diff
@@ -577,6 +584,35 @@ class ActivityService(
     }
   }
 
+  private fun replacePayBandAllocationBeforePayRemoval(
```
matt-crowson-moj, line 587: 🤔 this possibly warrants and integration test?

There should also be an event raised on the back of changing allocations, could make sure that happens as well in an integration test.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityService.kt
```diff
@@ -577,6 +584,35 @@ class ActivityService(
...
+      oldPay.none { p -> p.incentiveNomisCode == it.incentiveNomisCode && p.payBand.prisonPayBandId == it.payBandId }
+    }
+
+    deltaPayBand?.let {
+      activity.schedules().forEach { schedule ->
+        val activeAllocations = schedule.allocations(excludeEnded = true)
+        val prisoners = prisonerSearchApiClient.findByPrisonerNumbers(activeAllocations.map { it.prisonerNumber }).block()!!
+
+        activeAllocations.forEach { allocation ->
+          val prisoner = prisoners.find { it.prisonerNumber == allocation.prisonerNumber }!!
```
matt-crowson-moj, line 604: Would suggest using `single { ... } `over `find { ... }`.  Single is saying there should only be one and it will fail if no match or more that one.

`val prisoner = prisoners.single { it.prisonerNumber == allocation.prisonerNumber }`

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityService.kt
```diff
@@ -577,6 +584,35 @@ class ActivityService(
...
+          if (newPay.none { pay -> pay.incentiveNomisCode == prisoner.currentIncentive?.level?.code && pay.payBandId == allocation.payBand.prisonPayBandId }) {
+            allocation.apply {
```
matt-crowson-moj, line 606: Will need to manually raise an event if the allocation is changed.  There is no longer a post update for allocations.

Probably easier to talk about this.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityService.kt
```diff
@@ -557,24 +571,61 @@ class ActivityService(
...
-          incentiveLevel = it.incentiveLevel!!,
-          payBand = prisonPayBands[it.payBandId]
-            ?: throw IllegalArgumentException("Pay band not found for prison '$prisonCode'"),
-          rate = it.rate,
-          pieceRate = it.pieceRate,
-          pieceRateItems = it.pieceRateItems,
-        )
+
+      return replacePayBandAllocationBeforePayRemoval(prisonCode, pay, activity, prisonPayBands).also {
+        activity.removePay()
```
matt-crowson-moj, line 581: Sorry. Need to have a chat with you about this removal line here. Want to check something with you.

matt-crowson-moj (approved):
LGTM

PR #632, opened 2023-10-17
SAA-1305: Include visits in movement list endpoints

iterative-it:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/client/prisonapi/overrides/LocationSummary.kt
```diff
@@ -0,0 +1,20 @@
+package uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.client.prisonapi.overrides
+
+import com.fasterxml.jackson.annotation.JsonProperty
+import io.swagger.v3.oas.annotations.media.Schema
+
+/**
+ * The GET /api/agencies/{agencyId}/eventLocationsBooked Prison API endpoint's documentation shows it returning the full
```
matt-crowson-moj, line 7: I wonder if this is down to the version of the swagger docs we are using?

It might be worth checking the version of the prison-api spec we have in the /openapi-specs folder?

iterative-it: I checked this and looked at the Swagger docs on all environments. Uses `Location` and suggests that it should return the full model. I believe it's slipped through because that endpoint was only used by DPS' UI. No type checking.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/client/prisonapi/api/PrisonApiClient.kt
```diff
@@ -393,6 +412,19 @@ class PrisonApiClient(private val prisonApiWebClient: WebClient) {
       .awaitBody()
   }
 
+  suspend fun getEventLocationsBookedAsync(prisonCode: String, date: LocalDate, timeSlot: TimeSlot?): List<LocationSummary> {
```
matt-crowson-moj, line 415: It is possible to add unit test coverage for this in that we can verify we are using the overridden return type.

Example [here](https://github.com/ministryofjustice/hmpps-activities-management-api/blob/main/src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/client/prisonapi/api/PrisonApiClientTest.kt#L562)

matt-crowson-moj (approved):
LGTM

matt-crowson-moj (approved):
LGTM

matt-crowson-moj (approved):
LGTM

PR #637, opened 2023-10-18
SAA-1221 reducing logging back down now we have identified the root cause affecting allocation expiration.

matt-crowson-moj:


timharrison-moj (approved):
LGTM

PR #640, opened 2023-10-20
SAA-1311 when we receive an END in an activities change event we should always end allocations if there are any.

matt-crowson-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/events/handlers/ActivitiesChangedEventHandler.kt
```diff
@@ -138,16 +128,25 @@ class ActivitiesChangedEventHandler(
...
-          log.warn("Unable to determine reason for prisoner ${event.prisonerNumber()}. $message")
-          throw IllegalStateException("Unable to determine release reason for prisoner '${event.prisonerNumber()}' for event '$event'")
+    prisonerSearchApiClient.findByPrisonerNumber(event.prisonerNumber())
+      .throwNullPointerIfNotFound { "Prisoner search lookup failed for prisoner ${event.prisonerNumber()}" }
+      .let { prisoner ->
+        when {
+          prisoner.isInactiveOut() -> RELEASED.also { log.info("Released inactive out prisoner ${event.prisonerNumber()} from prison ${event.prisonCode()}") }
+          prisoner.isAtDifferentLocationTo(event.prisonCode()) -> TEMPORARILY_RELEASED.also { log.info("Temporary release or transfer of prisoner ${event.prisonerNumber()} from prison ${event.prisonCode()}") }
+          // The user has explicitly chosen END, so we know they want to end but cannot easily determine a reason. If we don't it causes issues for the prisons.
+          else -> OTHER.also {
```
matt-crowson-moj, line 138: Fundamental change is here.  If someone has chosen to END then we should end regardless by not doing so we end up with messages stuck on the DLQ and prisoner allocations not being ended when they should be.

PR #641, opened 2023-10-20
SAA-1322 when we receive an END in an activities change event we should always end allocations if there are any.

matt-crowson-moj:
This is different to AC 6 on the parent ticket [here](https://dsdmoj.atlassian.net/browse/SAA-1311).

Not ending a prisoners allocations when a user has explicitly chosen to END is having ramifications in production.  Allocations are not always being ended even though prisoners have left the prison. Messages are getting stuck on DLQ's as well causing alerts.

The rules with regards to determining a release/deallocation reason have been relaxed also.  Our recording of a deallocation reason should not stop allocations from being ended.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/events/handlers/ActivitiesChangedEventHandler.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/events/handlers/ActivitiesChangedEventHandler.kt
@@ -4,14 +4,17 @@ import org.slf4j.LoggerFactory
 import org.springframework.stereotype.Component
 import org.springframework.transaction.annotation.Transactional
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.client.prisonersearchapi.api.PrisonerSearchApiApplicationClient
-import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.client.prisonersearchapi.extensions.isPermanentlyReleased
-import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.client.prisonersearchapi.extensions.isTemporarilyReleased
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.client.prisonersearchapi.extensions.isAtDifferentLocationTo
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.client.prisonersearchapi.extensions.isInactiveOut
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.client.prisonersearchapi.extensions.lastMovementType
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.client.prisonersearchapi.model.Prisoner
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.entity.Allocation
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.entity.Attendance
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.entity.AttendanceReasonEnum
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.entity.AttendanceStatus
-import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.entity.DeallocationReason
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.entity.DeallocationReason.OTHER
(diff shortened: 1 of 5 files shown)


iterative-it (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/events/handlers/ActivitiesChangedEventHandler.kt
```diff
@@ -138,16 +128,25 @@ class ActivitiesChangedEventHandler(
...
-      when {
-        prisoner.isTemporarilyReleased() -> DeallocationReason.TEMPORARILY_RELEASED
-        prisoner.isPermanentlyReleased() -> DeallocationReason.RELEASED
-        else -> {
-          val message =
-            "Prisoner prison code '${prisoner.prisonId}', status '${prisoner.status}', last movement type code '${prisoner.lastMovementType()}', confirmed release date '${prisoner.confirmedReleaseDate}'"
-          log.warn("Unable to determine reason for prisoner ${event.prisonerNumber()}. $message")
-          throw IllegalStateException("Unable to determine release reason for prisoner '${event.prisonerNumber()}' for event '$event'")
+    prisonerSearchApiClient.findByPrisonerNumber(event.prisonerNumber())
+      .throwNullPointerIfNotFound { "Prisoner search lookup failed for prisoner ${event.prisonerNumber()}" }
```
matt-crowson-moj, line 132: Technically this should NPE never happen but we have to handle it either way.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/events/handlers/ActivitiesChangedEventHandler.kt
```diff
@@ -138,16 +128,25 @@ class ActivitiesChangedEventHandler(
...
+            log.info("Prisoner prison code '${prisoner.prisonId}', prisoner '${prisoner.prisonerNumber}', status '${prisoner.status}', last movement type code '${prisoner.lastMovementType()}'")
+            log.info("Defaulting to OTHER for deallocation reason for prisoner ${event.prisonerNumber()} from prison ${event.prisonCode()}")
+          }
         }
       }
+
+  private fun Prisoner?.throwNullPointerIfNotFound(message: () -> String): Prisoner {
+    if (this == null) {
+      throw NullPointerException(message())
     }
```
timharrison-moj, line 148: If we throw this, the message will fail and retry later, so getting over any temporary issue with prisoner-search.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/integration/InboundEventsIntegrationTest.kt
```diff
@@ -558,8 +559,53 @@ class InboundEventsIntegrationTest : IntegrationTestBase() {
...
+  @Sql("classpath:test_data/seed-activities-changed-event-deletes-attendances.sql")
+  fun `allocations are ended and future attendances are removed for activities changed event set to END on temporary release`() {
+    prisonerSearchApiMockServer.stubSearchByPrisonerNumber(
+      PrisonerSearchPrisonerFixture.instance(
+        prisonerNumber = "A22222A",
+        inOutStatus = Prisoner.InOutStatus.IN,
+        status = "ACTIVE IN",
+        prisonId = moorlandPrisonCode,
+      ),
+    )
```
timharrison-moj, line 609: Do you have a test for when a prisoner is in location `TRN` - Transfer? So they are out but not yet booked into their intended new place?

timharrison-moj (approved):
LGTM. One question only.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/events/handlers/ActivitiesChangedEventHandler.kt
matt-crowson-moj: Indeed it should.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/integration/InboundEventsIntegrationTest.kt
matt-crowson-moj: We don't no.  This will fall into category other though which I think is OK.

I can add another test if you think it is worth it.

PR #644, opened 2023-10-24
SAA-1262: Added waiting list search endpoint

leejacobson-moj:
## Overview

Adds new endpoint, `POST/waiting-list-applications/{prisonCode}/search` to support the searching of waiting list applications

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/model/request/WaitingListSearchRequest.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/model/request/WaitingListSearchRequest.kt
@@ -1,0 +1,43 @@
+package uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.model.request
+
+import com.fasterxml.jackson.annotation.JsonFormat
+import io.swagger.v3.oas.annotations.media.Schema
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.entity.WaitingListStatus
+import java.time.LocalDate
+
+@Schema(
+  description = "Returns all waiting list applications that match the search criteria.",
+)
+data class WaitingListSearchRequest(
+  @Schema(
+    description = "Filter applications with a request date on or after provided date",
+    example = "2023-06-20",
+  )
+  @JsonFormat(pattern = "yyyy-MM-dd")
+  val applicationDateFrom: LocalDate? = null,
+
+  @Schema(
+    description = "Filter applications with a request date on or before provided date",
+    example = "2023-06-20",
+  )
+  @JsonFormat(pattern = "yyyy-MM-dd")
+  val applicationDateTo: LocalDate? = null,
+
+  @Schema(
+    description = "The activity to return waiting list applications for.",
+    example = "3",
+  )
+  val activityId: Long? = null,
+
+  @Schema(
+    description = "The prisoner or prisoners to retrieve waiting list applications for.",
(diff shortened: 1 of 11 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/model/request/WaitingListSearchRequest.kt
```diff
@@ -0,0 +1,43 @@
...
+    example = "2023-06-20",
+  )
+  @JsonFormat(pattern = "yyyy-MM-dd")
+  val applicationDateTo: LocalDate? = null,
+
+  @Schema(
+    description = "The activity to return waiting list applications for.",
+    example = "3",
+  )
+  val activityId: Long? = null,
```
matt-crowson-moj, line 30: I wonder if this should be using the schedule ID instead?  We use the schedule ID on creation.

leejacobson-moj: Allocations happen at the schedule level so creating an waiting list application would require the schedule id to allocate to.

When searching for waiting lists it's more likely that someone would want to get all applications for a given activity rather than the applications for a given schedule of that activity. This is also how it's currently being used on the waiting list dashboard where the user can select an activity to return applications for.

I thought about adding schedule id as an extra filter but I decided to leave it since I'm not convinced there's a use case.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/repository/WaitingListSearchSpecification.kt
```diff
@@ -0,0 +1,34 @@
+package uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.repository
+
+import org.springframework.data.jpa.domain.Specification
+import org.springframework.stereotype.Component
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.entity.Activity
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.entity.WaitingList
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.entity.WaitingListStatus
+import java.time.LocalDate
+
+@Component
```
matt-crowson-moj, line 10: This is me not knowing.  But why are specifications `@Components`?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/integration/WaitingListApplicationIntegrationTest.kt
```diff
@@ -90,4 +93,64 @@ class WaitingListApplicationIntegrationTest : IntegrationTestBase() {
       .expectHeader().contentType(MediaType.APPLICATION_JSON)
       .expectBody(WaitingListApplication::class.java)
       .returnResult().responseBody
+
+  @Sql("classpath:test_data/seed-activity-id-22.sql")
+  @Test
+  fun `search all waiting list applications`() {
```
matt-crowson-moj, line 99: Feels like this should be returning multiple for all waiting list applications?  The filtered test makes a bit more sense then perhaps?  Could use same SQL file but expect fewer results in the filtered one.

matt-crowson-moj (dismissed):
LGTM

matt-crowson-moj (approved):
LGTM

PR #645, opened 2023-10-24
SAA-1325 jobs should throw runtime exception on safe job failure to ensure alerts get propagated accordingly.

matt-crowson-moj:
Errors do not get propagated on safe job execution, as a result we are not seeing DPS alerts when a job fails to execute.

To date this has not been a problem in production.  It was noticed in dev due to external API services we use being disabled out of normal office hours e.g. prison search API is turned off late evening and comes back online at approx 6:20 UTC.

A runtime exception will now be thrown when a job fails to execute cleanly.  Note we still record the failure in the activities DB.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/job/SafeJobRunner.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/job/SafeJobRunner.kt
@@ -2,21 +2,24 @@ package uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.job
 
 import org.slf4j.LoggerFactory
 import org.springframework.stereotype.Component
+import org.springframework.transaction.annotation.Transactional
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.entity.Job
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.entity.JobType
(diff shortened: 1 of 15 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/job/SafeJobRunner.kt
```diff
@@ -2,21 +2,24 @@ package uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.job
...
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.entity.JobType
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.repository.JobRepository
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.service.TransactionHandler
 import java.time.LocalDateTime
 import java.util.concurrent.atomic.AtomicBoolean
 
 @Component
-class SafeJobRunner(private val jobRepository: JobRepository) {
+@Transactional
+class SafeJobRunner(private val jobRepository: JobRepository, private val transactionHandler: TransactionHandler) {
```
timharrison-moj, line 15: Not sure we want to make the job runner transactional - as we've already made the jobs themselves transactional?

matt-crowson-moj: Good spot.  Removed.

PR #651, opened 2023-10-26
SAA-1335 use prison code off event for release events only.

matt-crowson-moj:
Small change CoC for release events.  Would have done this on the original PR bit having issues with github.

We should record the prison id/code from the release event itself and not where the prisoner currently is as this may be different and if it was different it would not show up on the prison they were released from on the CoC UI.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/client/prisonapi/overrides/Movement.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/client/prisonapi/overrides/Movement.kt
@@ -5,7 +5,7 @@ import java.time.LocalDateTime
 import java.time.LocalTime
 
 /**
- * Overriding `toAgency` field. It can be null despite being non-nullable in the API specification.
+ * Overriding `fromAgency' and 'toAgency` fields. Both can be null despite being non-nullable in the API specification.
  */
 data class Movement(
   @get:JsonProperty("offenderNo", required = true) val offenderNo: String,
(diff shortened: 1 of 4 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/client/prisonapi/model/PrisonApiTypesTest.kt
```diff
@@ -48,4 +48,11 @@ class PrisonApiTypesTest {
 
     assertThat(field.returnType.isMarkedNullable).isTrue
   }
+
+  @Test
+  fun `toAgency field on overridden Movement DTO type should be nullable`() {
```
matt-crowson-moj, line 53: Unrelated to this ticket.  I noticed I had missed this off from a previous ticket.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/client/prisonapi/overrides/Movement.kt
```diff
@@ -5,7 +5,7 @@ import java.time.LocalDateTime
 import java.time.LocalTime
 
 /**
- * Overriding `toAgency` field. It can be null despite being non-nullable in the API specification.
+ * Overriding `fromAgency' and 'toAgency` fields. Both can be null despite being non-nullable in the API specification.
```
matt-crowson-moj, line 8: Updating comment so it is clear what and why we are overriding.

iterative-it (approved):
LGTM - good spot

PR #652, opened 2023-10-26
SAA-1300 Absent Reason added to attendance

nagendramoj:
I can provide some context regarding the necessary modification. The user interface, specifically the attendance-details.njk file, relies on the `otherAbsenceReason` property from the `attendance` object. However, during the transformation process from Model to Data classes, this particular value is not being assigned.

Resolution:
I have addressed this issue by initializing the `otherAbsenceReason` field within the `attendance` data class.

Verification:
This change has been thoroughly verified, and existing test cases cover this scenario.


timharrison-moj (approved):
LGTM

PR #653, opened 2023-10-26
SAA-1255: Set attendee properties when adding and removing

iterative-it:
**MEDIUM RISK**

This pr moves from hard deleting attendees when they are removed to soft deleting them with a reason. This matches the business logic when prisoner attendee records are removed following being released, transferred or when the user elects to cancel for any other transfer reason.

The fact that the soft deletion process has been tested separately via the above workflows is why this is medium rather than high risk.

It has been tested locally and will be tested again, including testing that sync is working when it goes to dev.


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/AppointmentTest.kt
```diff
@@ -719,6 +721,157 @@ class AppointmentTest {
...
+      bookingId isEqualTo 123
+      addedTime isEqualTo addedTime
+      addedBy isEqualTo "ADDED_BY_USER"
+      attended isEqualTo null
+      attendanceRecordedTime isEqualTo null
+      attendanceRecordedBy isEqualTo null
+      removedTime isEqualTo null
+      removalReason isEqualTo null
+      removedBy isEqualTo null
+      isRemoved() isEqualTo false
```
matt-crowson-moj, line 742: It doesn't really matter as the result is the same.  There is also an `isBool` function.

matt-crowson-moj (approved):
LGTM

PR #655, opened 2023-10-26
SAA-1337 pulled job times out into reusable env vars so can have differenent job times for each environment where needed.

matt-crowson-moj:
**LOW (MEDIUM?) RISK**

Pulled out the job times into env vars in the helm config so we can alter the times at which jobs run for each environment where needed.

The driver for this is due to the fact some core services we use are turned off overnight.  The also means we can remove any jobs which have currently been manually applied to dev as a temporary workaround.


timharrison-moj (approved):
LGTM

PR #657, opened 2023-10-27
SAA-1341 external generated API's should have their own generation root folder to avoid files being overwritten by other API's generated.

matt-crowson-moj:
**LOW RISK**

We have been sharing the same root folder for the external API we generate which can cause issues with files being potentially overwritten.

This change ensures each external API we generated has it's own root folder to avoid any clashes.


iterative-it (approved):
LGTM

PR #660, opened 2023-10-30
SAA-1166: Added activity organiser

leejacobson-moj:
## Overview

Adds support for setting tier and organiser of activity.

Changes:
- Adds organiser types
- Adds support for adding and editing activity organiser
- Updates activity tiers


src/main/resources/migrations/common/V2023.10.24__activity_tiers_and_organiser.sql
```diff
@@ -0,0 +1,22 @@
...
+
+INSERT INTO activity_organiser (activity_organiser_id, code, description) VALUES
+    (1, 'PRISON_STAFF', 'Prison staff'),
+    (2, 'PRISONER', 'A prisoner or group of prisoners'),
+    (3, 'EXTERNAL_PROVIDER', 'An external provider'),
+    (4, 'OTHER', 'Someone else');
+
+ALTER TABLE activity ADD COLUMN activity_organiser_id bigint REFERENCES activity_organiser (activity_organiser_id);
+
+UPDATE activity_tier SET code = 'TIER_1' WHERE activity_tier_id = 1;
```
leejacobson-moj, line 18: Updated activity tier, "Tier1" -> "TIER_1" for consistency with other codes such as activity category and attendance reason.

src/main/resources/migrations/common/V2023.10.30__activity_tiers_and_organiser.sql
```diff
@@ -0,0 +1,22 @@
...
+
+INSERT INTO activity_organiser (activity_organiser_id, code, description) VALUES
+    (1, 'PRISON_STAFF', 'Prison staff'),
+    (2, 'PRISONER', 'A prisoner or group of prisoners'),
+    (3, 'EXTERNAL_PROVIDER', 'An external provider'),
+    (4, 'OTHER', 'Someone else');
+
+ALTER TABLE activity ADD COLUMN activity_organiser_id bigint REFERENCES activity_organiser (activity_organiser_id);
+
+UPDATE activity_tier SET code = 'TIER_1', description = 'Tier 1' WHERE activity_tier_id = 1;
```
leejacobson-moj, line 18: Updated activity tier, "Tier1" -> "TIER_1" for consistency with other codes such as activity category and attendance reason.

src/main/resources/migrations/common/V2023.10.30__activity_tiers_and_organiser.sql
```diff
@@ -0,0 +1,22 @@
...
+
+INSERT INTO activity_tier (activity_tier_id, code, description) VALUES
+    (2, 'TIER_2', 'Tier 2'),
+    (3, 'FOUNDATION', 'Routine activities also called "Foundation"');
```
timharrison-moj, line 22: Should we update existing data to the new tiers? Or are all tiers set to 1 at present, and we can leave the staff at Risley to set them up themselves?

timharrison-moj (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityService.kt
```diff
@@ -368,6 +373,18 @@ class ActivityService(
   ) {
     request.tierId?.apply {
       activity.activityTier = activityTierRepository.findOrThrowIllegalArgument(this)
+      if (activity.activityTier?.code !== "TIER_2") {
```
matt-crowson-moj, line 376: Should be using `'!='` and not `'!=='`

src/main/resources/migrations/common/V2023.10.30__activity_tiers_and_organiser.sql
leejacobson-moj: You might want to confirm this on prod, but I believe the only activity which should have tiers currently are migrated activities, with those created through our service having a null tier.

I believe the only tier currently in prod is tier 1, so those activities with tiers currently don't need updating – they'll still be on "Tier 1".

Activities that have been created through our service and don't a tier will show as "Tier: Not set". The user can go in and add a tier from view activity page if this is the case.

If you'd prefer I can add SQL to update any null tiers in the DB, but I think at a minimum we'd need to ensure any "not in work" activities are defaulted to, "Tier 3". Then I assume the rest could be safely assigned "Tier 1"?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityServiceTest.kt
```diff
@@ -1619,6 +1632,57 @@ class ActivityServiceTest {
       .hasMessage("Activity location can only be maximum one of offWing, onWing, inCell, or a specified location")
   }
 
+  @Test
+  fun `updateActivity - add organiser`() {
+    val activity = activityEntity()
```
matt-crowson-moj, line 1637: Same here perhaps, something to assert the organiser is null beforehand perhaps?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityService.kt
```diff
@@ -368,6 +373,18 @@ class ActivityService(
...
+        activity.organiser = null
+      }
+    }
+  }
+
+  private fun applyOrganiserUpdate(
+    request: ActivityUpdateRequest,
+    activity: Activity,
+  ) {
+    request.organiserId?.apply {
```
matt-crowson-moj, line 386: Given the two tier restrictions, should the it be checked here as well perhaps?  i.e. if we get to this point and is not tier two then runtime?

The front end might not allow this but the API still will allow it ...

src/main/resources/migrations/common/V2023.11.03__event_tiers_and_organisers.sql
```diff
@@ -0,0 +1,28 @@
...
+
+INSERT INTO event_organiser (event_organiser_id, code, description) VALUES
+    (1, 'PRISON_STAFF', 'Prison staff'),
+    (2, 'PRISONER', 'A prisoner or group of prisoners'),
+    (3, 'EXTERNAL_PROVIDER', 'An external provider'),
+    (4, 'OTHER', 'Someone else');
+
+ALTER TABLE activity ADD COLUMN activity_organiser_id bigint REFERENCES event_organiser (event_organiser_id);
+
+ALTER TABLE activity_tier RENAME TO event_tier;
```
leejacobson-moj, line 18: It required some additional refactoring, but I think removing the `activity` namespace in favour of the more generic `event` namespace helps to imply these tables belong to neither activities or appointments and could be used for other types of events.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityService.kt
leejacobson-moj: Thanks Matt. I've added this constraint to the Activity entity.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityService.kt
```diff
@@ -203,11 +204,30 @@ class ActivityService(
         schedule.addSlots(request.slots!!)
         schedule.addInstances()
 
-        return transform(activityRepository.saveAndFlush(activity))
+        val activityModel = transform(activityRepository.saveAndFlush(activity))
+
+        publishCreateTelemetryEvent(activityModel)
```
leejacobson-moj, line 209: Probably doesn't matter much, but I've moved this to after the `.saveAndFlush()` call so we can be certain that the activity was created. Previously it could have failed after the telemetry event was sent because `request.slots` was null, for example.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/AllocationsService.kt
```diff
@@ -112,7 +112,7 @@ class AllocationsService(
     updatedBy: String,
   ) {
     request.endDate?.apply {
-      require(allocation.endDate !== null || request.reasonCode !== null) {
+      require(allocation.endDate != null || request.reasonCode != null) {
```
leejacobson-moj, line 115: After @matt-crowson-moj 's comment about `!==` I did a search and found this. Fixed and added a test.

src/main/resources/migrations/common/V2023.11.03__event_tiers_and_organisers.sql
```diff
@@ -0,0 +1,34 @@
...
+CREATE UNIQUE INDEX idx_event_tier_code ON event_tier (code);
+
+UPDATE event_tier SET code = 'TIER_1', description = 'Tier 1' WHERE event_tier_id = 1;
+
+INSERT INTO event_tier (event_tier_id, code, description) VALUES
+    (2, 'TIER_2', 'Tier 2'),
+    (3, 'FOUNDATION', 'Routine activities also called "Foundation"');
+
+-- Update existing "not in work" activities
+UPDATE activity SET activity_tier_id = 3 WHERE activity_category_id = 8;
```
leejacobson-moj, line 34: Decided to add this since the UI prevents tier updates to "non in work" activities effectively locking it to "not set" until the category is changed.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/EventTier.kt
```diff
@@ -0,0 +1,24 @@
...
+  val eventTierId: Long = 0,
+
+  @Column(nullable = false)
+  val code: String,
+
+  @Column(nullable = false)
+  val description: String,
+)
+
+fun EventTier.isTierTwo() = this.code == "TIER_2"
```
matt-crowson-moj, line 24: Minor point but this should really be in the class and not an extension function.

```
@Entity
@Table(name = "event_tier")
data class EventTier(...) {
  fun isTierTwo() = code == "TIER_2"
}

```

matt-crowson-moj (approved):
LGTM

PR #662, opened 2023-10-30
SAA-1345 job to automatically retry messages on DLQ's and reduce the level of manual intervention.

matt-crowson-moj:
At the moment when messages are put onto the DLQ's it requires manual intervention to retry them.  This change automates the retry.

Note: It will still raise an alert if the retry fails.

### DIFF
--- a/helm_deploy/hmpps-activities-management-api/templates/queue-housekeeping-cronjob.yaml
+++ b/helm_deploy/hmpps-activities-management-api/templates/queue-housekeeping-cronjob.yaml
@@ -1,0 +1,24 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: hmpps-activities-management-api-queue-housekeeping
+spec:
+  schedule: "*/10 * * * *"
+  concurrencyPolicy: Forbid
+  failedJobsHistoryLimit: 5
+  startingDeadlineSeconds: 600
+  successfulJobsHistoryLimit: 5
+  jobTemplate:
+    spec:
+      # Tidy up all jobs after 4 days
+      ttlSecondsAfterFinished: 345600
+      template:
+        spec:
+          containers:
+            - name: housekeeping
+              image: ghcr.io/ministryofjustice/hmpps-devops-tools
+              args:
+                - /bin/sh
+                - -c
+                - curl --retry 2 -XPUT http://hmpps-activities-management-api/queue-admin/retry-all-dlqs
+          restartPolicy: Never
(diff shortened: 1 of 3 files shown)


helm_deploy/hmpps-activities-management-api/templates/queue-housekeeping-cronjob.yaml
```diff
@@ -0,0 +1,24 @@
...
+      template:
+        spec:
+          containers:
+            - name: housekeeping
+              image: ghcr.io/ministryofjustice/hmpps-devops-tools
+              args:
+                - /bin/sh
+                - -c
+                - curl --retry 2 -XPUT http://hmpps-activities-management-api/queue-admin/retry-all-dlqs
+          restartPolicy: Never
```
timharrison-moj, line 24: How does the ROLE_QUEUE_ADMIN work here? I would expect the curl to fail with a 403?

mikehalmamoj: That endpoint is deliberately unprotected as it is intended for this scenario - to be called from a Kube cronjob. See the (queue library)[https://github.com/ministryofjustice/hmpps-spring-boot-sqs/blob/main/hmpps-sqs-spring-boot-autoconfigure/src/main/kotlin/uk/gov/justice/hmpps/sqs/HmppsQueueResource.kt#L24] for more details.

However, we do usually block the endpoint from outside the ingress so that it's not wide open - like [this](https://github.com/ministryofjustice/hmpps-prisoner-to-nomis-update/blob/main/helm_deploy/hmpps-prisoner-to-nomis-update/values.yaml#L20) - meaning that it will only work for the above cronjob.

mikehalmamoj: But yes there is also a need to stop Spring Security from protecting this endpoint - it will need adding to the resource server, like [this](https://github.com/ministryofjustice/hmpps-prisoner-to-nomis-update/blob/main/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/config/ResourceServerConfiguration.kt#L25).

timharrison-moj (approved):
LGTM

helm_deploy/hmpps-activities-management-api/templates/queue-housekeeping-cronjob.yaml
matt-crowson-moj: 🤦

PR #665, opened 2023-10-31
SAA-1334 this is changing the interesting event handler back to prison-api due a race condition with prisoner-search.

matt-crowson-moj:
**LOW RISK**

This changes the interesting event back to using prison-api instead of prisoner-search due a race condition happening with incentive events being processed and the information related to those events not being quite ready in prisoner-search due to timing.


timharrison-moj (approved):
LGTM

PR #668, opened 2023-11-06
SAA-1353 work in progress for introducing user location description on scheduled events.

matt-crowson-moj:
**MEDIUM TO HIGH RISK**

The purpose of this change is to provide the user description as well as t
he standard description for an event location if there is one in the scheduled events calls for activities, appointments, visits and adjudications.

Given this the UI can then make the call/decision which description to use e.g. on the unlock list favour the user description over the standard description.

This is a non-breaking change in that the API can be merged/deployed without any further changes to the UI.  However ideally the UI should be updated as well to take advantage of this API change.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/util/TransformFunctions.kt
```diff
@@ -142,6 +145,7 @@ fun transformAppointmentInstanceToScheduledEvents(
     bookingId = it.bookingId,
     internalLocationId = it.internalLocationId,
     internalLocationCode = locationsForAppointmentsMap[it.internalLocationId]?.internalLocationCode ?: "No information available",
+    internalLocationUserDescription = locationsForAppointmentsMap[it.internalLocationId]?.userDescription ?: "No information available",
```
matt-crowson-moj, line 148: Note the appointment was already setting the (original) description to the user description.  I have left the original field the same as before i.e. the user description also so as not to break anything in the UI.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/util/PrisonApiTransformations.kt
```diff
@@ -60,7 +66,8 @@ private fun List<PrisonApiPrisonerSchedule>.prisonerScheduleToScheduledEvents(
     scheduledInstanceId = null,
     internalLocationId = it.locationId,
     internalLocationCode = null,
-    internalLocationDescription = it.eventLocation ?: "External",
+    internalLocationUserDescription = mayBeInternalLocation?.userDescription ?: "External",
+    internalLocationDescription = mayBeInternalLocation?.description ?: (it.eventLocation ?: "External"),
```
timharrison-moj, line 70: Should the `internalUserDescription` be trying userDescription, Description, and locationCode in order - depending on what is null, rather than falling back on `External`?

timharrison-moj (dismissed):
LGTM.
I'd say this was quite a high-risk for deployment - so needs a good shake-out on DEV before progressing? Good to merge after 12.30pm

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/util/PrisonApiTransformations.kt
matt-crowson-moj: I have updated this @timharrison-moj .  User description will be null if not present.

timharrison-moj (approved):
LGTM

PR #671, opened 2023-11-08
SAA-1354: Add endpoint to view list of scheduled attendees for an instance

smcveigh941:
* This endpoint is to be used to view a list of prisoners due to attend a scheduled instance - either in the past or in the future
* This makes use of the view `v_prisoner_scheduled_activities`
* We will use this endpoint now to fetch the list of attendees when viewing the attendance list, and when we add exclusions to the view, excluded prisoners will not show up on the list of future attendances

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/resource/ScheduledInstanceController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/resource/ScheduledInstanceController.kt
@@ -108,7 +108,7 @@ class ScheduledInstanceController(
     value = [
       ApiResponse(
         responseCode = "200",
-        description = "Scheduled instance found",
+        description = "Scheduled attendees found",
         content = [
           Content(
             mediaType = "application/json",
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/resource/ScheduledInstanceController.kt
```diff
@@ -97,6 +98,62 @@ class ScheduledInstanceController(
...
+  @ResponseBody
+  @Operation(
+    summary = "Get a list of scheduled attendees for a scheduled instance",
+    description = "Returns a list of prisoners who are scheduled to attend a given scheduled instance.",
+  )
+  @ApiResponses(
+    value = [
+      ApiResponse(
+        responseCode = "200",
+        description = "Scheduled instance found",
```
matt-crowson-moj, line 111: Scheduled attendees found?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ScheduledInstanceService.kt
```diff
@@ -66,6 +70,13 @@ class ScheduledInstanceService(
     }
   }
 
+  fun getAttendeesForScheduledInstance(id: Long): List<ScheduledAttendee> {
+    val activityScheduleInstance = repository.findById(id)
```
matt-crowson-moj, line 74: There is an extension function for findById.  So you could do this instead

`val activityScheduleInstance = repository.findOrThrowNotFound(id)`

Meh.

matt-crowson-moj (dismissed):
LGTM - a couple of minors.

matt-crowson-moj (approved):
LGTM

PR #673, opened 2023-11-10
SAA-1258: Publish appointment sync events

leejacobson-moj:
## Overview

Removes appointment entity listeners and adds appointment create and update sync events to service methods.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Appointment.kt
```diff
@@ -113,36 +111,33 @@ data class Appointment(
...
+      addedTime = addedTime,
+      addedBy = addedBy,
+    )
+    return addAttendee(attendee)
   }
 
-  fun removeAttendee(prisonerNumber: String, removedTime: LocalDateTime = LocalDateTime.now(), removalReason: AppointmentAttendeeRemovalReason, removedBy: String?) {
-    findAttendee(prisonerNumber).forEach {
+  fun removeAttendee(prisonerNumber: String, removedTime: LocalDateTime = LocalDateTime.now(), removalReason: AppointmentAttendeeRemovalReason, removedBy: String?) =
+    findAttendee(prisonerNumber).map {
```
matt-crowson-moj, line 138: I might be missing something here (and I appreciate this was already here but) why would `findAttendee` return a list and not a single attendee.  I get they could have been added and removed before but still just feels a bit odd.

That aside the map { ... } function is not really what you want here.  The onEach { ... } is what you are after e.g.

```
  fun removeAttendee(prisonerNumber: String, removedTime: LocalDateTime = LocalDateTime.now(), removalReason: AppointmentAttendeeRemovalReason, removedBy: String?) =
    findAttendee(prisonerNumber).onEach { it.remove(removedTime, removalReason, removedBy) }

```

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/AppointmentCreateDomainService.kt
```diff
@@ -119,9 +122,13 @@ class AppointmentCreateDomainService(
               }
             },
           )
+        }.attendees().forEach {
+          if (!appointmentSeries.isMigrated) {
```
matt-crowson-moj, line 126: Doesn't feel quite right having the `isMigrated` check here.

This will check it for each attendee when it only needs to be checked the once I think right? I could be wrong.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/AppointmentSetService.kt
```diff
@@ -84,6 +87,9 @@ class AppointmentSetService(
       )
     }.also {
       // TODO: publish appointment instance created messages post transaction
+      it.appointmentSeries().flatMap { appointmentSeries -> appointmentSeries.appointments().flatMap { appointment -> appointment.attendees() } }.forEach { attendee ->
```
matt-crowson-moj, line 90: Might be nice to include a more readable (extension) function for line 90.  Meh.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/AppointmentTest.kt
```diff
@@ -744,6 +744,28 @@ class AppointmentTest {
...
+      bookingId isEqualTo 123
+      addedTime isEqualTo addedTime
+      addedBy isEqualTo "ADDED_BY_USER"
+      attended isEqualTo null
+      attendanceRecordedTime isEqualTo null
+      attendanceRecordedBy isEqualTo null
+      removedTime isEqualTo null
+      removalReason isEqualTo null
+      removedBy isEqualTo null
+      isRemoved() isEqualTo false
```
matt-crowson-moj, line 764: There is an isBool extension function as well.  Behaviour is ultimately the same though.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Appointment.kt
leejacobson-moj: That confused me too. I think the method behaves correctly but is named wrong. I think a better name is `findAttendeeRecords` since a prisoner number doesn't necessarily map to a single attendee record. I've updated that and updated the `map` to `onEach`.

matt-crowson-moj (approved):
LGTM

PR #674, opened 2023-11-10
SAA-1357 add Sentry integration to capture and raise alerts for all job failures.

matt-crowson-moj:
This is the first step towards raising visibility of job failures to the team (mainly the dev team initially).

The intention is to raise both Slack and email alerts.  This requires some initial configuration vi [Sentry.IO](https://sentry.io)

I have set up a project and team in Sentry. Team members will need to log into and join to receive email alerts. You can log in with you Github account.

I have added the relevant secret to the k8s dev namespace.  Once that is proven I will adde to the higher environments.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/MonitoringService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/MonitoringService.kt
@@ -1,0 +1,28 @@
+package uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.service
+
+import io.sentry.Sentry
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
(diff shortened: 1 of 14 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/MonitoringService.kt
```diff
@@ -0,0 +1,28 @@
...
+ */
+@Service
+class MonitoringService {
+
+  companion object {
+    private val log: Logger = LoggerFactory.getLogger(this::class.java)
+  }
+
+  init {
+    log.info("Monitoring service is ${if (Sentry.isEnabled()) "enabled" else "disabled"}")
```
matt-crowson-moj, line 21: Will check this when then service is deployed after merging into main, if this shows up as disabled then I have not configured the Sentry secret correctly.

PR #680, opened 2023-11-13
SAA-1357 add missing call to monitoring service where a subsequent job can fail (not run) due to the first not working.

matt-crowson-moj:
Adding (missed) call to monitoring service in the safe job runner.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/job/SafeJobRunner.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/job/SafeJobRunner.kt
@@ -39,7 +39,9 @@ class SafeJobRunner(private val jobRepository: JobRepository, private val monito
       } else {
         log.warn("Ignoring job ${job.jobType} due to failure in a dependent job.")
 
-        jobRepository.saveAndFlush(Job.failed(job.jobType, LocalDateTime.now()))
+        jobRepository.saveAndFlush(Job.failed(job.jobType, LocalDateTime.now())).also { failedJob ->
+          monitoringService.capture("Dependant job '${failedJob.jobType}' for job id '${failedJob.jobId}' failed")
+        }
       }
     }
   }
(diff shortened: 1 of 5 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ManageAttendancesService.kt
```diff
@@ -83,6 +83,8 @@ class ManageAttendancesService(
                 log.info("Sending sync event for attendance ID ${saved.attendanceId} ${saved.prisonerNumber} ${saved.scheduledInstance.activitySchedule.description}")
                 outboundEventsService.send(OutboundEvent.PRISONER_ATTENDANCE_CREATED, saved.attendanceId)
               }
+            }.onFailure {
+              log.error("Error occurred saving attendances", it)
```
matt-crowson-moj, line 87: I noticed we are potentially swallowing errors here without even knowing about it.

Given we effectively have a try catch here I did wonder if we should also be raising it via the monitoring service or do away with the try catch and let it propagate up the stack?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/job/SafeJobRunner.kt
```diff
@@ -39,7 +39,9 @@ class SafeJobRunner(private val jobRepository: JobRepository, private val monito
       } else {
         log.warn("Ignoring job ${job.jobType} due to failure in a dependent job.")
 
-        jobRepository.saveAndFlush(Job.failed(job.jobType, LocalDateTime.now()))
+        jobRepository.saveAndFlush(Job.failed(job.jobType, LocalDateTime.now())).also { failedJob ->
+          monitoringService.capture("Job '${failedJob.jobType}' for job id '${failedJob.jobId}' failed")
```
matt-crowson-moj, line 43: Missed off previous PR.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/job/SafeJobRunner.kt
```diff
@@ -39,7 +39,9 @@ class SafeJobRunner(private val jobRepository: JobRepository, private val monito
       } else {
         log.warn("Ignoring job ${job.jobType} due to failure in a dependent job.")
 
-        jobRepository.saveAndFlush(Job.failed(job.jobType, LocalDateTime.now()))
+        jobRepository.saveAndFlush(Job.failed(job.jobType, LocalDateTime.now())).also { failedJob ->
+          monitoringService.capture("Dependant job '${failedJob.jobType}' for job id '${failedJob.jobId}' failed")
```
matt-crowson-moj, line 43: Should capture here as well.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ManageAttendancesService.kt
timharrison-moj: Would it continue after the failure? i.e. just affect one attendance row, or one activity? If they're all failing, not sure we want to be spammed with 1300 alerts?  The 5-mile-island problem!

timharrison-moj (approved):
LGTM

PR #681, opened 2023-11-14
SAA-1357 increase logging to help with debugging monitoring service behaviour.

matt-crowson-moj:
Small change to add extra logging in monitoring service to confirm it is at least being called on a job failure.

### DIFF
--- a/helm_deploy/hmpps-activities-management-api/values.yaml
+++ b/helm_deploy/hmpps-activities-management-api/values.yaml
@@ -100,7 +100,7 @@ generic-service:
       DPR_USER: "DPR_USER"
       DPR_PASSWORD: "DPR_PASSWORD"
     sentry:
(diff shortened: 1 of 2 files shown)


helm_deploy/hmpps-activities-management-api/values.yaml
```diff
@@ -100,7 +100,7 @@ generic-service:
       DPR_USER: "DPR_USER"
       DPR_PASSWORD: "DPR_PASSWORD"
     sentry:
-      sentry_dsn: "SENTRY_DSN"
+      SENTRY_DSN: "SENTRY_DSN"
```
matt-crowson-moj, line 103: This shouldn't make any difference but changing to upper just in case.

timharrison-moj (approved):
LGTM

PR #686, opened 2023-11-14
SAA-862: Fix offender merge where PrisonerSearchAPI client being used outside of request scope

smcveigh941:
The `PrisonerSearchApiClient` is scoped to `@RequestScope`, this means that it throws a scoping error when used inside the scope of a Sqs Event, and therefore outside of the scope of a HTTP request. The `PrisonerSearchApiApplicationClient` needs to be used instead.


matt-crowson-moj (approved):
LGTM

PR #692, opened 2023-11-17
SAA-1354: Remove existing exclusions which are not present in the update request

smcveigh941:
- The update endpoint for exclusions should remove any existing exclusions which are not present in the request
- It also makes sense for the exclusions to be a Set rather than a List, since they are unique within an allocation

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/ActivityScheduleSlot.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/ActivityScheduleSlot.kt
@@ -52,7 +52,7 @@ data class ActivityScheduleSlot(
 
   @OneToMany(mappedBy = "activityScheduleSlot", fetch = FetchType.LAZY, cascade = [CascadeType.ALL], orphanRemoval = true)
   @Fetch(FetchMode.SUBSELECT)
-  val exclusions: MutableSet<Exclusion> = mutableSetOf(),
+  private val exclusions: MutableSet<Exclusion> = mutableSetOf(),
 ) {
   init {
     failIfNoDaysSelectedForSlot()
@@ -102,6 +102,9 @@ data class ActivityScheduleSlot(
       sundayFlag = daysOfWeek.contains(DayOfWeek.SUNDAY),
     )
   }
+
+  fun exclusions() = exclusions.toList()
+  fun addExclusion(exclusion: Exclusion) = exclusions.add(exclusion)
 
   fun update(daysOfWeek: Set<DayOfWeek>): AllocationIds {
     require(daysOfWeek.isNotEmpty()) { "A slot must run on at least one day." }
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Allocation.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Allocation.kt
@@ -84,7 +84,7 @@ data class Allocation(
 
   @OneToMany(mappedBy = "allocation", fetch = FetchType.LAZY, cascade = [CascadeType.ALL], orphanRemoval = true)
   @Fetch(FetchMode.SUBSELECT)
-  val exclusions: MutableSet<Exclusion> = mutableSetOf()
+  private val exclusions: MutableSet<Exclusion> = mutableSetOf()
 
   var deallocatedTime: LocalDateTime? = null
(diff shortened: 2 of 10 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/AllocationsService.kt
```diff
@@ -88,6 +89,20 @@ class AllocationsService(
...
+    request.exclusions?.apply {
+      val newExclusions = this.map { ex -> ex.weekNumber to ex.timeSlot }
+
+      val exclusionsToRemove = mutableListOf<Exclusion>()
+      allocation.exclusions.onEach {
+        val oldExclusion = it.getWeekNumber() to it.getTimeSlot().toString()
+        if (oldExclusion !in newExclusions) {
+          exclusionsToRemove.add(it)
+        }
+      }
```
timharrison-moj, line 102: Might be a neater way here?
`val exclusionsToRemove = allocation.exclusions.mapNotNull { exclusion ->
       val oldExclusion = it.getWeekNumber() to it.getTimeSlot().toString()
        if (oldExclusion !in newExclusions) {
           return oldExclusion
        }
}`

Or similar.

timharrison-moj (dismissed):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Exclusion.kt
```diff
@@ -39,6 +39,8 @@ data class Exclusion(
 
   var sundayFlag: Boolean = false,
 ) {
+  override fun hashCode(): Int = exclusionId.hashCode()
```
matt-crowson-moj, line 42: Just curious why you have added this?  Not aware we have done this anywhere else.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/AllocationsService.kt
```diff
@@ -88,6 +89,20 @@ class AllocationsService(
...
+
+      allocation.exclusions.removeAll(exclusionsToRemove.toSet())
```
matt-crowson-moj, line 104: Just me but not a fan of direct access to Collections on entities.  You basically bypass any business rules/logic.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Allocation.kt
```diff
@@ -84,7 +84,7 @@ data class Allocation(
 
   @OneToMany(mappedBy = "allocation", fetch = FetchType.LAZY, cascade = [CascadeType.ALL], orphanRemoval = true)
   @Fetch(FetchMode.SUBSELECT)
-  val exclusions: MutableList<Exclusion> = mutableListOf()
+  val exclusions: MutableSet<Exclusion> = mutableSetOf()
```
matt-crowson-moj, line 87: As per comment on the service that changes it I'd prefer this to be private and access via a function and not direct.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/ActivityScheduleSlot.kt
```diff
@@ -52,7 +52,7 @@ data class ActivityScheduleSlot(
 
   @OneToMany(mappedBy = "activityScheduleSlot", fetch = FetchType.LAZY, cascade = [CascadeType.ALL], orphanRemoval = true)
   @Fetch(FetchMode.SUBSELECT)
-  var exclusions: MutableList<Exclusion> = mutableListOf(),
+  val exclusions: MutableSet<Exclusion> = mutableSetOf(),
```
matt-crowson-moj, line 55: Oh and this one here as well, direct access.  Sorry missed this before.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Exclusion.kt
matt-crowson-moj: Ah yeah that makes total sense.

timharrison-moj (approved):
LGTM

matt-crowson-moj (approved):
LGTM

PR #693, opened 2023-11-20
SAA-1380 added attendance counter log statement.

matt-crowson-moj:
Just adding some simple logging to log number of attendances created for a prison when the job runs.


timharrison-moj (approved):
LGTM

PR #695, opened 2023-11-21
SAA-1383: Fixed allocation end date logic

leejacobson-moj:
## Overview

Small change to fix a bug where an updated allocation end date is validated against the existing allocation end date.


matt-crowson-moj (approved):
LGTM

PR #699, opened 2023-11-28
SAA-1396 changes to attendances to support paid and unpaid activities.

matt-crowson-moj:
**MEDIUM RISK**

Changes to support paid/unpaid attendance.

I am questioning the value of yet another attribute on Attendance.  Technically we can derive it in most cases. It could be useful for SQL queries perhaps?

One more change to follow this is updates to paid/unpaid activities.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Attendance.kt
```diff
@@ -51,14 +51,28 @@ data class Attendance(
...
+  @Transient
+  var initialIssuePayment: Boolean? = null,
 
   var caseNoteId: Long? = null,
 
   var incentiveLevelWarningIssued: Boolean? = null,
 
   var otherAbsenceReason: String? = null,
 ) {
+  val paid: Boolean = scheduledInstance.isPaid()
```
matt-crowson-moj, line 63: @timharrison-moj

I am questioning the value of having this `paid` attribute on the Attendance entity now.  I think in most cases, if not all, we can derive it.

matt-crowson-moj: Yes we will always have an existing attendance record when marking.

leejacobson-moj: I don't think it's needed here, like you say, it can be derived from the activity.

I'm also not sure about the naming of the attribute here because I think I'd assume an attendance record flagged `paid` would mean that the attendance has been paid, not that it's payable. At the activity level you don't have that ambiguity.

matt-crowson-moj: So interestingly I this was originally called `payable`.  Not sure why I changed it to this now.

matt-crowson-moj: I have removed the `paid` attribute from Attendance and just derive it when and where it is needed/used.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/AttendancesService.kt
```diff
@@ -59,7 +59,7 @@ class AttendancesService(
           reason = attendanceReasonsByCode[updateRequest.maybeAttendanceReason()],
           newStatus = updateRequest.status,
           newComment = updateRequest.comment,
-          newIssuePayment = updateRequest.issuePayment,
+          newIssuePayment = if (updateRequest.issuePayment == true && !attendance.isPayable()) false else updateRequest.issuePayment,
           newIncentiveLevelWarningIssued = updateRequest.incentiveLevelWarningIssued,
```
timharrison-moj, line 63: label?  `newIssuePayment` - won't be new for long :-)

timharrison-moj (approved):
LGTM. 
Important changes - so I'll get together a list of manual test scenarios to run through.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/AttendancesService.kt
matt-crowson-moj: Yeah so it was always called this. I guess so as not to conflict with the Attendance entity when the mark method is called.

PR #700, opened 2023-11-28
SAA-1373: Migration changes for default activity tiers, zero pay rates, optional pay bands,  rules on paid/unpaid activities

timharrison-moj:
This PR alters the migration endpoint for activities and allocations, to:

* Remove the (unused) Risley-specific split regime rules.
* Implement a generic split regime rule set - for future use.
* Adds additional incentive levels
* Cater for unpaid activities
* Cater for allocations with no pay band - to unpaid activities
* Cater for new rules around pay bands
* Map the NOMIS program service codes to default the activity tiers/organisers
* Implement additional tests for exclusions as part of allocation migration

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/MigrateActivityService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/MigrateActivityService.kt
@@ -18,18 +18,20 @@ import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.entity.EventTie
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.entity.PrisonPayBand
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.model.request.ActivityMigrateRequest
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.model.request.AllocationMigrateRequest
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.model.request.NomisPayRate
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.model.request.NomisScheduleRule
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.model.response.ActivityMigrateResponse
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.model.response.AllocationMigrateResponse
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.repository.ActivityCategoryRepository
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.repository.ActivityRepository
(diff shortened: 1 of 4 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/MigrateActivityService.kt
```diff
@@ -383,10 +351,85 @@ class MigrateActivityService(
...
+      // Association
+      programServiceCode.startsWith("ASSOC") -> tiers.isFoundation()
+
+      programServiceCode.startsWith("VIDEO") -> tiers.isFoundation()
+
+      // Everything else is tier 2 by default
+      else -> tiers.isTierTwo()
+    }
+
+    return tier ?: throw ValidationException("Could not map $programServiceCode to a tier")
```
matt-crowson-moj, line 432: Won't there always be a (default) tier here due to the `else` in the `when` above?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/MigrateActivityService.kt
```diff
@@ -383,10 +351,85 @@ class MigrateActivityService(
...
+  // Helper functions - tiers
+  fun List<EventTier>.isTierOne() = this.find { it.code == "TIER_1" }
+  fun List<EventTier>.isTierTwo() = this.find { it.code == "TIER_2" }
+  fun List<EventTier>.isFoundation() = this.find { it.code == "FOUNDATION" }
+
   fun mapProgramToTier(programServiceCode: String): EventTier? {
-    // We only have one tier
     val tiers = eventTierRepository.findAll()
-    return if (tiers.size > 0) tiers.first() else null
+    val tier = when {
```
matt-crowson-moj, line 361: Does it make sense for `tier` to be nullable.  If any of the tiers are missing then it should just fail right?

matt-crowson-moj (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/MigrateActivityService.kt
timharrison-moj: Will just check..

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/MigrateActivityService.kt
timharrison-moj: No - still required as utility functions may not find a tier in the list retrieved, so correct as-is I think.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/MigrateActivityService.kt
timharrison-moj: Tiers needs to be nullable - utility functions `isTierN()` may not find what they are looking for from data, and in this case, will fail with a validation exception.

PR #706, opened 2023-12-04
SAA-1362: Remove existing exclusions which would not sync to nomis on a change of a slot

smcveigh941:
We have already established that exclusions cannot be created on slots where the slot exists across multiple weeks in a multi-week schedule.

This PR is needed to remove existing exclusions which exist on a slot, if another slot is added on to another week in the schedule, so that the sync can happen successfully.

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/AllocationTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/AllocationTest.kt
@@ -611,6 +611,120 @@ class AllocationTest {
   }
 
   @Test
+  fun `syncExclusionsWithScheduleSlots - removes present exclusions on a multi-week model where they do not map to nomis data model`() {
+    val activity = activityEntity(noSchedules = true)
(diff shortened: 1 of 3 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/AllocationTest.kt
```diff
@@ -610,6 +610,120 @@ class AllocationTest {
...
+          monday = true,
+          thursday = true,
+        ),
+      ),
+    )
+
+    with(allocation) {
+      exclusions(ExclusionsFilter.PRESENT).single().getDaysOfWeek() isEqualTo setOf(DayOfWeek.MONDAY, DayOfWeek.FRIDAY)
+      exclusions(ExclusionsFilter.ACTIVE).single().getDaysOfWeek() isEqualTo setOf(DayOfWeek.FRIDAY)
+    }
```
timharrison-moj, line 667: So this has made the Monday exclusion in week 2 inactive? Because there is a Monday exclusion in week 1?

timharrison-moj (approved):
LGTM. I `think` I understand the test... almost.

PR #707, opened 2023-12-05
SAA-1412 persist offender merge event changes across affected entities

matt-crowson-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Allocation.kt
```diff
@@ -404,6 +413,15 @@ data class Allocation(
     exclusions.add(exclusion)
   }
 
+  fun merge(offenderMergeDetails: OffenderMergeDetails) {
+    require(offenderMergeDetails.oldNumber == prisonerNumber) {
+      "Current prisoner number '$prisonerNumber' does not match the old prisoner merge number '${offenderMergeDetails.oldNumber}' for allocation ID '$allocationId'"
+    }
+
+    prisonerNumber = offenderMergeDetails.newNumber
+    offenderMergeDetails.newBookingId?.let { bookingId = it }
```
matt-crowson-moj, line 422: @timharrison-moj 

Speaking with Mike H he seems to thing the booking number would remain the same, so not 100% sure if we event need to support changing of the booking ID.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/events/handlers/OffenderMergedEventHandler.kt
```diff
@@ -96,9 +87,47 @@ class OffenderMergedEventHandler(
...
-        message = "Prisoner number $oldNumber was merged to a new prisoner number $newNumber",
+        prisonCode = offenderMergeDetails.prisonCode,
+        message = "Prisoner number ${offenderMergeDetails.oldNumber} was merged to a new prisoner number ${offenderMergeDetails.newNumber}",
       ),
     )
   }
+
+  private fun mergeAllocations(mergeDetails: OffenderMergeDetails) {
+    allocationRepository.findByPrisonCodeAndPrisonerNumber(mergeDetails.prisonCode, mergeDetails.oldNumber)
+      .onEach { allocation -> allocation.merge(mergeDetails) }
```
matt-crowson-moj, line 98: @timharrison-moj only this one does anything at the moment.  I did wonder whether it is worth feature toggling this?  Easy enough to do.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Allocation.kt
timharrison-moj: I wasn't sure either - but the ticket included a bit of info on it, also suggesting the booking ID would not change. However, I don't see how that can be true .. two offender records, two bookings, merged into one - so one of them must be different.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/LocalAuditRecord.kt
```diff
@@ -55,4 +56,8 @@ data class LocalAuditRecord(
       message,
     )
   }
+
+  fun merge(offenderMergeDetails: OffenderMergeDetails) {
```
timharrison-moj, line 60: I suppose the idea of passing the responsibility for merging down into the entities has one drawback - that there needs to be an find, iterate through each row type loop somewhere. A quicker way might be to do this in one bulk SQL - update all attendance, all audit, all allocations?

timharrison-moj: Not saying which is right or wrong - just a thought.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/AllocationTest.kt
```diff
@@ -839,13 +841,39 @@ class AllocationTest {
...
+  fun `allocation offender merge fails if old number does not match`() {
+    val allocation = activityEntity().schedules().first().allocations().first().copy(allocatedPrisonerNumber = "ABC123", allocatedBookingId = 1)
+
+    assertThatThrownBy {
+      allocation.merge(OffenderMergeDetails(prisonCode = moorlandPrisonCode, oldNumber = "BDC123", newNumber = "DEF123", newBookingId = 2))
+    }.isInstanceOf(IllegalArgumentException::class.java)
+      .hasMessage("Current prisoner number 'ABC123' does not match the old prisoner merge number 'BDC123' for allocation ID '0'")
+
+    allocation.prisonerNumber isEqualTo "ABC123"
+    allocation.bookingId isEqualTo 1
```
timharrison-moj, line 877: Neat way of testing by moving the merge operation into the entity.

timharrison-moj (approved):
Looks good so far. 
A merge might take a while to iterate through some of the lists of entities - but is more testable this way. An alternative would be to treat this operation as an SQL data-update - to do them in one shot?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Allocation.kt
matt-crowson-moj: Yes but they always keep the latest booking ID on a merge so (in theory) we should always have the latest booking ID.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/LocalAuditRecord.kt
matt-crowson-moj: So thumbs up or down in terms of which way?

PR #708, opened 2023-12-05
SAA-1413: Ensure migrated activities match the prison regime

smcveigh941:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/PrisonRegime.kt
```diff
@@ -37,4 +38,18 @@ data class PrisonRegime(
 
   fun hasExpired(predicate: () -> LocalDate?) =
     predicate()?.onOrBefore(LocalDate.now().minusDays(maxDaysToExpiry.toLong())) == true
+
+  fun startTimeForTimeSlot(timeSlot: TimeSlot): LocalTime =
```
matt-crowson-moj, line 42: Suggestion. I'd just do this as one function which returns a `Pair<LocalTime, LocalTime>` e.g.

```
fun startAndEndTimesFor(timeSlot: TimeSlot) =
  when (timeSlot) {
    TimeSlot.AM -> amStart to amFinish
    TimeSlot.PM -> pmStart to pmFinish
    TimeSlot.ED -> edStart to edFinish
  }

```

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/MigrateActivityService.kt
```diff
@@ -546,4 +551,12 @@ class MigrateActivityService(
       ?: logAndThrowValidationException("Failed to delete. Activity $activityId was not found at prison $prisonCode")
     activityRepository.delete(activity)
   }
+
+  private fun ActivitySchedule.addRegimeSlot(weekNumber: Int, startTime: LocalTime, daysOfWeek: Set<DayOfWeek>, prisonCode: String): ActivityScheduleSlot {
+    val prisonRegime = prisonRegimeRepository.findByPrisonCode(prisonCode)!!
```
matt-crowson-moj, line 556: Minor optimisation suggestion. I wouldn't get the prison regime in here.  It will call it for every slot you create for an activity.

Would just pass in as a param perhaps.

PR #709, opened 2023-12-06
SAA-1415 fix to paid toggle on Activity. Calling with the same set value results in no actual change to the activity ergo not firing validation when not needed.

matt-crowson-moj:
Setting the paid attribute on the Activity with same value should result in no actual change to the activity.

As it stood it was firing some validation that shouldn't have been even if the value had not changed.


leejacobson-moj (approved):
Looks good. Approved.

PR #710, opened 2023-12-06
SAA-1413: Remove constraint on duplicate TimeSlot row on slots, and allow for exclusions on granular slots migrated from nomis

smcveigh941:


src/main/resources/migrations/common/V2023.12.06__exclusions_add_slot_end_time.sql
```diff
@@ -0,0 +1,87 @@
+-- ==================================================================
+-- Add slot_end_time column
+-- ==================================================================
+ALTER TABLE exclusion ADD COLUMN slot_end_time time;
+
+-- ==================================================================
+-- Populate the column with appropriate values
+-- ==================================================================
+-- update "active" exclusions first since we know for sure the slot it associates with
+UPDATE exclusion ex SET slot_end_time = (
```
matt-crowson-moj, line 10: Do we really need these updates?

It is only used in dev?  I'd be tempted to just bin/delete what exclusions we have and start on a clean slate.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Allocation.kt
```diff
@@ -326,9 +327,9 @@ data class Allocation(
   /**
    * Returns true if the date is between the start and end date, no clashing exclusions and not ended, otherwise false.
    */
-  fun canAttendOn(date: LocalDate, timeSlot: TimeSlot) = date.between(startDate, maybeEndDate()) && isExcluded(date, timeSlot).not() && prisonerStatus != PrisonerStatus.ENDED
+  fun canAttendOn(date: LocalDate, slotTimes: SlotTimes) = date.between(startDate, maybeEndDate()) && isExcluded(date, slotTimes).not() && prisonerStatus != PrisonerStatus.ENDED
 
-  fun syncExclusionsWithScheduleSlots(scheduleSlots: List<Slot>): Long? {
+  fun syncExclusionsWithScheduleSlots(scheduleSlots: List<Slot>, regimeTimeSlots: Map<TimeSlot, SlotTimes>): Long? {
```
matt-crowson-moj, line 332: I be tempted just pass in the regime entity itself here instead of naming something to say it is from the regime?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/ActivitySchedule.kt
```diff
@@ -341,32 +343,30 @@ data class ActivitySchedule(
     this.instancesLastUpdatedTime = LocalDateTime.now()
   }
 
-  fun updateSlots(updates: Map<Pair<Int, Pair<LocalTime, LocalTime>>, Set<DayOfWeek>>) {
+  fun updateSlots(updates: Map<Pair<Int, SlotTimes>, Set<DayOfWeek>>) {
```
matt-crowson-moj, line 346: Perhaps we should have a type (alias) or something for these? `Map<Pair<Int, SlotTimes>`.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/ActivitySchedule.kt
```diff
@@ -275,9 +276,10 @@ data class ActivitySchedule(
       ).apply {
         this.endDate = endDate?.also { deallocateOn(it, DeallocationReason.PLANNED, allocatedBy) }
         exclusions?.onEach { exclusion ->
-          slot(exclusion.weekNumber, exclusion.timeSlot())
-            .apply { require(this != null) { "Allocating to schedule ${activitySchedule.activityScheduleId}: No single ${exclusion.timeSlot()} slots in week number ${exclusion.weekNumber}" } }
-            .let { slot -> this.updateExclusion(slot!!, exclusion.getDaysOfWeek()) }
+          slots(exclusion.weekNumber, exclusion.timeSlot())
+            .apply { require(this.isNotEmpty()) { "Allocating to schedule ${activitySchedule.activityScheduleId}: No ${exclusion.timeSlot()} slots in week number ${exclusion.weekNumber}" } }
+            .filter { slot -> slot.getDaysOfWeek().intersect(exclusion.getDaysOfWeek()).isNotEmpty() }
```
matt-crowson-moj, line 281: Could we name some of the filters with extension functions at add readability?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/ActivitySchedule.kt
```diff
@@ -341,32 +343,30 @@ data class ActivitySchedule(
...
-    updates.keys.filterNot { key ->
-      slots.map { Pair(it.weekNumber, it.startTime to it.endTime) }.contains(key)
-    }.forEach {
-      addSlot(it.first, it.second.first, it.second.second, updates[it]!!)
-    }
+  private fun addNewSlots(updates: Map<Pair<Int, SlotTimes>, Set<DayOfWeek>>) {
+    updates.keys
+      .filterNot { key -> slots.map { Pair(it.weekNumber, it.slotTimes()) }.contains(key) }
+      .forEach { addSlot(it.first, it.second, updates[it]!!) }
   }
```
timharrison-moj, line 370: Small one - for these a comment might be useful to describe what's happening? What rules are we enforcing etc.

timharrison-moj (approved):
LGTM
Great work and look like it was a complicated set of changes.
I think we'll come back to this when considering schedules with specific times (not needed yet for Risley or Liverpool).

PR #712, opened 2023-12-07
SAA-1412 changes to support feature switched updates on the back of an offender merge event being intercepted

matt-crowson-moj:
As it stands this only merges (replaces the old number with) the new prisoner number.  It does not replace the existing booking ID. I have been told on the back of an offender merge the existing booking ID is maintained and does not change but the prisoner number does.

The actual updates are feature switched (off by default) so should be safe to release this as is.

Ideally going forwards we should seriously consider modelling a prisoner entity which should be associated with each affected entity. This way we would only have to deal with one single update to a prisoner entity instead of having to update all entities where a prisoner number is stored.  This should be done as a separate exercise, it is beyond the scope of this change at the moment.

### DIFF
--- a/helm_deploy/hmpps-activities-management-api/values.yaml
+++ b/helm_deploy/hmpps-activities-management-api/values.yaml
@@ -72,6 +72,7 @@ generic-service:
     FEATURE_AUDIT_SERVICE_HMPPS_ENABLED: false
     FEATURE_AUDIT_SERVICE_LOCAL_ENABLED: true
     FEATURE_MIGRATE_SPLIT_REGIME_ENABLED: false
+    FEATURE_OFFENDER_MERGE_ENABLED: false
 
   # Pre-existing kubernetes secrets to load as environment variables in the deployment.
   # namespace_secrets:
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/repository/AttendanceRepository.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/repository/AttendanceRepository.kt
@@ -1,6 +1,7 @@
 package uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.repository
 
 import org.springframework.data.jpa.repository.JpaRepository
+import org.springframework.data.jpa.repository.Modifying
 import org.springframework.data.jpa.repository.Query
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.entity.Attendance
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.entity.AttendanceStatus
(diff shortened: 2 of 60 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/events/handlers/OffenderMergedEventHandler.kt
```diff
@@ -96,9 +91,72 @@ class OffenderMergedEventHandler(
...
     )
   }
+
+  private fun mergeAllocations(mergeDetails: OffenderMergeDetails) {
+    allocationRepository.findByPrisonCodeAndPrisonerNumber(mergeDetails.prisonCode, mergeDetails.oldNumber)
+      .also { log.info("MERGE: Would alter ${it.size} allocation rows (including new booking id ${mergeDetails.newBookingId}") }
+
+    if (mergeIsEnabled) {
+      log.info("MERGE: old allocation prisoner number ${mergeDetails.oldNumber} to new prisoner number ${mergeDetails.newNumber}")
+      allocationRepository.mergeOffender(oldNumber = mergeDetails.oldNumber, newNumber = mergeDetails.newNumber)
```
matt-crowson-moj, line 107: **Note:** 

Updates do not restrict to prison codes, they apply updates across the board for any matching prisoner numbers.

helm_deploy/hmpps-activities-management-api/values.yaml
```diff
@@ -72,6 +72,7 @@ generic-service:
     FEATURE_AUDIT_SERVICE_HMPPS_ENABLED: false
     FEATURE_AUDIT_SERVICE_LOCAL_ENABLED: true
     FEATURE_MIGRATE_SPLIT_REGIME_ENABLED: false
+    FEATURE_OFFENDER_MERGE_ENABLED: false
```
matt-crowson-moj, line 75: The actual merge update is turned off by default.

It is safe to release this code as it stands as the previous functionality is pretty much the same as before.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/repository/AttendanceRepository.kt
```diff
@@ -62,4 +63,8 @@ interface AttendanceRepository : JpaRepository<Attendance, Long> {
   ): List<BookingCount>
 
   fun findByPrisonerNumber(prisonerNumber: String): List<Attendance>
+
+  @Query(value = "UPDATE Attendance a SET a.prisonerNumber = :newNumber WHERE a.prisonerNumber = :oldNumber")
```
matt-crowson-moj, line 67: Over time some of these updates could potentially be quite large, hence moving to modelling the Prisoner as an entity would remove this potential overhead/risk.

timharrison-moj: Agreed.

timharrison-moj: However, I think in almost all cases merges would happen a few days after a record is created, so the history of it will be very, very limited.

timharrison-moj (approved):
LGTM.
Very clear what its doing.

PR #713, opened 2023-12-08
SAA-1416: Remove minimum incentive level from activity requests and models

leejacobson-moj:
## Overview

Removes minimum incentive level columns from the activity table and related request and activity models

This change needs to be confirmed with syscon before merging.


src/main/resources/migrations/common/V2023.12.07.1__remove_minimum_incentive_levels.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE activity DROP COLUMN IF EXISTS minimum_incentive_nomis_code;
```
matt-crowson-moj, line 1: You might want to bump up the migration number to the current date?  There is already a migration for 11th Dec.

matt-crowson-moj (approved):
LGTM

PR #718, opened 2023-12-11
SAA-1427 changing back to the use the JRE after investigation into memory/heap sizes.

matt-crowson-moj:


timharrison-moj (approved):
LGTM

PR #719, opened 2023-12-12
SAA-1431 increase preprod memory limits to factor in JVM heap changes.

matt-crowson-moj:


helm_deploy/values-preprod.yaml
```diff
@@ -20,6 +20,14 @@ generic-service:
     FEATURE_AUDIT_SERVICE_HMPPS_ENABLED: true
     SENTRY_ENVIRONMENT: preproduction
 
+  resources:
+    requests:
+      cpu: 10m
+      memory: 768Mi
```
matt-crowson-moj, line 26: Default is 512.  Heap size is set at 512 which leaves no free space (I think).

timharrison-moj (approved):
LGTM

PR #721, opened 2023-12-12
SAA-1426: Migrate waitlist requester to use codes rather than description - To be deployed to prod alongside a UI change

smcveigh941:


timharrison-moj (approved):
LGTM. Careful watch on DEV and PREPROD before a prod deploy probably needed?

PR #727, opened 2023-12-14
SAA-1412 using prison API in favour of prisoner search in effort to avoid a race condition when looking up merged prisoner details in prisoner search.

matt-crowson-moj:
Switching to use prison API for looking up merged offender details instead of prisoner search API.

In the time we have received a prisoner merge event and attempted to process it (which in turn looks up the prisoner details via an external API call) it looks like the prisoner changes have not yet been indexed in prisoner search.

Hopefully moving to prisoner API should remove the race condition.  If it doesn't then we will have to revisit this.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/client/prisonapi/api/PrisonApiClient.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/client/prisonapi/api/PrisonApiClient.kt
@@ -31,6 +31,8 @@ inline fun <reified T> typeReference() = object : ParameterizedTypeReference<T>(
 @Service
 class PrisonApiClient(private val prisonApiWebClient: WebClient) {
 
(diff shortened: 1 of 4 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/client/prisonapi/api/PrisonApiClient.kt
```diff
@@ -31,6 +31,8 @@ inline fun <reified T> typeReference() = object : ParameterizedTypeReference<T>(
 @Service
 class PrisonApiClient(private val prisonApiWebClient: WebClient) {
 
+  fun getPrisonerDetailsLite(prisonerNumber: String): InmateDetail? = getPrisonerDetails(prisonerNumber, fullInfo = false).block()
+
   fun getPrisonerDetails(
     prisonerNumber: String,
     fullInfo: Boolean = true,
```
matt-crowson-moj, line 38: Not to be answered here but I wonder if `fullInfo` should be `false` by default ...

PR #728, opened 2023-12-14
SAA-1404: Refactor posting a case note to allow passing in a subType

smcveigh941:
Some prep for upcoming case notes work
Also added a pre-commit hook for linting


pre-commit
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+echo "*****Running Lint******"
+./gradlew ktlintFormat
+exit $?
```
timharrison-moj, line 4: Is this in danger of formatting differently depending on local settings? I can try when merged.

timharrison-moj (approved):
LGTM

PR #730, opened 2023-12-14
SAA-1447: Remove waitlist applications when the prisoner is released

smcveigh941:


timharrison-moj (approved):
LGTM.
There is another conversation happening with DPR, as they want to know the history of status changes on the waiting list, but can be added later as a table recording each status change, or removal.

PR #731, opened 2023-12-15
SAA-1412 enable offender merge updates in preprod and prod.

matt-crowson-moj:
Also improving indexing on the local audit table which the offender merge process will use when retrieving and updating prisoner audit records.

### DIFF
--- a/src/main/resources/migrations/common/V2023.12.15__add_index_to_local_audit.sql
+++ b/src/main/resources/migrations/common/V2023.12.15__add_index_to_local_audit.sql
@@ -1,0 +1,2 @@
+CREATE INDEX IF NOT EXISTS idx_local_audit_prison_code_prisoner_number on local_audit(prison_code, prisoner_number);
+CREATE INDEX IF NOT EXISTS idx_local_audit_prisoner_number on local_audit(prisoner_number);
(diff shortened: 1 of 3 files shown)


src/main/resources/migrations/common/V2023.12.15__add_index_to_local_audit.sql
```diff
@@ -0,0 +1,2 @@
+CREATE INDEX IF NOT EXISTS idx_prison_code_prisoner_number on local_audit(prison_code, prisoner_number);
```
matt-crowson-moj, line 1: Also included as noticed missing indexes of audit table which will be hit by the offender merge queries.

timharrison-moj: Are these unique names, Matt? I usually create with the table and column names combined to make each unique.

timharrison-moj (dismissed):
LGTM

src/main/resources/migrations/common/V2023.12.15__add_index_to_local_audit.sql
matt-crowson-moj: Will change, yes unique as ran locally.

timharrison-moj (approved):
LGTM

PR #732, opened 2023-12-18
SAA-1436 3rd party dependency updates and small code changes to help resolve security warnings.

matt-crowson-moj:


.trivyignore
```diff
@@ -7,28 +7,9 @@
...
-# Suppression for jackson databind 2.13.3 as bundled with application insights
-#   Can be suppressed as don't parse untrusted json in application insights
-CVE-2022-42004
-# Suppression for apache common-text 1.9 as bundled with application insights
-#   can be suppressed for the time being as it will be fixed in next version of application insights
-CVE-2022-42889
 # Suppression for h2 2.1.214 password on command line vulnerability
 #   can be suppressed as we only run h2 locally and not on build environments
 CVE-2022-45868
+# Suppression for logback-classic and logback-core as we don't let third parties control our appenders.
```
matt-crowson-moj, line 13: This is changed as a result of picking up a more recent version of the dps gradle plugin.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/config/OpenApiConfiguration.kt
```diff
@@ -87,7 +88,7 @@ class OpenApiConfiguration(buildProperties: BuildProperties) {
     }
   }
 
-  @Bean
+  @PostConstruct
```
matt-crowson-moj, line 91: Started failing on tests so had to change. It was complaining about this being void and not returning a (Bean) type which is technically correct.

The end result is the same and the partial time is enabled.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/resource/WaitingListApplicationControllerTest.kt
```diff
@@ -67,7 +68,7 @@ class WaitingListApplicationControllerTest : ControllerTestBase<WaitingListAppli
   fun `200 response when searching waiting list application`() {
     val request = WaitingListSearchRequest()
     val waitingListApplication = waitingList().toModel()
-    val pagedResult = PageImpl(listOf(waitingListApplication))
+    val pagedResult = PageImpl(listOf(waitingListApplication), Pageable.ofSize(1), 1)
```
matt-crowson-moj, line 71: Started failing as a result of upgrading.  Now need to specify the size of the expected page result in the test for it to pass.

Integration test is unaffected here so all good there.

build.gradle.kts
```diff
@@ -140,6 +140,8 @@ tasks.named<Test>("integrationTest") {
   filter {
     includeTestsMatching("*.integration.*")
   }
+  minHeapSize = "512m"
```
matt-crowson-moj, line 143: Was getting out of memory errors when running the integration tests.

build.gradle.kts
```diff
@@ -140,6 +140,7 @@ tasks.named<Test>("integrationTest") {
   filter {
     includeTestsMatching("*.integration.*")
   }
+  maxHeapSize = "1024m"
```
matt-crowson-moj, line 143: Getting out of memory errors locally.  Have increased in the circle config also.

PR #733, opened 2023-12-19
SAA-1453 removes the allocation and schedule entity listeners by using the outbound message service direct in the services.

matt-crowson-moj:
**HIGH RISK - DO NOT MERGE**

Other than the auditing entity listener (which is staying) this is the final nail in the coffin for entity listeners that were publishing events.

This change moves the allocation creation, schedule creation and schedule update events directly into the places where they are needed whilst also wrapping the transactions in sub-transactions to ensure commits are complete before raising the events.

Whilst this change was always on the cards it has been brought forwards due to sync issues in production when a transaction fails our side.  Even though transactions were quite rightly failing in the API e.g. trying to allocate the same person twice to an activity (via a double click in the UI), sync events were still being published effectively causing ghost events to be raised.  In this allocation example mentioned it was sending an allocation created event for a non-existent allocation.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityScheduleService.kt
```diff
@@ -141,71 +137,71 @@ class ActivityScheduleService(
...
-      val maybeWaitingList = waitingListRepository.findByPrisonCodeAndPrisonerNumberAndActivitySchedule(
-        schedule.activity.prisonCode,
-        request.prisonerNumber,
-        schedule,
-      )
-        .also {
-          val pending = it.filter { it.status == WaitingListStatus.PENDING }
-          require(pending.isEmpty()) {
-            "Prisoner has a PENDING waiting list application. It must be APPROVED before they can be allocated."
+      val activePrisoner = prisonerSearchApiClient.findByPrisonerNumber(request.prisonerNumber)
```
matt-crowson-moj, line 157: Replaced with prisoner search as we do not need to be using prison API here.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/MigrateActivityService.kt
```diff
@@ -73,44 +77,52 @@ class MigrateActivityService(
       throw ValidationException("The requested prison ${request.prisonCode} is not rolled-out for activities")
     }
 
-    // Get the pay bands/aliases configured for this prison
-    val payBands = prisonPayBandRepository.findByPrisonCode(request.prisonCode)
+    return transactionHandler.newSpringTransaction {
+      // Get the pay bands/aliases configured for this prison
```
timharrison-moj, line 81: This is called directly by the migration service, once per activity being migrated, so probably doesn't need to be wrapped in a transactional block? Or do you think the timing of the event and it being picked up and actioned could happen before this thread has completed?

timharrison-moj (approved):
LGTM. Would need to go on its own to DEV and get a good shakedown there - test the reconciliation that Mike has etc..

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/MigrateActivityService.kt
matt-crowson-moj: You are probably right in that it would be picked up by a retry if the timing was out.

I still thing we should wrap it though so it is consistent with other code changes for raising events i.e. make sure everything is persisted before raising anything.  Belt and braces really.

timharrison-moj (approved):
LGTM

PR #734, opened 2023-12-19
SAA-1415: Connection pool settings to cater for high load during migrations.

timharrison-moj:
There are two changes here:

1. Adjust connection pool size, idle time and wait times so that it will grow to cater for high-activity during migrations.
2. Set the queue admin role to ROLE_ACTIVITY_QUEUE_ADMIN that we can give to our personal clients.


matt-crowson-moj (approved):
LGTM

PR #739, opened 2024-01-04
SAA-1480 reducing the data requested (and load) from prison API in the interesting event handler as we don't need the extra information.

matt-crowson-moj:
The interesting event handler has been requesting full prisoner information from prison-api when it doesn't actually need it.  

The impact of doing so means prison-api has to do a lot more work in terms of servicing the requests here.

This change means we are asking for minimal information which in turns reduces the load on an already heavy hit prison-api.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/events/handlers/InterestingEventHandler.kt
```diff
@@ -120,7 +124,7 @@ class InterestingEventHandler(
     return Outcome.success()
   }
 
-  private fun InboundEvent.getEventMessage(prisoner: InmateDetail, oldPrisonerNumber: String? = null): String {
```
matt-crowson-moj, line 123: `oldPrisonerNumber` is unused.

PR #740, opened 2024-01-04
SAA-1448: Return previous de-allocation from an activity when checking suitability for allocation

smcveigh941:
Returns reasons for previous de-allocation from an activity when checking suitability for allocation, along with any case notes which have been added to that deallocation.

**Note** that the user is not able to add a case note for a deallocation yet, this PR only adds the work to return that case note when one has been added

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/client/casenotesapi/api/CaseNotesApiClient.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/client/casenotesapi/api/CaseNotesApiClient.kt
@@ -19,15 +19,11 @@ class CaseNotesApiClient(@Qualifier("caseNotesApiWebClient") private val webClie
       .block()
   }
 
-  fun getCaseNote(prisonerNumber: String, caseNoteId: Long?): CaseNote? {
-    return if (caseNoteId != null) {
-      webClient.get()
-        .uri("/case-notes/{offenderNo}/{caseNoteId}", prisonerNumber, caseNoteId)
-        .retrieve()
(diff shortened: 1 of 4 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/client/casenotesapi/api/CaseNotesApiClient.kt
```diff
@@ -20,10 +20,14 @@ class CaseNotesApiClient(@Qualifier("caseNotesApiWebClient") private val webClie
   }
 
   fun getCaseNote(prisonerNumber: String, caseNoteId: Long?): CaseNote? {
```
matt-crowson-moj, line 22: I realise it was already like this but why would this even be able to pass in a null case note id? 

Perhaps make it mandatory (non-nullable) instead.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/client/casenotesapi/api/CaseNotesApiClientTest.kt
```diff
@@ -45,6 +45,12 @@ class CaseNotesApiClientTest {
     assertThat(caseNote?.text).isEqualTo("Case Note Text")
   }
 
+  @Test
+  fun `getCaseNote without passing ID - success`() {
+    val caseNote = caseNotesApiClient.getCaseNote("A1234AA", null)
```
matt-crowson-moj, line 50: Kinda linked to my other comment, don't think this is needed/makes sense?

matt-crowson-moj (approved):
LGTM

PR #743, opened 2024-01-05
SAA-1460: Not in work activities filter

leejacobson-moj:
## Overview

Updates behaviour of in work filter to consider "not in work" activity allocations.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/CandidatesService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/CandidatesService.kt
@@ -16,8 +16,8 @@ import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.client.prisoner
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.client.prisonersearchapi.model.PrisonerAlert
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.entity.Activity
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.entity.ActivityMinimumEducationLevel
-import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.entity.DeallocationReason
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.entity.Allocation
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.entity.DeallocationReason
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.entity.PrisonerStatus
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.entity.WaitingListStatus
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.model.response.ActivityCandidate
@@ -129,16 +129,17 @@ class CandidatesService(
 
     prisoners = prisoners.filter { filterByEmployment(it, prisonerAllocations, inWork) }
 
-    return prisoners.map {
+    return prisoners.map { prisoner ->
       val thisPersonsAllocations = prisonerAllocations.toModelPrisonerAllocations()
-          .find { a -> it.prisonerNumber == a.prisonerNumber }?.allocations ?: emptyList()
+        .filter { a -> prisoner.prisonerNumber == a.prisonerNumber }
+        .flatMap { it.allocations }
 
       ActivityCandidate(
-        name = "${it.firstName} ${it.lastName}",
-        prisonerNumber = it.prisonerNumber,
-        cellLocation = it.cellLocation,
+        name = "${prisoner.firstName} ${prisoner.lastName}",
+        prisonerNumber = prisoner.prisonerNumber,
+        cellLocation = prisoner.cellLocation,
         otherAllocations = thisPersonsAllocations,
-        earliestReleaseDate = determineEarliestReleaseDate(it),
+        earliestReleaseDate = determineEarliestReleaseDate(prisoner),
       )
     }
   }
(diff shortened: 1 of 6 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/CandidatesService.kt
```diff
@@ -100,7 +100,7 @@ class CandidatesService(
     scheduleId: Long,
     suitableIncentiveLevels: List<String>?,
     suitableRiskLevels: List<String>?,
-    suitableForEmployed: Boolean?,
+    inWork: Boolean?,
```
leejacobson-moj, line 103: I decided to rename this because I found the wording confusing.

Firstly, this filter isn't checking whether the prisoner is suitable for employment, but just checking work status.

I also thought the use was confusing because the current behaviour meant that if `suitableForEmployed == true` only prisoners with allocations would be returned, where as the filter can be used in the opposite way – to find prisoners without work to be allocated for work.

Finally, this filter is used to support the "in work" filter on the UI so `inWork` brings it inline with the UI.

I haven't change the request param in the controller which is still called "suitableForEmployed", but I'll raise a ticket to update that separately. I thought there would be too many unrelated changes if I also updated that in this PR.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/CandidatesService.kt
```diff
@@ -126,15 +126,12 @@ class CandidatesService(
...
-    prisoners =
-      prisoners.filter { filterByEmployment(it, prisonerAllocations, suitableForEmployed) }
+    prisoners = prisoners.filter { filterByEmployment(it, prisonerAllocations, inWork) }
 
     return prisoners.map {
-      val thisPersonsAllocations =
-        prisonerAllocations.find { a -> it.prisonerNumber == a.prisonerNumber }?.allocations
-          ?: emptyList()
+      val thisPersonsAllocations = prisonerAllocations.toModelPrisonerAllocations()
+        .find { a -> it.prisonerNumber == a.prisonerNumber }?.allocations ?: emptyList()
```
matt-crowson-moj, line 134: Can this also be changed to use `filter  { ... }` instead of `find { ... }` ?  Should remove the `?: emptyList()` part is all I think.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/CandidatesServiceTest.kt
```diff
@@ -613,10 +621,88 @@ class CandidatesServiceTest {
...
+
+      val candidates = service.getActivityCandidates(
+        activity.schedules().first().activityScheduleId,
+        null,
+        null,
+        false,
+        null,
+      )
+
+      candidates.find { it.prisonerNumber == "A1234BC" } isNotEqualTo null
```
matt-crowson-moj, line 658: A word of caution using `find` here.  This doesn't tell you if there is more than one.  It finds the first matching, technically there could be more than one.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/CandidatesServiceTest.kt
```diff
@@ -613,10 +621,88 @@ class CandidatesServiceTest {
...
+
+      val candidates = service.getActivityCandidates(
+        activity.schedules().first().activityScheduleId,
+        null,
+        null,
+        true,
+        null,
+      )
+
+      candidates.find { it.prisonerNumber == "A1234BC" } isEqualTo null
```
matt-crowson-moj, line 697: Same here with use of `find`.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/CandidatesServiceTest.kt
```diff
@@ -613,10 +621,88 @@ class CandidatesServiceTest {
...
+      candidates.find { it.prisonerNumber == "A1234BC" } isNotEqualTo null
+      candidates.find { it.prisonerNumber == "B2345CD" } isEqualTo null
+      candidates.find { it.prisonerNumber == "C3456DE" } isNotEqualTo null
+    }
+
+    @Test
+    fun `Employment filter should filter candidates with allocations that are not for 'not in work' activities when inWork == true`() {
+      val activity = activityEntity()
+
+      val notInWorkActivity = activityEntity(category = activityCategory("SAA_NOT_IN_WORK"))
```
matt-crowson-moj, line 667: Minor comment really, could push the `activityCategory("SAA_NOT_IN_WORK")` into the `ActivityFactory` class e.g.

`val notInWorkCategory = activityCategory("SAA_NOT_IN_WORK")`

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/CandidatesServiceTest.kt
leejacobson-moj: Yeah, that's true. It shouldn't happen, but no harm asserting against the number of matches here just in case.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/CandidatesServiceTest.kt
```diff
@@ -613,10 +620,88 @@ class CandidatesServiceTest {
...
+
+      val candidates = service.getActivityCandidates(
+        activity.schedules().first().activityScheduleId,
+        null,
+        null,
+        false,
+        null,
+      )
+
+      candidates.filter { it.prisonerNumber == "A1234BC" }.size isEqualTo 1
```
matt-crowson-moj, line 657: Could use hasSize meh - `candidates.filter { it.prisonerNumber == "XXXXXX" } hasSize N`

matt-crowson-moj (approved):
LGTM

PR #745, opened 2024-01-07
SAA-1467: Auto-remove waitlist applications on released event and de-allocate job

smcveigh941:


timharrison-moj (approved):
LGTM. Just one question.

PR #753, opened 2024-01-11
SAA-1464 tidy up calls to prisoner API when getting prisoner information to restrict to minimal view of prisoner only, we don't need anymore than that.

matt-crowson-moj:
This PR tidy's up a loose end with regards to one of the prison API calls.

It warranted some documentation as well to explain why it is the way it is.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/client/prisonapi/api/PrisonApiClient.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/client/prisonapi/api/PrisonApiClient.kt
@@ -1,5 +1,6 @@
 package uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.client.prisonapi.api
 
+import kotlinx.coroutines.runBlocking
 import org.springframework.core.ParameterizedTypeReference
 import org.springframework.stereotype.Service
 import org.springframework.web.reactive.function.client.WebClient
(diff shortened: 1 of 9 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/client/prisonapi/api/PrisonApiClient.kt
```diff
@@ -28,26 +29,32 @@ typealias PrisonLocations = Map<Long, Location>
...
+  /**
+   * Returns a minimal view of the prisoner's attributes in the [InmateDetail].
+   *
+   * The rationale for a minimal view is down to what we actually need in terms of information but also to reduce the
+   * load on the prison API. Asking for more information than we actually need puts unnecessary load on the prison API.
+   *
+   * With the minimal view 'status' isn't populated but 'activeFlag' is. This means you can check if a prisoner is
+   * active at a prison but not say if they are IN or OUT of the prison at the time.
+   */
+  fun getPrisonerDetailsLite(prisonerNumber: String): InmateDetail = runBlocking {
```
matt-crowson-moj, line 49: Will never be null.  If there is no match then will get a 404 anyway.  The same likely goes for anywhere we return nulls from out external API calls.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/events/handlers/InterestingEventHandler.kt
```diff
@@ -115,7 +115,6 @@ class InterestingEventHandler(
           ),
         )
         log.debug("Saved interesting event ID ${saved.eventReviewId} - ${mergedEvent.eventType()} - replaced ${mergedEvent.removedPrisonerNumber()} with ${mergedEvent.prisonerNumber()}")
-        return Outcome.success()
```
matt-crowson-moj, line 118: Line redundant/not needed.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/client/prisonapi/api/PrisonApiClient.kt
```diff
@@ -28,26 +29,32 @@ typealias PrisonLocations = Map<Long, Location>
...
+    prisonApiWebClient.get()
       .uri { uriBuilder: UriBuilder ->
         uriBuilder
           .path("/api/bookings/offenderNo/{prisonerNumber}")
-          .queryParam("fullInfo", fullInfo)
-          .maybeQueryParam("extraInfo", extraInfo)
           .build(prisonerNumber)
       }
       .retrieve()
-      .bodyToMono(typeReference<InmateDetail>())
```
matt-crowson-moj, line 50: Using bodyToMono forces you to handle possible null values.  This can never happen as will get http responses e.g. 4xx, 5xx etc.

PR #756, opened 2024-01-15
SAA-1487: Fetch prison incentive levels from incentives API during migration

leejacobson-moj:
## Overview

Adds incentives API client and updates activity migration service call to fetch prison incentive levels.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/client/incentives/api/IncentivesApiClient.kt
```diff
@@ -0,0 +1,24 @@
...
+      .uri("/incentive/prison-levels/$prisonId")
+      .retrieve()
+      .bodyToMono(typeReference<List<PrisonIncentiveLevel>>())
+      .block() ?: emptyList()
+  }
+
+  @Cacheable(CacheConfiguration.PRISON_INCENTIVE_LEVELS_CACHE_NAME)
+  fun getIncentiveLevelsCached(prisonId: String): List<PrisonIncentiveLevel> =
+    getIncentiveLevels(prisonId)
+}
```
leejacobson-moj, line 24: This method does the same as above, but caches the result to reduce the number of API calls to the incentives API while migrating activities.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/MigrateActivityService.kt
```diff
@@ -92,14 +97,14 @@ class MigrateActivityService(
...
             val payBand = payBands.find { pb -> pb.nomisPayBand.toString() == it.nomisPayBand }
-            if (payBand != null) {
-              activity.addPay(it.incentiveLevel, mapIncentiveLevel(it.incentiveLevel), payBand, it.rate, null, null)
-            } else {
-              logAndThrowValidationException("Failed to migrate activity ${request.description}. No prison pay band for Nomis pay band ${it.nomisPayBand}")
-            }
+              ?: logAndThrowValidationException("Failed to migrate activity ${request.description}. No prison pay band for Nomis pay band ${it.nomisPayBand}")
+
+            val iep = prisonIncentiveLevels.find { iep -> iep.levelCode == it.incentiveLevel && iep.active }
+              ?: logAndThrowValidationException("Failed to migrate activity ${request.description}. Activity incentive level ${it.incentiveLevel} is not active in this prison")
```
leejacobson-moj, line 105: Previously the code would ignore pay rates with unknown IEP levels. I've updated this to now throw a validation error since there is no reason that activity migration request should be sent with an invalid IEP level now we are pulling them from the incentives API. If it happens it's more likely the request is invalid or there is a problem we should be investigating.

generate-incentives-api-types.sh
```diff
@@ -0,0 +1,4 @@
+openapi-generator generate -i https://incentives-api-dev.hmpps.service.justice.gov.uk/v3/api-docs \
```
matt-crowson-moj, line 1: Really going forwards we should avoid using an installed on your local machine generator this way as you cannot guarantee the version being used.

This also means you should not need to add copies of model objects in our own packages.

We should be using the gradle plugin, see [here in the build file](https://github.com/ministryofjustice/hmpps-activities-management-api/blob/main/build.gradle.kts#L105) for an example

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/client/incentives/model/PrisonIncentiveLevel.kt
```diff
@@ -0,0 +1,63 @@
...
+ * @param active Indicates that this incentive level is enabled in this prison
+ * @param defaultOnAdmission Indicates that this incentive level is the default for new admissions
+ * @param remandTransferLimitInPence The amount transferred weekly from the private cash account to the spends account for a remand prisoner to use
+ * @param remandSpendLimitInPence The maximum amount allowed in the spends account for a remand prisoner
+ * @param convictedTransferLimitInPence The amount transferred weekly from the private cash account to the spends account for a convicted prisoner to use
+ * @param convictedSpendLimitInPence The maximum amount allowed in the spends account for a convicted prisoner
+ * @param visitOrders The number of weekday visits for a convicted prisoner per fortnight
+ * @param privilegedVisitOrders The number of privileged/weekend visits for a convicted prisoner per 4 weeks
+ */
+data class PrisonIncentiveLevel(
```
matt-crowson-moj, line 18: If we use build file to generate the classes then this shouldn't be needed.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/client/incentives/api/IncentivesApiClient.kt
```diff
@@ -0,0 +1,24 @@
...
+  fun getIncentiveLevels(prisonId: String): List<PrisonIncentiveLevel> {
+    return incentivesApiWebClient.get()
+      .uri("/incentive/prison-levels/$prisonId")
+      .retrieve()
+      .bodyToMono(typeReference<List<PrisonIncentiveLevel>>())
+      .block() ?: emptyList()
+  }
+
+  @Cacheable(CacheConfiguration.PRISON_INCENTIVE_LEVELS_CACHE_NAME)
+  fun getIncentiveLevelsCached(prisonId: String): List<PrisonIncentiveLevel> =
```
matt-crowson-moj, line 22: Nice to have this cached I think as very unlikely to change.  Do we really need the two methods though one cached and one not?

The fact it is cached is implementation detail.  Just my take on it not saying you have to change it though.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/config/WebClientConfiguration.kt
```diff
@@ -78,6 +79,20 @@ class WebClientConfiguration(
...
   }
 
+  @Bean
+  @RequestScope
+  fun incentivesApiWebClient(
+    clientRegistrationRepository: ClientRegistrationRepository,
+    authorizedClientRepository: OAuth2AuthorizedClientRepository,
+    builder: WebClient.Builder,
+  ): WebClient {
+    return getPrisonApiOAuthWebClient(
```
matt-crowson-moj, line 89: I know technically under the covers they are using the same creds but really should be using it owns client registration and not prison-api's in the in the application.yml file.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/MigrateActivityService.kt
```diff
@@ -77,6 +76,12 @@ class MigrateActivityService(
       throw ValidationException("The requested prison ${request.prisonCode} is not rolled-out for activities")
     }
 
+    // Get the incentive levels for this prison
+    val prisonIncentiveLevels = incentivesApiClient.getIncentiveLevelsCached(request.prisonCode)
+    if (prisonIncentiveLevels.isEmpty()) {
+      throw RuntimeException("No incentive levels found for the requested prison ${request.prisonCode}")
```
matt-crowson-moj, line 82: Minor point here but why not just throw a ValidationException for consistency?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/client/incentives/api/IncentivesApiClient.kt
```diff
@@ -0,0 +1,24 @@
...
+import org.springframework.stereotype.Service
+import org.springframework.web.reactive.function.client.WebClient
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.client.incentives.model.PrisonIncentiveLevel
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.client.prisonapi.api.typeReference
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.config.CacheConfiguration
+
+@Service
+class IncentivesApiClient(private val incentivesApiWebClient: WebClient) {
+
+  fun getIncentiveLevels(prisonId: String): List<PrisonIncentiveLevel> {
```
matt-crowson-moj, line 13: This is a style thing for me so more of a personal opinoin.  Not a fan of the Mono API, it has led to some rather clunky look code IMO.  To be fair this is hidden here in that the Mono is not returned.

Prefer a push towards using co-routines API even if it is blocking under the covers example [here](https://github.com/ministryofjustice/hmpps-activities-management-api/blob/main/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/client/prisonapi/api/PrisonApiClient.kt#L49).

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/config/WebClientConfiguration.kt
```diff
@@ -78,6 +79,20 @@ class WebClientConfiguration(
     )
   }
 
+  @Bean
+  @RequestScope
```
matt-crowson-moj, line 83: This is not really aimed at just your change (maybe I am missing something) but why is a requestScope bean?

I have seen this on a few of our API clients configs as well but never really questioned it.

generate-incentives-api-types.sh
leejacobson-moj: Thanks! Wasn't aware of that. Updated now to use the gradle plugin and have removed the model.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/config/WebClientConfiguration.kt
leejacobson-moj: Thanks for spotting – that wasn't intentional. Agreed that it should have it's own client config. Updated now.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/client/incentives/api/IncentivesApiClient.kt
leejacobson-moj: Updated to use coroutines

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/config/WebClientConfiguration.kt
leejacobson-moj: I've removed and tested without `@RequestScore` annotation and it seems to be working fine. I wondered whether it was added to ensure a valid token is always generated for each request, but the docs seem to suggest new tokens are generated automatically if they've expired so that shouldn't be an issue. We also don't have it on non-associations client so if it was needed we should have seen issues by now. I'm fairly sure it's safe to remove.

matt-crowson-moj (approved):
LGTM

PR #757, opened 2024-01-15
SAA-1499 renaming method to remove need for ktlint suppression. Also tied up impacted integration test around change.

matt-crowson-moj:


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/integration/PrisonIntegrationTest.kt
```diff
@@ -57,18 +62,17 @@ class PrisonIntegrationTest : IntegrationTestBase() {
         createdTime = LocalDateTime.of(2022, 9, 21, 0, 0, 0),
         activityState = ActivityState.LIVE,
         paid = true,
-      ),
-    )
+      )
   }
 
   @Sql(
     "classpath:test_data/seed-activity-id-1.sql",
```
matt-crowson-moj, line 69: Technically the seed file used here (which is also used in other integrations tests) does not really reflect the 1:1 mapping we now have between activities and schedules i.e. the seed file has a single activity with multiple schedules which is not longer the case.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/repository/ActivityScheduleRepository.kt
```diff
@@ -11,8 +11,7 @@ import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.entity.Activity
 @Repository
 interface ActivityScheduleRepository : JpaRepository<ActivitySchedule, Long>, ActivityScheduleRepositoryCustom {
   @EntityGraph(attributePaths = ["instances"], type = EntityGraph.EntityGraphType.LOAD)
-  @Suppress("ktlint:standard:function-naming")
```
matt-crowson-moj, line 14: @smcveigh941 

Didn't fancy doing this change on a Friday afternoon as wanted to do a bit more tidy up on it.

matt-crowson-moj: This should work according to the Spring Docs [here](https://docs.spring.io/spring-data/jpa/reference/repositories/query-methods-details.html#repositories.query-methods.query-property-expressions)

Integration tests do back it up.

PR #758, opened 2024-01-15
Updated timestamp should not be populated on the first creation of an activity during migration

smcveigh941:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/MigrateActivityService.kt
```diff
@@ -221,8 +221,6 @@ class MigrateActivityService(
       riskLevel = DEFAULT_RISK_LEVEL,
       createdTime = LocalDateTime.now(),
       createdBy = MIGRATION_USER,
-      updatedTime = LocalDateTime.now(),
```
matt-crowson-moj, line 224: Should break a test really ...

matt-crowson-moj (approved):
LGTM

PR #759, opened 2024-01-16
SAA-1502 adding logging to the WebClientConfiguration to monitor usage/instance creation in particular for @RequestScope beans.

matt-crowson-moj:
Adding to configuration logging for monitoring purposes to see if we need the config the way it is or whether it can/should be changed.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/config/WebClientConfiguration.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/config/WebClientConfiguration.kt
@@ -1,5 +1,7 @@
 package uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.config
 
+import org.slf4j.Logger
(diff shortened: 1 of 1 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/config/WebClientConfiguration.kt
```diff
@@ -71,21 +74,19 @@ class WebClientConfiguration(
     clientRegistrationRepository: ClientRegistrationRepository,
```
matt-crowson-moj, line 74: This has been a RequestScope been from the very early days of the project a long with some others.  Not sure it is actually needed.

Keen to know how often these are called/created.

matt-crowson-moj: OK right well if that is the case then I think there must be an alternative way to achieve this.

As it stands to me this feels a little heavy handed, I could be completly wrong though.

PR #760, opened 2024-01-17
SAA-1497: Fix check for expired movements at a given prison

smcveigh941:
Our deallocation job currently identifies expired prisoners by checking that the latest movement from the from the prison is more than 5 days ago. However, the endpoint to fetch these is missing the `latestOnly = false` query parameter. This means that if someone has multiple transfers within the 5 days, then only the movement out of the latest prison will be returned, and the movement from the prison in question will not be returned.

We have a case in production where someone had movements from Risley -> Peterborough -> Norwich, and only the movement from Peterborough -> Norwich was returned from the movements endpoint, and the person was not correctly identified as an expired prisoner from Risley

Adding this query parameter means that the movement out of Risley is returned, and the service layer will identify this movement as expired.


matt-crowson-moj (approved):
LGTM

PR #763, opened 2024-01-23
SAA-1498: Return prison code and prisoner status with the allocations and appointments

smcveigh941:
Suspended (i.e. transferred) prisoners are currently having their cell location shown on their allocation/appointment records on the UI, even if they have transferred to a new prison.

For example, Risley can see cell locations from Kirkham on suspended allocation records. I am adding the prison code and prisoner status to the prisoner information returned from these endpoints so that the UI can display a more suitable cell location (i.e. "Out of prison")


matt-crowson-moj (approved):
LGTM

PR #765, opened 2024-01-25
SAA-1514 3rd party dependency updates to resolve OWASP security warnings.  Also had to update a failing test on back of changes.

matt-crowson-moj:


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/client/prisonapi/api/PrisonApiClientTest.kt
```diff
@@ -349,9 +349,10 @@ class PrisonApiClientTest {
     val agencyId = "LEI"
     val locationType = "CELL"
     prisonApiMockServer.stubGetLocationsForTypeUnrestrictedNotFound(agencyId, locationType)
-    assertThatThrownBy { prisonApiClient.getLocationsForTypeUnrestricted(agencyId, locationType).block() }
-      .isInstanceOf(WebClientResponseException::class.java)
-      .hasMessage("404 Not Found from GET http://localhost:8999/api/agencies/LEI/locations?eventType=CELL")
+
+    assertThrows<WebClientResponseException>(
```
matt-crowson-moj, line 353: So after changes this test started to fail.

Changing the assertion to work this way seemed to resolve it.  I am not really sure why, it is a stubbed call anyway so not overly concerned.

PR #769, opened 2024-01-30
SAA-1527: Case note prefixing is moved to the API layer from the UI layer

smcveigh941:
Moving the prefixing to the API improves user experience by:

- Reducing the number of API calls from the frontend needed to get the information
- User does not see extra data added into the fields which they did not enter
- Removes the chance of the prefix being duplicated when the user edits the case note

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/common/LocalDateTimeExt.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/common/LocalDateTimeExt.kt
@@ -2,8 +2,8 @@ package uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.common
 
 import java.time.LocalDateTime
 import java.time.format.DateTimeFormatter
-import java.time.format.FormatStyle
+import java.util.Locale
 
 fun LocalDateTime.toIsoDateTime(): String = this.format(DateTimeFormatter.ISO_DATE_TIME)
 
-fun LocalDateTime.toMediumFormatStyle(): String = this.format(DateTimeFormatter.ofLocalizedDateTime(FormatStyle.MEDIUM))
+fun LocalDateTime.toMediumFormatStyle(): String = this.format(DateTimeFormatter.ofPattern("d MMM yyyy, HH:mm:ss", Locale.ENGLISH))


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/AttendancesService.kt
```diff
@@ -84,10 +86,25 @@ class AttendancesService(
     log.info("Attendance marking done for ${markedAttendanceIds.size} attendance record(s)")
   }
 
-  private fun AttendanceUpdateRequest.mayBeCaseNote(prisonerNumber: String): CaseNote? =
+  private fun AttendanceUpdateRequest.mayBeCaseNote(attendance: Attendance): CaseNote? =
     caseNote?.let {
+      val caseNoteReason = if (attendance.issuePayment == true && issuePayment == false) "Pay removed" else if (maybeAttendanceReason() == AttendanceReasonEnum.REFUSED) "Refused to attend" else null
+      val activityName = attendance.scheduledInstance.activitySchedule.activity.summary
```
matt-crowson-moj, line 92: I'd prefer not to see chained method call quite as deep as this, bit of a smell ...

```
attendance.scheduledInstance.activitySchedule.activity.summary
attendance.scheduledInstance.activitySchedule.internalLocationDescription
```

Would suggest adding some functions to the instance (or even attendance) perhaps which under the covers delegates

```
attendance.scheduledInstance.summary()
attendance.scheduledInstance.internalLocationDescription()
```

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/AttendancesService.kt
```diff
@@ -84,10 +86,25 @@ class AttendancesService(
...
+      val location = attendance.scheduledInstance.activitySchedule.internalLocationDescription
+      val datetime = attendance.scheduledInstance.sessionDate.atTime(attendance.scheduledInstance.startTime).toMediumFormatStyle()
+      val prefix = if (caseNoteReason != null) {
```
matt-crowson-moj, line 95: let { ... } is your friend here perhaps

```
      val prefix = caseNoteReason?.let {
        listOfNotNull(
          caseNoteReason,
          activityName,
          location,
          datetime,
        ).joinToString(" - ")
      }

```

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/AttendancesService.kt
```diff
@@ -84,10 +86,25 @@ class AttendancesService(
     log.info("Attendance marking done for ${markedAttendanceIds.size} attendance record(s)")
   }
 
-  private fun AttendanceUpdateRequest.mayBeCaseNote(prisonerNumber: String): CaseNote? =
+  private fun AttendanceUpdateRequest.mayBeCaseNote(attendance: Attendance): CaseNote? =
```
matt-crowson-moj, line 89: Hmmm so looking at the AttendanceUpdateRequest object there doesn't appear to be any restriction on the size of the case note.

Note sure what the rules are around this now, lost track.  So I think the UI was restricting to 3800 is this still the case?

I guess firstly it probably should be enforcing some sort of limit anyway?  That aside should it be reduced to allow for the the prefix to be added?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/AttendancesServiceTest.kt
```diff
@@ -91,16 +92,40 @@ class AttendancesServiceTest {
   }
 
   @Test
-  fun `mark attendance record with case note`() {
+  fun `mark attendance record with case note - Refused to attend`() {
+    val expectedCaseNotePrefix = "Refused to attend - Maths - Education - R1 - ${LocalDate.now().atTime(attendance.scheduledInstance.startTime).toMediumFormatStyle()}"
```
matt-crowson-moj, line 96: Not essential but would nice to see at least one test showing medium style format is?  This just just uses what has been added as an extension.  How do we know what that is and it is correct?

matt-crowson-moj: Perhaps a simple assertion in the LocalDateExtTest class?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/common/LocalDateTimeExtTest.kt
```diff
@@ -0,0 +1,18 @@
...
+
+import org.junit.jupiter.api.Test
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.helpers.isEqualTo
+import java.time.LocalDateTime
+
+class LocalDateTimeExtTest {
+
+  @Test
+  fun `to ISO format`() {
+    LocalDateTime.of(2024, 1, 1, 10, 0).toIsoDateTime() isEqualTo "2024-01-01T10:00:00"
```
matt-crowson-moj, line 11: Thanks for adding this one.  I should have done this way back.

matt-crowson-moj (approved):
LGTM

PR #775, opened 2024-02-06
SAA-1533: Add ability to plan a suspension for an allocation

smcveigh941:
- Adds `planned_suspension` table
- Service code assumes that only 1 suspension can be planned at once. Cannot currently plan 2 suspensions on different dates. Although the many-to-one table mapping allows for this should we need to extend the functionality in future.
- When a suspension is set, the suspension is applied or removed immediately to attendances, should the new dates relate to TODAY
- If the suspension is set to begin/end in the future, the 3am job to manage allocations will suspend/unsuspend depending on the planned dates
- If a suspension is planned for the prisoner at the moment they are received back into prison from a temporary release, their allocation will be set to `SUSPENDED` from `AUTO_SUSPENDED`
- If a prisoner leaves prison temporarily during a suspension, their SUSPENDED allocation will be set to AUTO_SUSPENDED
- SUSPENDED allocations can be ENDED should they reach the allocation end date during the suspension

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/MigrateActivityService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/MigrateActivityService.kt
@@ -50,6 +50,7 @@ const val RISLEY_PRISON_CODE = "RSI"
 class MigrateActivityService(
   private val rolloutPrisonService: RolloutPrisonService,
   private val activityRepository: ActivityRepository,
+  private val prisonRegimeService: PrisonRegimeService,
   private val activityScheduleRepository: ActivityScheduleRepository,
   private val prisonerSearchApiClient: PrisonerSearchApiClient,
   private val incentivesApiClient: IncentivesApiClient,
@@ -151,8 +152,10 @@ class MigrateActivityService(
   fun buildSingleActivity(request: ActivityMigrateRequest): List<Activity> {
     log.info("Migrating activity ${request.description} on a 1-2-1 basis")
     val activity = buildActivityEntity(request)
+    val prisonRegime = prisonRegimeService.getPrisonTimeSlots(request.prisonCode)
     request.scheduleRules.consolidateMatchingScheduleSlots().forEach {
-      activity.schedules().first().addSlot(1, it.startTime to it.endTime, getRequestDaysOfWeek(it))
+      val regime = TimeSlot.slot(it.startTime).let { slot -> prisonRegime[slot]!! }
+      activity.schedules().first().addSlot(1, regime, getRequestDaysOfWeek(it))
     }
     return listOf(activity)
   }
@@ -172,14 +175,16 @@ class MigrateActivityService(
   fun genericSplitActivity(request: ActivityMigrateRequest): List<Activity> {
     // Build the first activity
     val activity1 = buildActivityEntity(request, true, 2, 1)
+    val prisonRegime = prisonRegimeService.getPrisonTimeSlots(request.prisonCode)
 
     // Add the morning sessions to week 1 and the afternoon sessions to week 2 (ignores evening slots!)
     request.scheduleRules.consolidateMatchingScheduleSlots().forEach {
+      val regime = TimeSlot.slot(it.startTime).let { slot -> prisonRegime[slot]!! }
       if (TimeSlot.slot(it.startTime) == TimeSlot.AM) {
-        activity1.schedules().first().addSlot(1, it.startTime to it.endTime, getRequestDaysOfWeek(it))
+        activity1.schedules().first().addSlot(1, regime, getRequestDaysOfWeek(it))
       }
       if (TimeSlot.slot(it.startTime) == TimeSlot.PM) {
-        activity1.schedules().first().addSlot(2, it.startTime to it.endTime, getRequestDaysOfWeek(it))
+        activity1.schedules().first().addSlot(2, regime, getRequestDaysOfWeek(it))
       }
     }
 
@@ -188,11 +193,12 @@ class MigrateActivityService(
 
     // Add the afternoon sessions to week 1 and the morning sessions to week 2
     request.scheduleRules.consolidateMatchingScheduleSlots().forEach {
+      val regime = TimeSlot.slot(it.startTime).let { slot -> prisonRegime[slot]!! }
       if (TimeSlot.slot(it.startTime) == TimeSlot.PM) {
-        activity2.schedules().first().addSlot(1, it.startTime to it.endTime, getRequestDaysOfWeek(it))
+        activity2.schedules().first().addSlot(1, regime, getRequestDaysOfWeek(it))
       }
       if (TimeSlot.slot(it.startTime) == TimeSlot.AM) {
-        activity2.schedules().first().addSlot(2, it.startTime to it.endTime, getRequestDaysOfWeek(it))
+        activity2.schedules().first().addSlot(2, regime, getRequestDaysOfWeek(it))
       }
     }
 
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/MigrateActivityServiceTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/MigrateActivityServiceTest.kt
@@ -66,6 +66,7 @@ class MigrateActivityServiceTest {
   private val featureSwitches: FeatureSwitches = mock()
   private val eventOrganiserRepository: EventOrganiserRepository = mock()
   private val outboundEventsService: OutboundEventsService = mock()
+  private val prisonRegimeService: PrisonRegimeService = mock()
 
   private val listOfCategories = listOf(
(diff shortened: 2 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/PlannedSuspension.kt
```diff
@@ -0,0 +1,55 @@
...
+  fun plannedReason() = plannedReason
+  fun plannedBy() = plannedBy
+
+  fun hasStarted() = plannedStartDate.onOrBefore(LocalDate.now())
+  fun endOn(date: LocalDate, byWhom: String) = run {
+    plannedEndDate = date
+    updatedBy = byWhom
+    updatedAt = LocalDateTime.now()
+  }
+  fun endNow() = run { plannedEndDate = LocalDate.now() }
```
matt-crowson-moj, line 46: `fun endNow() { plannedEndDate = LocalDate.now() }` ?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/PlannedSuspension.kt
```diff
@@ -0,0 +1,55 @@
...
+  @ManyToOne
+  @JoinColumn(name = "allocation_id", nullable = false)
+  val allocation: Allocation,
+
+  private var plannedStartDate: LocalDate,
+  private var plannedReason: String,
+  private var plannedBy: String,
+  private var plannedAt: LocalDateTime = LocalDateTime.now(),
+  private var updatedBy: String,
+  private var updatedAt: LocalDateTime = LocalDateTime.now(),
```
matt-crowson-moj, line 30: Seems odd to be setting/defaulting the updatedAt in the constructor.  It hasn't been updated?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/PlannedSuspension.kt
```diff
@@ -0,0 +1,55 @@
...
+  private var plannedEndDate: LocalDate? = null,
+) {
+
+  fun allocation() = allocation
```
matt-crowson-moj, line 34: Do you need this?  `allocation` is public?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/PlannedSuspension.kt
```diff
@@ -0,0 +1,55 @@
...
+  fun plannedBy() = plannedBy
+
+  fun hasStarted() = plannedStartDate.onOrBefore(LocalDate.now())
+  fun endOn(date: LocalDate, byWhom: String) = run {
+    plannedEndDate = date
+    updatedBy = byWhom
+    updatedAt = LocalDateTime.now()
+  }
+  fun endNow() = run { plannedEndDate = LocalDate.now() }
+  fun plan(reason: String, startDate: LocalDate, byWhom: String) = run {
```
matt-crowson-moj, line 47: Just curious as I tend to use `apply { ... }` over run `{ ... }`.

You can achieve the same with both but I think `apply` is more commonly used I think for mutation of `this`, `apply { ... }` would also return a reference to the object itself it could be chained into other things.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/ScheduledInstance.kt
```diff
@@ -145,6 +145,9 @@ data class ScheduledInstance(
   fun internalLocationDescription() = activitySchedule.internalLocationDescription
 
   fun dateTime() = sessionDate.atTime(startTime)
+
+  fun isFuture(dateTime: LocalDateTime) = sessionDate == dateTime.toLocalDate() &&
```
matt-crowson-moj, line 149: Not sure but is this what you are trying to say here?

`fun inFuture(dateTime: LocalDateTime) = sessionDate.atTime(startTime).isAfter(dateTime)`

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ManageAllocationsService.kt
```diff
@@ -140,6 +152,16 @@ class ManageAllocationsService(
   private fun List<ActivitySchedule>.allocationsDueToStartOnOrBefore(date: LocalDate) =
     flatMap { it.allocations().filter { allocation -> allocation.startDate <= date } }
 
+  private fun allocationsDueToBeSuspended(): List<Allocation> =
+    forEachRolledOutPrison()
```
matt-crowson-moj, line 156: So there is a bit of a push to get the service to do one prison at a time and let the job that invokes the service to call it with a prison code.  I think if any of the suspensions fails here it will fail for all prisons?

One of the thoughts being it opens up the possibly (in future) to make the jobs a little more flexible in that perhaps if we need to rerun it for one prison then it would potentially be easier.

It also means the transactions should be smaller.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Allocation.kt
```diff
@@ -120,6 +124,16 @@ data class Allocation(
   var suspendedReason: String? = null
     private set
 
+  fun plannedSuspension() = plannedSuspensions.singleOrNull { it.endDate()?.isAfter(LocalDate.now()) ?: true }
+
+  fun isCurrentlySuspended() = plannedSuspension()?.hasStarted() == true
+
+  fun addPlannedSuspension(suspension: PlannedSuspension) = run {
+    require(suspension.allocation == this) { "Cannot add this suspension associated to the allocation with id ${suspension.allocation.allocationId} to allocation with id $allocationId" }
+    require(plannedSuspension() == null) { "Cannot add this planned suspension to allocation with id $allocationId because another planned suspension is already active" }
```
matt-crowson-moj, line 133: `... another planned suspension is already active` -> `... another planned suspension already exists`?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/AllocationsService.kt
```diff
@@ -169,6 +176,56 @@ class AllocationsService(
...
+      require(allocation.plannedEndDate() == null || suspendUntil.onOrBefore(allocation.plannedEndDate()!!)) {
+        "Suspension end date must be on or before the allocation end date: ${allocation.plannedEndDate()!!.toIsoDate()}"
+      }
+
+      plannedSuspension.endOn(suspendUntil, byWhom)
+    }
+
+    allocation.takeIf { it.status(PrisonerStatus.ACTIVE) && it.isCurrentlySuspended() }?.let {
+      it.activatePlannedSuspension()
+      attendanceSuspensionService.suspendFutureAttendancesForAllocation(LocalDateTime.now(), it)
```
matt-crowson-moj, line 221: Hmm I am actually not sure about this.  This will commit changes to attendance records and publish?  However we haven't saved the allocation yet.  What if it fails to save the allocation?

src/main/resources/migrations/common/V2024.02.16.1__add_planned_suspension.sql
```diff
@@ -0,0 +1,14 @@
+-- ==================================================================
+-- Create the new table
+-- ==================================================================
+CREATE TABLE planned_suspension (
```
matt-crowson-moj, line 4: Just something to be aware of we have a few migrations done recently and incoming recently so good to double check before merging.

src/main/resources/migrations/common/V2024.02.16.1__add_planned_suspension.sql
```diff
@@ -0,0 +1,14 @@
...
+  planned_suspension_id   bigserial    NOT NULL CONSTRAINT planned_suspension_pk PRIMARY KEY,
+  allocation_id           bigint       NOT NULL REFERENCES allocation (allocation_id),
+  planned_reason          varchar(100) NOT NULL,
+  planned_start_date      date         NOT NULL,
+  planned_end_date        date,
+  planned_by              varchar(100) NOT NULL,
+  planned_at              timestamp    NOT NULL,
+  updated_by              varchar(100) NOT NULL,
```
matt-crowson-moj, line 12: Feels to me like the updated columns should be nullable?  Kinda relates to another question ...

src/test/resources/application-test.yml
```diff
@@ -13,6 +13,7 @@ spring:
     show-sql: true
     properties:
       hibernate:
+        enable_lazy_load_no_trans: true
```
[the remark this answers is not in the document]
matt-crowson-moj: Is this something we can apply on a need only basis.

Just worried we might allow this when we perhaps shouldn't?

davidthomas22 (approved):
LGTM

PR #776, opened 2024-02-08
SAA-1536 remove unused deprecated endpoint GET /activities/{ID}, this is no longer used.

matt-crowson-moj:
This removes the unused deprecated endpoint `GET /activities/{ID}`.

All calls that require to get the activity by its ID should go through the filtered endpoint `GET /activities/{ID}/filtered`.

I have checked Application Insights and cannot see this endpoint used anymore.

Goes with this PR [here](https://github.com/ministryofjustice/hmpps-activities-management/pull/748).


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityService.kt
```diff
@@ -79,18 +80,13 @@ class ActivityService(
...
     val activity = activityRepository.getActivityByIdWithFilters(activityId, earliestSession)
-      ?: throw (EntityNotFoundException("Activity $activityId not found"))
-    return transform(activity)
-  }
+      ?: throw EntityNotFoundException("Activity $activityId not found")
 
-  fun getActivityById(activityId: Long): ModelActivity {
-    val activity = activityRepository.findById(activityId)
-      .orElseThrow { EntityNotFoundException("Activity $activityId not found") }
     checkCaseloadAccess(activity.prisonCode)
```
matt-crowson-moj, line 88: Also now doing case load check.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityService.kt
```diff
@@ -79,18 +80,13 @@ class ActivityService(
...
```
matt-crowson-moj, line 88: Also now doing case load check.

PR #779, opened 2024-02-09
SAA-1550 specifiying the Java version for the OWASP task to see if the fixes it. It looks like it is using the wrong version looking at Circle CI.

matt-crowson-moj:
Not 100% sure this will fix it.  Either way it should not be using JDK 17, it should be 21.

image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,12 +11,15 @@ parameters:
   releases-slack-channel:
     type: string
     default: activities-and-appointments-alerts
+  java-version:
+    type: string
+    default: "21.0"
 
 jobs:
   validate:
     executor:
       name: hmpps/java
-      tag: "21.0"
+      tag: << pipeline.parameters.java-version >>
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -137,6 +140,7 @@ workflows:
           context:
             - hmpps-common-vars
           cache_key: "v3_0"
+          jdk_tag: << pipeline.parameters.java-version >>
```
matt-crowson-moj, line 143: This is new for the OWASP job here.

.circleci/config.yml
```diff
@@ -11,12 +11,15 @@ parameters:
   releases-slack-channel:
     type: string
     default: activities-and-appointments-alerts
+  java-version:
+    type: string
+    default: "21.0"
```
matt-crowson-moj, line 16: Pulled out JDK version into a param so only defined in one place.

PR #784, opened 2024-02-14
SAA-1555 update dependencies to resolve OWASP security job failure in the CI pipeline.

matt-crowson-moj:
One of the tests had to be updated as a result of bumping the SQS library, it is not cause for concern.  I have commented on the test impacted.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -4,12 +4,12 @@ import org.jlleitschuh.gradle.ktlint.tasks.KtLintFormatTask
 import org.openapitools.generator.gradle.plugin.tasks.GenerateTask
 
 plugins {
-  id("uk.gov.justice.hmpps.gradle-spring-boot") version "5.15.1"
+  id("uk.gov.justice.hmpps.gradle-spring-boot") version "5.15.2"
   kotlin("plugin.spring") version "1.9.22"
   kotlin("plugin.jpa") version "1.9.22"
   jacoco
-  id("org.openapi.generator") version "7.2.0"
-  id("io.sentry.jvm.gradle") version "4.2.0"
+  id("org.openapi.generator") version "7.3.0"
+  id("io.sentry.jvm.gradle") version "4.3.0"
 }
 
 allOpen {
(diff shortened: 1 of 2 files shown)


build.gradle.kts
```diff
@@ -31,7 +31,7 @@ dependencies {
   implementation("org.springframework.boot:spring-boot-starter-oauth2-resource-server")
   implementation("org.springframework.boot:spring-boot-starter-webflux")
   implementation("org.springframework.boot:spring-boot-starter-data-jpa")
-  implementation("uk.gov.justice.service.hmpps:hmpps-sqs-spring-boot-starter:2.2.1")
+  implementation("uk.gov.justice.service.hmpps:hmpps-sqs-spring-boot-starter:3.1.1")
```
matt-crowson-moj, line 34: Note this is a major version bump.  Checking the release notes and chatting with the author of this version no further changes are required.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/resource/EndpointSecurityCheck.kt
```diff
@@ -4,48 +4,50 @@ import org.junit.jupiter.api.Test
...
     if (controllers.isNotEmpty()) {
       fail("Role checks missing in following locations: ${controllers.joinToString("\n")}\n")
     }
   }
 
   private fun getAllUnprotectedControllers() = ClassPathScanningCandidateComponentProvider(false)
-    .also { it.addIncludeFilter(AnnotationTypeFilter(RestController::class.java)) }
-    .findCandidateComponents("uk.gov.justice")
+    .apply { addIncludeFilter(AnnotationTypeFilter(RestController::class.java)) }
+    .findCandidateComponents("uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi")
```
matt-crowson-moj, line 41: Extended this to only include endpoints we are in control of.  This started to fail as it was picking a controller from the SQS library (and failing) which we have no control of.

Note this should not cause concern, it just means the SQS one does not annotate the controllers with the same rules.  The SQS endpoint is still protected by the ingress.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/resource/EndpointSecurityCheck.kt
```diff
@@ -4,48 +4,50 @@ import org.junit.jupiter.api.Test
...
+  @Test
+  fun `Ensure checks are working by referencing fake unprotected controller`() {
+    getAllUnprotectedControllers().map(ControllerInfo::controller) containsExactly listOf("class ${FakeUnprotectedController::class.qualifiedName}")
   }
 
   @Test
   fun `Ensure endpoints are checking roles`() {
-    val controllers = getAllUnprotectedControllers()
+    val controllers =
+      getAllUnprotectedControllers().filterNot { it.controller == "class ${FakeUnprotectedController::class.qualifiedName}" }
```
matt-crowson-moj, line 32: Excluding the fake controller here as would otherwise fail.

PR #785, opened 2024-02-15
SAA-1556 this change now updates booking ids on offender merge events, they were not included as part of the original offender merge work.

matt-crowson-moj:
Whilst it was understood that booking ID's should generally be safe from change what was missed was the fact before the merge takes place a prisoner can be allocated with their new prisoner number before the merge has actually happened.  This means the prisoner can both allocated with their old and new prisoner numbers but the one with the new prisoner number is using a different booking id (because it hasn't been merged to the old one).

**I am not 100% sure but I think there is the potential for duplicate allocations to slip through on the back of a merge event because we can have both the old and new prisoner numbers allocated prior to a merge happening.  The chance is small but it could happen.**

It would be prudent to check for offender more merges in production happening before this is released as the migration may need additions.


src/main/resources/migrations/prod/V2024.02.17__fix_offender_merge_booking_ids.sql
```diff
@@ -0,0 +1,6 @@
+-- This fixes the data where we missed updating booking ID's on the back of an offender merge.
+update allocation set booking_id = 2866250 where prisoner_number = 'A7158AF' and booking_id != 2866250;
```
matt-crowson-moj, line 2: We could do this direct but this keeps an audit trail.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/repository/AllocationRepository.kt
```diff
@@ -100,7 +100,11 @@ interface AllocationRepository : JpaRepository<Allocation, Long> {
     startDate: LocalDate,
   ): List<Allocation>
 
-  @Query(value = "UPDATE Allocation a SET a.prisonerNumber = :newNumber WHERE a.prisonerNumber = :oldNumber")
+  @Query(value = "UPDATE Allocation a SET a.prisonerNumber = :newNumber, a.bookingId = coalesce(:newBookingId, a.bookingId) WHERE a.prisonerNumber = :oldNumber")
   @Modifying
```
timharrison-moj, line 104: Is :newBooking ever null?

matt-crowson-moj: It's a good point and possibly not, unfortunately this is mainly driven by the prison-api model saying it can be null.

PR #786, opened 2024-02-16
SAA-1561 removing index originally introduced to help prevent duplication submission of allocations as this is impacting deallocations.

matt-crowson-moj:
This index was added to prevent duplicate allocations to the same activity, however this has had a knock on effect with deallocation

We have since found that an offender can be allocated to the same activity on the same date provided the earlier one is already ended, however if you then try to end the later allocation a duplicate key violation is incurred.

This is causing issues the deallocate expiring allocations job.


timharrison-moj (dismissed):
LGTM

PR #787, opened 2024-02-16
Task/saa 1528 subject access request appointments search. Return an array of appointments in the response

dafhoustonmoj:
Retrieve appointments for a prisoner between 2 dates. Retrieves all appointments based on the appointment created time. Also shows the appointment details and the attendance for a given appointment


src/main/resources/migrations/common/V2024.02.15__add_allocation_date_subject_access_request_appointments_view.sql
```diff
@@ -0,0 +1,22 @@
+-- =================================================================================================
```
matt-crowson-moj, line 1: Check the state of play with the latest migration numbers.  This version number will need bumping.

I also have a PR out [here](https://github.com/ministryofjustice/hmpps-activities-management-api/pull/785) for review with a migration in it so we need to make sure they don't clash.

dafhoustonmoj: will do. that filename is incorrect as well

matt-crowson-moj (approved):
LGTM

PR #788, opened 2024-02-16
SAA-1510: Add time window to deallocate by ending job

munslowl:
- Add time window to deallocate ending job to cover that failed for previous runs within the window, e.g. 3 days.
- Remove use of session filter


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ManageAllocationsService.kt
```diff
@@ -66,7 +66,7 @@ class ManageAllocationsService(
     }
 
     transactionHandler.newSpringTransaction {
-      activityScheduleRepository.getActivitySchedulesWithFilteredInstances(prisonCode, date).flatMap { schedule ->
+      activityScheduleRepository.findAllByActivityPrisonCode(prisonCode).flatMap { schedule ->
```
munslowl, line 69: Reverting a change that used a session date filter that was unnecessary.

PR #789, opened 2024-02-17
SAA-1562: Activity slot times are calculated from the given prison regime during migration

smcveigh941:
Activity start and end times should fit within the regime settings we specify in the `prison_regime` table. Activities imported during migration may have different times to these, so we need to use the prison regime we have been given to calculate the AM, PM and ED slot times for the activity should it have a different start/end time to the regime.


davidthomas22 (approved):
LGTM

PR #794, opened 2024-02-26
SAA-1578 adds the missing root node 'content' to the SAR response object.

matt-crowson-moj:
The SAR response object was not conforming to the API specification and was missing the root node 'content'.

All SAR data should be in the root node as follows below:

```
{
  content: { .. data goes here ... }
}
```

image


dafhoustonmoj (approved):
LGTM

PR #795, opened 2024-02-27
SAA-1128: Update duplicate activity summary validation

davidthomas22:
*Update duplicate activity summary validation to check **active** activity names and disable this option
*Drop the constraint for the activity **summary** field to enable the function to fully handle a duplicate name request
*Move the h2 integration test migration dependency into the test folder to ensure that it only runs in the required environments


src/main/resources/migrations/common/V2024.02.21__drop_activities_unique_constraint.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE activity DROP CONSTRAINT IF EXISTS activity_prison_code_summary_key;
```
matt-crowson-moj, line 1: The number (date) of the migration will need bumping.  We are now at V2024.02.26.

src/test/resources/migrations/h2/V2024.02.26__drop_activity_summary_constraint.sql
```diff
@@ -0,0 +1,3 @@
+-- Compatibility with integration tests using H2 database
+ALTER TABLE activity DROP CONSTRAINT IF EXISTS CONSTRAINT_CBF1;
```
matt-crowson-moj, line 2: Do we really need this?  Not saying you are wrong but seems a little strange given we haven't had to do this anywhere else.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/repository/ActivityRepository.kt
```diff
@@ -35,6 +35,15 @@ interface ActivityRepository : JpaRepository<Activity, Long>, ActivityRepository
...
   fun findByActivityIdAndPrisonCode(activityId: Long, prisonCode: String): Activity?
 
+  @Query(
+    value =
+    "select case when count(a) > 0 then true else false end " +
+      "from Activity a " + "where a.prisonCode = :prisonCode " +
+      "and a.summary = :summary " +
+      "and (a.endDate is null or a.endDate > :endDate)",
+  )
+  fun findExistingLiveActivity(prisonCode: String, summary: String, endDate: LocalDate): Boolean
```
matt-crowson-moj, line 45: More a naming thing (I can understand) but `find` implies it returns an Activity. `exists` is generally used for boolean return types so `existsLiveActivity` or just `existingLiveActivity` perhaps?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/integration/ActivityIntegrationTest.kt
```diff
@@ -301,6 +301,23 @@ class ActivityIntegrationTest : IntegrationTestBase() {
...
+      "prisonapi/location-id-1.json",
+    )
+
+    val newActivityWithDuplicateSummary = activityCreateRequest(
+      prisonCode = MOORLAND_PRISON_CODE,
+      educationLevel = prisonApiMockServer.stubGetReferenceCode("EDU_LEVEL", "1", "prisonapi/education-level-code-1.json"),
+      studyArea = prisonApiMockServer.stubGetReferenceCode("STUDY_AREA", "ENGLA", "prisonapi/study-area-code-ENGLA.json"),
+    ).copy(summary = "Maths")
+
+    val createdActivity = webTestClient.createActivity(newActivityWithDuplicateSummary)
```
matt-crowson-moj, line 318: Would perhaps be nice to see some assertions here?

src/test/resources/migrations/h2/V2024.02.26__drop_activity_summary_constraint.sql
munslowl: Once before - [here](https://github.com/ministryofjustice/hmpps-activities-management-api/blob/main/src/main/resources/migrations/common/V2023.06.22__bulk_appointment_add_columns.sql#L31).

matt-crowson-moj: In which case I would be inclined to copy that approach for now perhaps instead of adding to the test folder.

I think going forwards we should move away from H2 and use TestContainers (not as part of this ticket).  There are other DPS services we can look at for examples use Postgres TestContainers.

matt-crowson-moj (dismissed):
LGTM

munslowl (approved):

PR #796, opened 2024-02-27
SAA-1559: Classify in cell appointments as on wing

munslowl:
- Classify in cell appointments as on wing
- Classify internal location as off wing 
- Ensure appointment creation to handle In Cell
- Ensure appointment Set creation to handle In Cell
- Ensure appointment Series creation to handle In Cell


matt-crowson-moj (approved):
LGTM

PR #798, opened 2024-02-28
SAA-1575: Update dependencies to clear CVEs

munslowl:
- Clear 3 CVEs
- Resolve subsequent issues with child entities following `saveAndFlush`


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/AppointmentUpdateDomainService.kt
```diff
@@ -67,6 +71,12 @@ class AppointmentUpdateDomainService(
 
       val updatedAppointmentSeries = appointmentSeriesRepository.saveAndFlush(appointmentSeries)
 
+      val addedAttendees = updatedAppointmentSeries.appointments()
```
munslowl, line 74: Since `> hibernate 6.4.2.Final` `Attendee` entities in `Appointment` are not same objects but a collection of new ones which means you can't look at the old entities to get the `id` which means `OutboundEvent.APPOINTMENT_INSTANCE_CREATED` is fired with `0` id.

I can pin the hibernate version id this upsets.

matt-crowson-moj (approved):
LGTM

matt-crowson-moj (approved):
LGTM

PR #799, opened 2024-02-29
SAA-1572: add earliest release date to waiting list applications

dafhoustonmoj:
The /schedules/{scheduleId}/waiting-list-applications needs to include the earliest release date as this is required to be shown on the Waitlist on Activity Allocation screen

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/model/WaitingListApplication.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/model/WaitingListApplication.kt
@@ -3,6 +3,7 @@ package uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.model
 import com.fasterxml.jackson.annotation.JsonFormat
 import io.swagger.v3.oas.annotations.media.Schema
 import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.entity.WaitingListStatus
+import uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.model.response.EarliestReleaseDate
 import java.time.LocalDate
 import java.time.LocalDateTime
 
@@ -114,4 +115,7 @@ data class WaitingListApplication(
     example = "Jane Doe",
   )
   val updatedBy: String? = null,
+
+  @Schema(description = "The earliest release date of the prisoner")
+  var earliestReleaseDate: EarliestReleaseDate? = null,
 )
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/model/WaitingListApplication.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/model/WaitingListApplication.kt
@@ -117,5 +117,5 @@ data class WaitingListApplication(
   val updatedBy: String? = null,
 
   @Schema(description = "The earliest release date of the prisoner")
-  var earliestReleaseDate: EarliestReleaseDate? = null,
+  val earliestReleaseDate: EarliestReleaseDate,
 )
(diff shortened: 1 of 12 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/WaitingListService.kt
```diff
@@ -56,7 +57,17 @@ class WaitingListService(
...
       .findOrThrowNotFound(id)
       .checkCaseloadAccess()
-      .let(waitingListRepository::findByActivitySchedule).map(WaitingList::toModel)
+      .let(waitingListRepository::findByActivitySchedule).map(WaitingList::toModel).apply {
+        val prisoners =
+          prisonerSearchApiClient.findByPrisonerNumbers(map { it.prisonerNumber })
+
+        map {
+          val prisoner = prisoners.find { p -> it.prisonerNumber == p.prisonerNumber }
+          if (prisoner != null) {
```
matt-crowson-moj, line 66: Hmm `prisoner == null` should never happen.

If it does then we should probably fail fast here, something is wrong with our data.

Doing this though obviously has ramifications as retrieval of waiting lists will fail also.

dafhoustonmoj: Does that mean the test I edited below (WaitingListServiceTest ) the line below could never happen? I should always return a value? If I added a return object it wouldn't need a null check. 
 
prisonerSearchApiClient.stub {
      on { findByPrisonerNumbers(listOf("123456")) } doReturn emptyList()
    }

The earliestReleaseDate is optional in the schema

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/model/WaitingListApplication.kt
```diff
@@ -114,4 +115,7 @@ data class WaitingListApplication(
     example = "Jane Doe",
   )
   val updatedBy: String? = null,
+
+  @Schema(description = "The earliest release date of the prisoner")
+  var earliestReleaseDate: EarliestReleaseDate? = null,
```
matt-crowson-moj, line 120: Will this ever be null?  It looks like it always returns something in the service ...

dafhoustonmoj: updated to be mandatory

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/WaitingListService.kt
```diff
@@ -49,14 +50,22 @@ class WaitingListService(
...
-      .let(waitingListRepository::findByActivitySchedule).map(WaitingList::toModel)
+
+    val prisoner =
+      prisonerSearchApiClient.findByPrisonerNumber(waitingList.prisonerNumber) ?: throw NullPointerException("Prisoner ${waitingList.prisonerNumber} not found for waiting list id $id")
+    return waitingList.toModel(determineEarliestReleaseDate(prisoner))
+  }
+
+  fun getWaitingListsBySchedule(id: Long) =
+    scheduleRepository.findOrThrowNotFound(id).checkCaseloadAccess().let(waitingListRepository::findByActivitySchedule).map {
+      val prisoner = prisonerSearchApiClient.findByPrisonerNumber(it.prisonerNumber) ?: throw NullPointerException("Prisoner ${it.prisonerNumber} not found for waiting list id $id")
```
matt-crowson-moj, line 65: Sorry Daf.  I don't think this one is quite right.  Have you changed this to how it was in your previous commit?

This is going to make a prisoner-search request for each waiting list on the schedule.  So e.g. 100 waiting lists on a schedule then 100 external API requests.

So before I think you had this:

```
fun getWaitingListsBySchedule(id: Long) =
    scheduleRepository
      .findOrThrowNotFound(id)
      .checkCaseloadAccess()
      .let(waitingListRepository::findByActivitySchedule).map(WaitingList::toModel).apply {
        val prisoners =
          prisonerSearchApiClient.findByPrisonerNumbers(map { it.prisonerNumber })

        map {
          val prisoner = prisoners.find { p -> it.prisonerNumber == p.prisonerNumber } ?: throw NullPointerException("Prisoner ${it.prisonerNumber} not found for waiting list id $id")
          it.earliestReleaseDate = determineEarliestReleaseDate(prisoner)
        }
      }
```

matt-crowson-moj (approved):
LGTM - I presume this has a dependent UI PR which will need to be merged at the same time as this one.

PR #800, opened 2024-02-29
SAA-1576: Return usernames for createdBy and updatedBy fields on appointments

smcveigh941:
Some `createdBy` and `updatedBy` fields returned a `UserSummary` object which was populated by a call to prison API to fetch the user details. There are a few problems with this:

- Manage Users API should be used instead of Prison API
- Extra calls to external APIs are being made when they are not necessarily needed
- Response models are inconsistent - some return the UserSummary and some return the username only

I have removed this call to Prison API, and all responses where relevant only return the username. Where necessary, an accompanying UI change will use the username to search Manage Users API for a display name for display purposes.


matt-crowson-moj (approved):
LGTM

munslowl (approved):

PR #807, opened 2024-03-07
SAA-1619 logic for canAttendOn(...) on Allocation was not factoring the planned allocation end date.

matt-crowson-moj:
This bug was spotted after the deallocation job failed and the attendance job ran successfully (in DEV)

An allocation that should have been ended due to having a planned end date did not end but an attendance record was still created when it shouldn't have been.  This was because it did not factor in the allocations planned end date.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Allocation.kt
```diff
@@ -192,6 +192,7 @@ data class Allocation(
   private fun maybeEndDate() =
```
matt-crowson-moj, line 192: This function is referenced by the canAttendOn( ... ) but the planned end date was not being factored in.

PR #810, opened 2024-03-07
SAA-1617 readme tidy up.

matt-crowson-moj:


docker-compose.yml
```diff
@@ -15,18 +15,5 @@ services:
       DPR_USER: dpr_user
       DPR_PASSWORD: dpr_password
 
-  prison-api:
```
matt-crowson-moj, line 18: never used so removing

util_scripts/migration/bulk-prisoner-allocate.sh
matt-crowson-moj, line 1: Same. No longer needed.  The was the original idea for migration of prisoners via a script.

docker-compose.yml
matt-crowson-moj, line 18: Removed.  We never really use it like this due to data needs.

util_scripts/migration/README.md
matt-crowson-moj, line 1: Removed.  This has been superseded with the migration service.

util_scripts/migration/bulk-prisoner-allocate.sh
matt-crowson-moj, line 1: Removed.  This has been superseded with the migration service.

PR #813, opened 2024-03-08
SAA-1621 record job start in DB at outset not just at end incase it fails to get that far e.g. out of memory error.

matt-crowson-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/job/SafeJobRunner.kt
```diff
@@ -30,39 +28,12 @@ class SafeJobRunner(
     runSafe(jobDefinition, true)
   }
 
-  /**
-   * This will run the dependent jobs in the order they are supplied.
-   *
-   * If a job in the sequence fails then the job(s) that follow (if there are any) will not run, this is intentional.
-   */
-  fun runDependentJobs(vararg jobDefinitions: JobDefinition) {
```
matt-crowson-moj, line 38: Removed as not used anywhere other than unit tests.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/job/SafeJobRunner.kt
```diff
@@ -30,39 +28,12 @@ class SafeJobRunner(
...
-      }
-      log.info("JOB: Elapsed time for job ${job.jobType} ${elapsed}ms")
-    }
-  }
-
   private fun runSafe(jobDefinition: JobDefinition, withRetry: Boolean = false): Result<Unit> {
     val startedAt = LocalDateTime.now()
 
-    log.info("JOB: Running job ${jobDefinition.jobType}")
+    val job = jobRepository.saveAndFlush(Job(jobType = jobDefinition.jobType, startedAt = startedAt))
```
matt-crowson-moj, line 34: Key change is here.  We record the start of the job in the DB prior to anything else.

This should mean if it fails e.g. due to out of memory error, at least we have a record of the fact it started.

munslowl (approved):

PR #815, opened 2024-03-12
SAA-1626 adding Sentry monitoring for caught exceptions in jobs so don't go unnoticed.

matt-crowson-moj:
Exceptions raised in some of the jobs are being missed due to the nature of `try ... catch` (and not propagating).

To be clear the `try ... catch` was always intentional e.g. if attendance creation fails for one prison doesn't mean it should fail for all.

In these scenarios the jobs can complete as successful even if there were some errors during the run.  If that does happen this change should at least provide some better visibility by essentially creating a Sentry alert which will propagate to the team alerts channel.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ManageAllocationsService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ManageAllocationsService.kt
@@ -246,7 +246,7 @@ class ManageAllocationsService(
             }?.map(Allocation::allocationId) ?: emptyList()
           }.let(::sendAllocationsAmendedEvents)
         },
-        failure = "An error occurred deallocating allocations on activity schedule ${schedule.activityScheduleId}.",
+        failure = "An error occurred deallocating allocations on activity schedule ${schedule.activityScheduleId}",
       )
     }
   }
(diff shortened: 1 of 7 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ManageAllocationsService.kt
```diff
@@ -280,7 +278,7 @@ class ManageAllocationsService(
           }.map(Allocation::allocationId)
         }.let(::sendAllocationsAmendedEvents)
       },
-      failure = "An error occurred while suspending allocations due to be suspended today",
+      failure = "An error occurred while unsuspending allocations due to be unsuspended today",
```
matt-crowson-moj, line 281: I noticed this failure message was wrong on my travels.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/MonitoringService.kt
```diff
@@ -21,10 +21,14 @@ class MonitoringService {
     log.info("Monitoring service is ${if (Sentry.isEnabled()) "enabled" else "disabled"}")
   }
 
-  fun capture(message: String) {
+  fun capture(message: String, error: Throwable? = null) {
     // This checks for the presence of the Sentry environment variable SENTRY_DSN, it is disabled if not found.
     if (Sentry.isEnabled()) {
```
matt-crowson-moj, line 26: Due to the nature of Sentry being accessed as a static object unit testing this is not really viable.

I will look at this independently, the service is very small so I am comfortable with this change.

munslowl (approved):

PR #816, opened 2024-03-13
SAA-1627 add retry to allocations jobs.

matt-crowson-moj:
The retry has already been applied to attendance creation and is out in production.

This change adds it to the allocation/deallocation jobs.

### DIFF
--- a/helm_deploy/values-dev.yaml
+++ b/helm_deploy/values-dev.yaml
@@ -38,7 +38,7 @@ generic-prometheus-alerts:
 # We are overriding the times in development due to core service dependencies being switched off overnight e.g. prisoner search API.
 cron:
   allocationsDeallocateEnding: "30 6 * * *"
-  allocationsDeallocateExpiring: "30 21 * * *"
+  allocationsDeallocateExpiring: "30 20 * * *"
   allocationsActivate: "45 6 * * *"
   appointmentAttendees: "45 6 * * *"
   createScheduledInstances: "0 7 * * *"
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/client/casenotesapi/api/CaseNotesApiClient.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/client/casenotesapi/api/CaseNotesApiClient.kt
@@ -19,6 +19,7 @@ class CaseNotesApiClient(@Qualifier("caseNotesApiWebClient") private val webClie
       .block()!!
   }
 
+  @Deprecated(message = "To keep things simpler (and take some pressure off the API) would prefer not to use this and return case note identifiers")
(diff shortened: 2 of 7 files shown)


helm_deploy/values-dev.yaml
```diff
@@ -38,7 +38,7 @@ generic-prometheus-alerts:
 # We are overriding the times in development due to core service dependencies being switched off overnight e.g. prisoner search API.
 cron:
   allocationsDeallocateEnding: "30 6 * * *"
-  allocationsDeallocateExpiring: "30 21 * * *"
+  allocationsDeallocateExpiring: "30 20 * * *"
```
matt-crowson-moj, line 41: I have brought this forwards an hour in DEV.

We are starting to see more errors around this time of evening so would like to if making a bit earlier calms things down.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/client/casenotesapi/api/CaseNotesApiClient.kt
```diff
@@ -19,6 +19,7 @@ class CaseNotesApiClient(@Qualifier("caseNotesApiWebClient") private val webClie
       .block()!!
   }
 
+  @Deprecated(message = "To keep things simpler (and take some pressure off the API) would prefer not to use this and return case note identifiers")
```
matt-crowson-moj, line 22: Unrelated but documenting here for future.  Going forwards it keeps things simpler if we don't have keep calling case notes API for the actual notes.

It will be up to calling services to worry about it.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/job/SafeJobRunner.kt
```diff
@@ -28,27 +28,29 @@ class SafeJobRunner(
...
-      } else {
-        jobDefinition.block()
-      }
-    }
-      .onSuccess { jobRepository.saveAndFlush(job.succeeded()) }
-      .onFailure {
-        log.error("JOB: Failed to run job ${jobDefinition.jobType} with job id '${job.jobId}'", it)
-        jobRepository.saveAndFlush(job.failed()).also { failedJob ->
-          monitoringService.capture("Job '${failedJob.jobType}' with job id '${failedJob.jobId}' failed")
+    val elapsed = measureTimeMillis {
```
matt-crowson-moj, line 36: Main change here is just to log how long the jobs takes.  Technically we do record this in the DB also but this provides something searchable in our logs.

helm_deploy/values-dev.yaml
matt-crowson-moj: For those that don't know - after 21:45 some of the services we rely on are turned off overnight (in DEV only).

dafhoustonmoj (approved):
TGTM

PR #817, opened 2024-03-14
SAA-1628 this change replaces the first of two prisoner domain events. The events we currently use are delayed by 45 mins, the new ones are not delayed.

matt-crowson-moj:


docker-compose-localstack.yml
```diff
@@ -37,7 +37,7 @@ services:
       - "/var/run/docker.sock:/var/run/docker.sock"
 
   prison-api:
-    image: rodolpheche/wiremock:latest
```
matt-crowson-moj, line 40: Replacing with a container that supports Arm architecture.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/events/handlers/PrisonerReceivedEventHandler.kt
```diff
@@ -27,21 +27,21 @@ import java.time.LocalDateTime
  */
 @Component
 @Transactional
-class OffenderReceivedEventHandler(
+class PrisonerReceivedEventHandler(
```
matt-crowson-moj, line 30: This is simply a rename to work in conjunction with the replacement event going forwards.  There are no functional changes.

PR #821, opened 2024-03-18
SAA-1128: Amend error content

davidthomas22:
Updating error content as per team review session amendments


matt-crowson-moj (approved):
LGTM

PR #822, opened 2024-03-18
SAA-1629 correcting the prisoner alert payload, personReference is not part of the payload, nomsNumber is though.

matt-crowson-moj:
End to end testing in dev confirms this alert was never tested, the structure of the message we were expecting was not correct.

This change should clean up a message we have sitting on our DLQ in dev.

Note preprod and prod are not affected as we are not listening for this event in those envs.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/events/InboundEvents.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/events/InboundEvents.kt
@@ -181,12 +181,8 @@ enum class Action {
 
 // ------------ New event contents here -------------------------------------------------------------
 
-data class AlertsUpdatedEvent(
-  val personReference: PersonReference,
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/events/InboundEvents.kt
```diff
@@ -181,12 +181,8 @@ enum class Action {
 
 // ------------ New event contents here -------------------------------------------------------------
 
-data class AlertsUpdatedEvent(
-  val personReference: PersonReference,
-  val additionalInformation: AlertsUpdatedInformation,
-) : InboundEvent, EventOfInterest {
-  override fun prisonerNumber(): String =
-    personReference.identifiers.first { it.type == "NOMS" }.value
```
matt-crowson-moj, line 189: Person ref is not part of the message payload.

munslowl (approved):

PR #834, opened 2024-03-26
SAA-1582: Migrate activity location user description rather than the location code

smcveigh941:
Fixes the following: 

image


dafhoustonmoj (approved):
LGTM

PR #837, opened 2024-03-27
SAA-1642: Default fromDate to epoch for SAR requests

munslowl:
- If no `fromDate` provided then default to `LocalDate.EPOCH` instead of today
- Use [shared module](https://github.com/ministryofjustice/hmpps-kotlin-lib/blob/main/readme-contents/SubjectAccessRequestEndpoint.md)


matt-crowson-moj (approved):
LGTM

PR #840, opened 2024-04-04
SAA-1646: enable RO access for service credentials to appropriate end…

Kristian-Gregory:
…points, specifically for PLP service here (#839)


Kristian-Gregory:
Abandoned, intended to fix commit msg but may not have worked or may have caused issues. not worth risk so leaving the commit msg with incorrect ticket num

PR #876, opened 2024-05-07
SAA-1688: Correct Activity Attendance Editing to Match NOMIS Constraints

dafhoustonmoj:
Implement the following rules on when attendance has can be recorded/changed:

If attendance has been recorded as attended or not attended & paid, stop any changes after the day the attendance was recorded
If attendance has been recorded as not attended & unpaid, changes can only be made for 7 days after the activity date
If attendance has not been recorded, the user can record attendance up to 14 days after the activity date


davidthomas22 (approved):
LGTM

PR #879, opened 2024-05-14
SAA-1655: Add max days to expiry to regime endpoint DTO

munslowl:
This is so the UI can determine that the prisoner's movement where they have left the prison occurred `x` days ago depending on the prison's `maxDaysToExpiry` value.


dafhoustonmoj (approved):
LGTM

PR #880, opened 2024-05-16
SAA-1655: Add original appointment id to event metric

munslowl:
If new appointment is copied from an existing appointment then add value to telemetry call using new `originalId` property.


davidthomas22 (approved):
LGTM

PR #897, opened 2024-05-29
SAA-1751: replace h2 with test containers postgres for integration tests

munslowl:
Replaced H2 with Postgres for integration tests.

Will use TestContainers locally unless there is already a Postgres DB on port 5432 in which case it will try to use that so maybe remap your port in [docker-compose.yml](https://github.com/ministryofjustice/hmpps-activities-management-api/blob/main/docker-compose.yml#L10).

CircleCI uses [java_postgres](https://circleci.com/developer/orbs/orb/ministryofjustice/hmpps#executors-java_postgres) orb to spin up postgres and then [waits for it](https://circleci.com/developer/orbs/orb/ministryofjustice/hmpps#commands-wait_till_ready_postgres) before running tests.

There were a few tests that were unnecessarily strict on ordering of lists so I have fixed those **but some more may arise**.

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/integration/AppointmentIntegrationTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/integration/AppointmentIntegrationTest.kt
@@ -832,13 +832,8 @@ class AppointmentIntegrationTest : IntegrationTestBase() {
     verify(eventsPublisher, times(12)).send(eventCaptor.capture())
 
     with(eventCaptor.allValues.filter { it.eventType == "appointments.appointment-instance.updated" }) {
-      size isEqualTo 12
-      assertThat(map { it.additionalInformation }).containsExactlyElementsOf(
-        // The update events for the specified appointment's instances are sent first
-        appointmentSeries.appointments.single { it.id == appointmentId }.attendees.map { AppointmentInstanceInformation(it.id) }
-          // Followed by the update events for the remaining instances
-          .union(appointmentSeries.appointments.filter { it.id != appointmentId }.flatMap { it.attendees }.map { AppointmentInstanceInformation(it.id) }),
-      )
+      assertThat(map { it.additionalInformation })
+        .hasSameElementsAs(appointmentSeries.appointments.flatMap { it.attendees }.map { AppointmentInstanceInformation(it.id) })
     }
 
     verifyNoMoreInteractions(eventsPublisher)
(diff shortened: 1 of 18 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/integration/AppointmentIntegrationTest.kt
```diff
@@ -832,13 +832,8 @@ class AppointmentIntegrationTest : IntegrationTestBase() {
...
     with(eventCaptor.allValues.filter { it.eventType == "appointments.appointment-instance.updated" }) {
-      size isEqualTo 12
-      assertThat(map { it.additionalInformation }).containsExactlyElementsOf(
-        // The update events for the specified appointment's instances are sent first
-        appointmentSeries.appointments.single { it.id == appointmentId }.attendees.map { AppointmentInstanceInformation(it.id) }
-          // Followed by the update events for the remaining instances
-          .union(appointmentSeries.appointments.filter { it.id != appointmentId }.flatMap { it.attendees }.map { AppointmentInstanceInformation(it.id) }),
-      )
+      assertThat(map { it.additionalInformation })
+        .hasSameElementsAs(appointmentSeries.appointments.flatMap { it.attendees }.map { AppointmentInstanceInformation(it.id) })
```
munslowl, line 836: Checked with syscon and message order is not important

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/integration/config/PostgresContainer.kt
```diff
@@ -0,0 +1,39 @@
...
+      setWaitStrategy(Wait.forListeningPort())
+      withReuse(true)
+
+      start()
+    }
+  }
+
+  private fun isPostgresRunning(): Boolean =
+    try {
+      val serverSocket = ServerSocket(5432)
```
munslowl, line 32: This means that local DB for general dev will need to use another port when running integration tests at the same time

src/test/resources/application-test.yml
```diff
@@ -16,17 +16,17 @@ spring:
...
 
   datasource:
-    url: 'jdbc:h2:mem:activities-db;MODE=PostgreSQL'
+    url: jdbc:postgresql://localhost:5432/activities
     username: activities
-    password: dummy
+    password: activities
+    hikari:
+      minimumIdle: 2
+      maximumPoolSize: 20
```
munslowl, line 24: Too many connections used by tests so had to reduce

dafhoustonmoj (approved):
LGTM

PR #908, opened 2024-06-07
SAA-1785 add purposeful activity report scripts (Sql)

KristianGregor:
Creates a new job that executes reports required for calculating Purposeful Activity and uploads them to the Analytical Platform S3 bucket.

This PR also contains work to introduce localstack into integration tests, locally and in the CI pipeline. Only new integration tests for Purposeful Activity use this at present and other tests will need to be adapted to use localstack in a future PR


helm_deploy/hmpps-activities-management-api/values.yaml
```diff
@@ -48,6 +48,8 @@ generic-service:
     SCHEDULE_AHEAD_DAYS_ONLINE: 60
     MAX_APPOINTMENT_INSTANCES: 20000
     MAX_SYNC_APPOINTMENT_INSTANCE_ACTIONS: 500
+    AWS_S3_AP_BUCKET: moj-reg-local-test
```
dafhoustonmoj, line 51: should this be placeholders?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/job/PurposefulActivityReportsJob.kt
```diff
@@ -0,0 +1,84 @@
...
+@Component
+class PurposefulActivityReportsJob(
+  private val jobRunner: SafeJobRunner,
+  private val reportRepo: PurposefulActivityRepositoryCustom,
+  private val s3Service: S3Service,
+) {
+
+  suspend fun execute(weekOffset: Int) {
+    jobRunner.runJob(
+      JobDefinition(JobType.PURPOSEFUL_ACTIVITY_REPORTS) {
```
dafhoustonmoj, line 22: maybe push the implementation to a service rather call the repo directly

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/repository/PurposefulActivityRepository.kt
```diff
@@ -0,0 +1,192 @@
...
+
+@Repository
+class PurposefulActivityRepositoryImpl : PurposefulActivityRepositoryCustom {
+  @PersistenceContext
+  private lateinit var entityManager: EntityManager
+
+  companion object {
+    private val log = LoggerFactory.getLogger(PurposefulActivityRepositoryImpl::class.java)
+  }
+
```
dafhoustonmoj, line 25: would we need to check the report data in support? If so maybe a report views would be better so we can query to the db easier

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/S3Service.kt
```diff
@@ -0,0 +1,59 @@
...
+import org.springframework.stereotype.Service
+import java.time.Instant
+import java.time.ZoneOffset
+import java.time.format.DateTimeFormatter
+
+@Service
+class S3Service(
+  private val s3ClientAnalyticalPlatform: S3Client,
+) {
+  @Value("\${aws.s3.ap.bucket}")
```
dafhoustonmoj, line 16: do we want a default in the code? or only from the config

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/integration/PurposefulActivityIntegrationTest.kt
```diff
@@ -0,0 +1,44 @@
...
+      .headers(setAuthorisation(roles = listOf()))
+      .exchange()
+      .expectStatus().isAccepted
+  }
+
+  @Sql(
+    "classpath:test_data/seed-purposeful-activity-appts.sql",
+    "classpath:test_data/seed-purposeful-activity-activities.sql",
+  )
+  @Test
```
dafhoustonmoj, line 27: is it possible to verify the report data?

helm_deploy/hmpps-activities-management-api/values.yaml
KristianGregor: Yeah these are junk values, they can be set in the environment-specific files and are so for each environment. Looks like the pattern is normally to omit env vars from the values.yml where they there's no generic default so I'll just delete from here.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/repository/PurposefulActivityRepository.kt
KristianGregor: Philosophical difference here between yourself and @timoj58 
I think Tim was keen for SQL to be in code rather than views/migration files. I may have interpreted this wrong here though, but if not, may need a quick chat to agree what the norm should be?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/S3Service.kt
KristianGregor: I was trying to avoid needing to update everyone's local .env file...but I think I've achieved that by having defaults on the application.yml if the env var isn't available.

So I think it'll all still work if i take out the defaults here, as long as you're happy with the defaults existing in the application.yml?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/repository/PurposefulActivityRepository.kt
dafhoustonmoj: If code is the only access point I'd agree, better in the code. I was more thinking if we expect to get adhoc support enquires it might be handy to have the view (End of the day we can still get the SQL from the code when required)

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/repository/PurposefulActivityRepository.kt
```diff
@@ -0,0 +1,174 @@
...
+      inner join appointment_attendee aa on aa.appointment_id = a.appointment_id
+      WHERE (a.start_date || ' ' || a.start_time)::timestamp BETWEEN
+         (SELECT start_date FROM date_range) AND
+         (SELECT end_date FROM date_range);
+  """
+
+  @Transactional
+  @Override
+  override fun getPurposefulActivityActivitiesReport(weekOffset: Int): MutableList<Any?>? {
+    val results = entityManager.createNativeQuery(
```
dafhoustonmoj, line 158: should use setParameter
e.g.
Query q = entityManager.createNativeQuery("SELECT .....");
q.setParameter("weekOffset", 1);

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/PurposefulActivityService.kt
```diff
@@ -0,0 +1,141 @@
...
+  }
+
+  private fun escapeCsvString(value: String): String {
+    // Escape double quotes by doubling them
+    val escapedValue = value.replace("\"", "\"\"")
+    // Enclose the value in quotes
+    return "\"$escapedValue\""
+  }
+
+  fun getNthPreviousSunday(n: Int): String {
```
dafhoustonmoj, line 127: not essential, but this could be useful outside this, made generic and put the util package 
e.g. getPrevious(offset: Int, dayOfWeek: DayOfWeek, format: String) : String

src/main/resources/application.yml
```diff
@@ -151,4 +151,11 @@ springdoc:
...
   appointments-live: ${APPOINTMENTS_LIVE:***}
-  prisons-live: ${PRISONS_LIVE:***}
\ No newline at end of file
+  prisons-live: ${PRISONS_LIVE:***}
+
+aws:
+  s3:
+    ap:
+      region: eu-west-2
+      bucket: ${AWS_S3_AP_BUCKET:dummy-bucket}
```
dafhoustonmoj, line 160: is this the correct bucket name?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/integration/PurposefulActivityIntegrationTest.kt
```diff
@@ -0,0 +1,108 @@
...
+  fun `Purposeful Activity Report activities report is uploaded to s3, downloaded again and verified`() {
+    val fileKey = purposefulActivityService.executeActivitiesReport(1)
+
+    runBlocking {
+      val file = s3Service.getFileFromS3(fileKey, "activities", purposefulActivityService.awsApS3BucketName)
+
+      val bufferedReader: BufferedReader = file.bufferedReader()
+      val csvContent = bufferedReader.use { it.readText() }
+
```
dafhoustonmoj, line 60: can csvContent be asserted?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/job/PurposefulActivityReportsJobTest.kt
```diff
@@ -0,0 +1,36 @@
...
+  @Test
+  fun `Create purposeful activity reports`() {
+    mockJobs(PURPOSEFUL_ACTIVITY_REPORTS)
+
+    runBlocking {
+      job.execute(1)
+    }
+
+    verify(safeJobRunner).runJob(jobDefinitionCaptor.capture())
+    // verify(service).create()
```
dafhoustonmoj, line 32: should this be commented out?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/S3ServiceTest.kt
```diff
@@ -0,0 +1,38 @@
...
+
+  @Test
+  fun `Push a file to AP AWS S3`(): Unit = runBlocking {
+    val mockBucketName = "aws-s3-ap-mock-bucket"
+    val putObjectResponse = PutObjectResponse {
+    }
+
+    coEvery { amazonS3.putObject(any()) } returns putObjectResponse
+
+    service.pushReportToAnalyticalPlatformS3("test-file".toByteArray(), "test-file.csv", "testTable", mockBucketName)
```
dafhoustonmoj, line 25: assert the returning filepath

src/main/resources/application.yml
KristianGregor: So in most contexts AWS_S3_AP_BUCKET will be set by environment variable, so the default is never used. I think its only not set in the circleci config at present for integration tests, and then it just needs to be some value, as long as it has a name it will create it prior to testing upload to that bucket. 

Given that i think that value may be best placed to live in the circleci config, so I'll give that a try.

dafhoustonmoj (approved):
LGTM

PR #923, opened 2024-06-24
SAA-1809 move views to repeatable

timoj58:


src/main/resources/migrations/common/R__v_all_attendance_summary.sql
munslowl, line 1: This view has already been dropped

PR #937, opened 2024-06-28
SAA-1758: Release prisoner attendance records

dafhoustonmoj:
Change deleting the attendance records to remove them when the session end date is in the future. Raise a delete event to synch back to Nomis


munslowl:
Is it worth adding test to [ManageAttendanceRecordsJobIntegrationTest](https://github.com/ministryofjustice/hmpps-activities-management-api/blob/main/src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/integration/ManageAttendanceRecordsJobIntegrationTest.kt)?

dafhoustonmoj:
It doesn't actually get called by the ManageAttendance job. Only in response to ActivityChangedEvents and PrisionerReleasedEvents

munslowl (approved):

PR #950, opened 2024-07-02
SAA-1839 improve performance (experimental)

timoj58:
unlikely to improve dashboard/id completely, this will not be the slowest query now, but the other is still over 1.5 seconds (allocations)

not ideal solution, more a case of improving what we have, (which is terrible).  ie move Pagination down a level, and reduce calls to  Allocation,  as the model structure is not great, and the current query uses hibernate to go up a level or two, to schedule, then to activity to find the prison code

Additionally, the front end actually needs very few fields, so the toModel calls are also costly, but not going to refactor all of that at present. 

This endpoint is only used on the dashboard so is in effect very inefficient given we simply dont need the majority of the data

Need to also factor in, this endpoint needs to get all the prisoners, and at some prisons that can be a large dataset, to parse over and over, even though the response only returns 5 rows, and this would get worse as we roll out.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/CandidatesService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/CandidatesService.kt
@@ -105,7 +105,7 @@ class CandidatesService(
     suitableIncentiveLevels: List<String>?,
     suitableRiskLevels: List<String>?,
     suitableForEmployed: Boolean?,
-    searchString: String?,
+    search: String?,
     pageable: Pageable,
   ): Page<ActivityCandidate> = coroutineScope {
     val schedule = activityScheduleRepository.findOrThrowNotFound(scheduleId)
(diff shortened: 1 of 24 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/CandidatesService.kt
```diff
@@ -95,54 +100,71 @@ class CandidatesService(
...
+
+    val prisonerAllocations = async {
+      allocationRepository.getCandidateAllocations(prisonCode = prisonCode)
+    }.await()
+
+    prisoners = prisoners
+      .filter { waitingList.none { w -> w.prisonerNumber == it.prisonerNumber } }
+      .filter {
+        filterByEmployment(
+          prisoner = it, prisonerAllocations = prisonerAllocations, suitableForEmployed = suitableForEmployed,
```
munslowl, line 141: Is it me or is `prisonerAllocations` always filtered down to `prisonerNumber`?

munslowl (dismissed):

munslowl (approved):

PR #955, opened 2024-07-03
SAA-1839 candidates performance (experimental)

timoj58:
slimmed down version after reverting previous PR
 
 this now has test data that will actually run the relevant code.  
 
 JPA query can be looked into next


dafhoustonmoj (approved):
LGTM

PR #977, opened 2024-07-23
SAA-1744 start to resolve issues around exclusions

timoj58:
few points around this, mainly the exclusions code and sql is flawed, and live.  These changes will resolve some issues, but the underlying faults remain

underlying faults

1.  exclusions keep creating records, so i can exclude and include a prisoner on the same day, and it records multiple records, which match the sql left join in effect sending duplicates back
2. exclusions - given the way its been coded (in the kotlin) it luckily doesnt pick up the wrong value, as the "end date" is set to now(), and only one record is not ended.  However, if someone was to exclude and include after midnight, the scheduler will incorrectly exclude, as now would now match now 
3. the UI, and exclusions in general.  We do not exclude people on the day, but the next day

solution wise for now 

1. view needs to ignore exclusions records when they start on the day of the session (ie point 3), this will allow the user to record attendance, (ie the bug - phantom allocations)
2. use distinct for now, as this is putting incorrect allocation numbers in dashboards, related to how many exclusion records the prisoner may have

What this doesnt address

1. the scheduling risk
2. the fact that the view "may" still present erroneous data, if an item exclusion and inclusion matched by the date filters (start to end date).   ie if both in and out match, the view will in effect say its included, even though it was actually excluded

### DIFF
--- a/src/main/resources/migrations/common/R__v_prisoner_scheduled_activities.sql
+++ b/src/main/resources/migrations/common/R__v_prisoner_scheduled_activities.sql
@@ -39,6 +39,7 @@ FROM scheduled_instance si
          JOIN activity_category category ON category.activity_category_id = act.activity_category_id
          LEFT JOIN exclusion ex ON alloc.allocation_id = ex.allocation_id AND
                                    ex.start_date <= si.session_date AND
+                                   ex.start_date != si.session_date AND
                                    (ex.end_date >= si.session_date OR ex.end_date IS NULL) AND
                                    ex.slot_start_time = si.start_time AND
                                    ex.slot_end_time = si.end_time AND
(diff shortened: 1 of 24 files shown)


src/main/resources/migrations/common/R__v_prisoner_scheduled_activities.sql
```diff
@@ -39,6 +39,7 @@ FROM scheduled_instance si
          JOIN activity_category category ON category.activity_category_id = act.activity_category_id
          LEFT JOIN exclusion ex ON alloc.allocation_id = ex.allocation_id AND
                                    ex.start_date <= si.session_date AND
+                                   ex.start_date != si.session_date AND
```
munslowl, line 42: Can't this be reduced to:

```sql
ex.start_date < si.session_date
```

munslowl (approved):

PR #978, opened 2024-07-24
postpone isle of wight (IWI) rollout

KristianGregor:
The original plan was to rollout IoW on 26th July. 
In the rollout_prison table there is now an entry for IoW (IWI) with dates of 26th Jul '24 but activities_to_be_rolled_out and appointments_to_be_rolled_out set to FALSE.

I believe that these dates are completely ineffectual until those flags are set to true and so this PR isn't necessary. Please could someone double check?


KristianGregor:
Determined there is no impact to the rollout date being set if the corresponding flags are not also set to TRUE, so abandoned this change.

PR #982, opened 2024-07-26
SAA-1890 iwi test rig, flag to skip exception

timoj58:
Can expand on this,  ie include exclusions, run scheduler later in test and so on.  and probably just put all the requests in json, and load from file at some point.

sample output added from the created activity bypassing exception


src/test/resources/samples/iwi_activity.json
```diff
@@ -0,0 +1,160 @@
+{
```
timoj58, line 1: sample output after activity created

KristianGregor (approved):

PR #984, opened 2024-07-29
SAA-1904 use custom time 

timoj58:
uses custom times on migrate, and add flag for whether it uses prison regime time or not

additional logic to update an activity slot with custom times, to avoid resetting to prison regime time

removed time_slot column as not used, and this is not the time to add it


natclamp-moj (approved):

dafhoustonmoj (approved):
LGTM

PR #1114, opened 2024-10-24
SAA-2092: change the activity update to also update the timeSlot when only the timeslot changes

dafhoustonmoj:
The scheduled instances recreated on a timeSlot change  did not reflect the new time slot when that was the only change. E.g. with a custom times activity the initial set up only had the incorrect timeslot.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityService.kt
```diff
@@ -504,6 +504,7 @@ class ActivityService(
 
   private fun ActivitySchedule.updateInstances() {
     this.removeRedundantInstances()
+    activityRepository.flush()
```
dafhoustonmoj, line 507: need to flush to ensure the deletes happen before the insert, otherwise when only the timeslot changes the unique index idx_scheduled_instance_schedule_id_date_times would fail

PR #1132, opened 2024-10-31
SAA-2065: Remove GET appointment unused endpoint

munslowl:


docker-compose-test.yml
munslowl, line 1: Just purring this in as I run individual integration tests often and this removes then need to have TestContainers spinning up a DB image every time, but at the price of Docker using extra resources.

Might be more relevant when Localstack is inclusded.

natclamp-moj (approved):

PR #1136, opened 2024-11-01
SAA-2140 enable aa early

dafhoustonmoj:
separate out the info service from the activities and appointments being live for the A&A service. This allows the system to be hidden, but useable, given early access for setting up a new prison regime.


davidthomas22 (approved):
LGTM

munslowl (approved):

PR #1157, opened 2024-11-12
SDIT-2238 Fix attendance reconciliation bug

mikehalmamoj:
There is a bug in the attendance reconciliation where a prisoner is deallocated from an activity and then reallocated on the same day (and then attended and marked with pay). The problem is that both allocations are considered "active" and so the attendance is counted towards the count twice. To fix we make sure the attendance is only counted once.

To find an example of the bad data run SQL `select * from allocation a where booking_id = 2694358 and activity_schedule_id = 553;` This scenario has been emulated in the test data for `seed-activity-id-27.sql`.


dafhoustonmoj (approved):
LGTM

PR #1222, opened 2024-12-17
Remove the 60 day restriction for viewing appointments at a location

smcveigh941:
The endpoint `/scheduled-events/prison/{prisonCode}/locations` was previously only used for generating movement lists.

Recently, it has also been used to allow the user to view the schedule of a room during the book a video link journey. The 60 day restriction is currently throwing the error up to the user when they try to create a booking > 60 days into the future. 

We believe the 60 day restriction was originally placed on this endpoint due to the fact that activity instanced are only scheduled 60 days in advance, however, appointments can be scheduled at any date in the future. This endpoint should be made available to view those appointments past the 60 day mark.


matt-crowson-moj (approved):
LGTM

PR #1246, opened 2025-01-15
SAA-2282: Add a query parameter to restrict rolled out prisons to only show live ones. 

dafhoustonmoj:
Add a query parameter to restrict rolled out prisons to only show live ones
Remove unused rollout date fields


davidthomas22 (approved):
LGTM

PR #1254, opened 2025-01-21
SAA-2279 prevent duplicate allocations

angela-pan-moj:
updated a few tests that were randomly failing.


src/main/resources/migrations/common/V2025.01.21__add_unique_constraint_to_allocations.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE allocation ADD UNIQUE (activity_schedule_id, prisoner_number, booking_id, allocated_time, prisoner_status);
```
munslowl, line 1: Just checking that the exception caused by this violation will be not specifically handled because this is an edge case?

munslowl: the allocated time is calculated as:

`LocalDateTime.now().truncatedTo(ChronoUnit.MINUTES),`

Does this mean user can still allocate the same prisoner at 11:59:59 and 12:00:00 ?

angela-pan-moj: removed allocated time from the script

src/main/resources/migrations/common/V2025.01.21__add_unique_constraint_to_allocations.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE allocation ADD CONSTRAINT unique_non_ended_prisoner_allocation_key
+    EXCLUDE (activity_schedule_id WITH =, prisoner_number WITH =, start_date WITH =, prisoner_status WITH =)
```
angela-pan-moj, line 2: having looked at the data in preprod, this should do what we intend (to prevent duplicate data in the db).

munslowl (approved):

PR #1286, opened 2025-02-03
SAA-2283 updated activity migration to use incoming start date

angela-pan-moj:
Need to wait for syscon to complete their bit before merging. (done)


KristianGregor (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/model/request/ActivityMigrateRequest.kt
```diff
@@ -26,7 +26,7 @@ data class ActivityMigrateRequest(
   @field:NotNull
   @Schema(description = "Date on which this activity starts or started. Can not be null.", example = "2022-12-23")
   @JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "yyyy-MM-dd")
-  val startDate: LocalDate?,
```
angela-pan-moj, line 29: As the startDate is mandatory and is not nullable in the object syscon passes in, I've updated this to a non-nullable type

munslowl (approved):

KristianGregor (approved):

PR #1309, opened 2025-02-12
SAA-2331: Set appointment start date limit to 5 years

munslowl:
Increasing the limit from 1 year to 5 years to allow some more consideration of what is correct.


davidthomas22 (approved):
LGTM

PR #1329, opened 2025-02-21
SAA-2284 remove risley and ranby locations regex

angela-pan-moj:


angela-pan-moj:
need to separate out risley and ranby as the locations team are going live with them seperately

PR #1404, opened 2025-03-27
SAA-2430: Use all non-residential locations

munslowl:
When determine locations and events for movements use all the prison's non-residential locations for resolving the actual locations as visits, etc can be assigned to any type of location not just locations that are designated as being of that usage type.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/InternalLocationService.kt
```diff
@@ -53,19 +54,25 @@ class InternalLocationService(
...
+
+    val locationsMap = locationsAsync
+      .await()
       .filter { dpsLocationIds.contains(it.id) }
       .associateBy { it.id }
 
-    val mappings = nomisMappingAPIClient.getLocationMappingsByDpsIds(dpsLocationIds).associateBy { it.dpsLocationId }
+    val mappings = mappingsAsync
+      .await()
+      .associateBy { it.dpsLocationId }
```
munslowl, line 69: Small tweak to get them to run concurrently

angela-pan-moj (approved):

dafhoustonmoj (approved):
LGTM

PR #1418, opened 2025-04-02
SAA-2387: Change exclusions and allocations to lazy loading

munslowl:
Remove eager loading from:
- allocations in activity schedule
- exclusions in allocation


rajaram-justice (approved):
LGTM

PR #1448, opened 2025-04-17
SAA-2429: Uncancel multiple scheduled instances

munslowl:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/model/request/ScheduleInstancesUncancelRequest.kt
```diff
@@ -0,0 +1,12 @@
+package uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.model.request
+
+import io.swagger.v3.oas.annotations.media.Schema
+import jakarta.validation.constraints.NotEmpty
+
+@Schema(description = "Request object for uncancelling multiple schedule instances")
+data class ScheduleInstancesUncancelRequest(
```
rajaram-justice, line 7: Can be renamed to ScheduledInstancesUncancelRequest to match with others.

munslowl: Yeah - `UncancelScheduledInstanceRequest` is wrongly named and the endpoint is now `@Deprecated` so we ill remove that one in due course

PR #1463, opened 2025-04-25
SAA-2414: Add endpoint to allow cancelled scheduled instance to be updated

munslowl:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/ScheduledInstance.kt
```diff
@@ -114,6 +114,34 @@ data class ScheduledInstance(
...
+
+    if (reason != null) {
+      cancelledReason = reason
+      comment = cancelComment
+      cancelledBy = updatedBy
+      cancelledTime = now
+    }
+
+    return attendances
+      .filterNot { issuePayment == null }
```
munslowl, line 140: If `issuePayment` is not set then we are not updating attendances.

natclamp-moj (approved):

PR #1476, opened 2025-05-01
Change queue adming role

munslowl:


src/main/resources/application.yml
```diff
@@ -102,7 +102,6 @@ spring:
     placeholders:
       dpr_user: ${DPR_USER}
       dpr_password: ${DPR_PASSWORD}
-    out-of-order: true
```
munslowl, line 105: Don't need this any more

src/main/resources/application.yml
```diff
@@ -157,7 +156,7 @@ applications:
 
 hmpps:
   sqs:
-    queueAdminRole: ACTIVITY_QUEUE_ADMIN
```
munslowl, line 160: Temporarily changed to allow devs to purge queue

rajaram-justice (approved):

PR #1508, opened 2025-05-22
SAA-3333: Ignore activity prisoners from movement list de-allocated from sessions later today

munslowl:
If a prisoner is deallocated from today and there are sessions later today then they will appear on the movement list for today.

As part of deallocation the API will remove the relevant attendance records for today.

Use the fact that there are no attendance records and that movement list is for today to filter out those prisoners.

NB: Attendance records are only create for one day.


davidthomas22 (approved):
LGTM

PR #1509, opened 2025-05-22
SAA-3333: Ignore activity prisoners from unlock list deallocated from sessions later today

munslowl:
If a prisoner is deallocated from today and there are sessions later today then they will appear on the unlock list for today.

As part of deallocation the API will remove the relevant attendance records for today.

Use the fact that there are no attendance records and that unlock list is for today to filter out those prisoners.

NB: Attendance records are only create for one day.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ScheduledEventService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ScheduledEventService.kt
@@ -468,6 +468,7 @@ class ScheduledEventService(
     date = date,
     timeSlot = slot,
   )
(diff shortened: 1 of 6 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ScheduledEventService.kt
```diff
@@ -468,6 +468,7 @@ class ScheduledEventService(
     date = date,
     timeSlot = slot,
   )
+    .filter { activity -> activity.sessionDate != LocalDate.now() || activity.attendanceStatus != null }
```
munslowl, line 472: If the session is for today and there are not activity records then the prisoner cannot be attending so remove from unlock list

rajaram-justice (approved):

PR #1539, opened 2025-06-09
SAA-2420: Recording edits to activity details with envers

rajaram-justice:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/config/JPAConfig.kt
```diff
@@ -0,0 +1,12 @@
...
+
+import org.springframework.context.annotation.Configuration
+import org.springframework.data.envers.repository.config.EnableEnversRepositories
+import org.springframework.data.jpa.repository.config.EnableJpaAuditing
+import org.springframework.data.jpa.repository.config.EnableJpaRepositories
+
+@Configuration
+@EnableJpaAuditing
+@EnableEnversRepositories
+@EnableJpaRepositories(basePackages = ["uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.repository"])
```
munslowl, line 11: Can these two be combined like [this](https://github.com/spring-projects/spring-data-examples/blob/ec8557c7da6b59499ec1f970a949ac2dedbd2624/jpa/envers/src/main/java/example/springdata/jpa/envers/EnversDemoConfiguration.java#L31)?

src/main/resources/migrations/common/V2025.06.09__create_activity_audit_table.sql
```diff
@@ -0,0 +1,34 @@
+CREATE TABLE revision
+(
+    id                           bigserial    NOT NULL CONSTRAINT revision_pkey PRIMARY KEY,
+    "timestamp"                  bigint       NOT NULL,
+    username                     varchar(100) NOT NULL
+);
+
+CREATE SEQUENCE revision_seq INCREMENT BY 50;
```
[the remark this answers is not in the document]
rajaram-justice: Hibernate envers uses this sequence to create revision entries, otherwise it will throw error.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/config/JPAConfig.kt
rajaram-justice: No, I couldn't. Getting compile time errors.

src/main/resources/migrations/common/V2025.06.09__create_activity_audit_table.sql
```diff
@@ -0,0 +1,34 @@
+CREATE TABLE revision
+(
+    id                           bigserial    NOT NULL CONSTRAINT revision_pkey PRIMARY KEY,
+    "timestamp"                  bigint       NOT NULL,
```
munslowl, line 4: Bit naff having timestamp as `bigint` if we need to query but looks like thats the way it is ... but looks like we have to live with it. 

I suppose we could add a generated column later?

src/main/resources/migrations/common/V2025.06.09__create_activity_audit_table.sql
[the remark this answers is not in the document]
rajaram-justice: Good catch. I initially tried with bigserial to avoid creating the sequence. As envers looking for revision_seq and not revision_id_seq, I had to create one explicitly and missed to change it to bigint. Will change it now.

src/main/resources/migrations/common/V2025.06.09__create_activity_audit_table.sql
munslowl: Yes. But also if we have a generated column which is not in the entity then we can also look at the data through SQL and understand it better.  But that can wait.

rajaram-justice: We can just convert the epoch time to date in the sql itself. Example below,
select id, to_char(to_timestamp("timestamp"/1000) ,'YYYY-MM-dd HH:MM:SS') as date_timestamp, username
	FROM public.revision

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/config/JPAConfig.kt
munslowl: Sorry - was having a play.

```@EnableJpaRepositories(repositoryFactoryBeanClass = EnversRevisionRepositoryFactoryBean::class, basePackages = ["uk.gov.justice.digital.hmpps.hmppsactivitiesmanagementapi.repository"])
``` worked for me but doesn't matter that much.

rajaram-justice: My bad - I was trying with @EnableEnversRepositories

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/entity/Activity.kt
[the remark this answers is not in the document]
rajaram-justice: Yep good point, added integration tests.

PR #1575, opened 2025-06-20
SAA-2421: Add job to fix DPS Location UUIDs in Appointment Series

munslowl:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/appointment/AppointmentCreateDomainService.kt
```diff
@@ -195,6 +195,7 @@ class AppointmentCreateDomainService(
     telemetryClient.trackEvent(TelemetryEvent.APPOINTMENT_SERIES_CREATED.value, propertiesMap, metricsMap)
   }
 
+  // TODO: SAA-2421 Add DPS Location ID
```
rajaram-justice, line 198: Will be done as part of separate ticket or missed to be removed?

PR #1601, opened 2025-06-30
SAA-3468: Handle missing times in Prison API external transfers response

munslowl:


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/helpers/PrisonApiFactory.kt
```diff
@@ -12,15 +11,6 @@ import java.time.LocalDate
 import java.time.LocalDateTime
 import java.time.LocalTime
 
-fun userCaseLoads(prisonCode: String) = listOf(
```
munslowl, line 15: No used

PR #1607, opened 2025-07-02
SAA-3414: Create history for existing pay rates

rajaram-justice:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/resource/MigrateActivityController.kt
```diff
@@ -266,4 +266,40 @@ class MigrateActivityController(
...
+          Content(
+            mediaType = "application/json",
+            schema = Schema(implementation = ErrorResponse::class),
+          ),
+        ],
+      ),
+    ],
+  )
+  @PreAuthorize("hasAnyRole('NOMIS_ACTIVITIES','ACTIVITY_ADMIN')")
+  fun createActivityPayHistory() = migrateActivityService.createActivityPayHistory()
```
munslowl, line 304: Is this purely for use whilst migrating activities?

If no then maybe belongs in another controller

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/MigrateActivityService.kt
```diff
@@ -149,6 +158,79 @@ class MigrateActivityService(
...
+
+    if (activityPayHistoryCount > 0) {
+      message = "Activities pay rate history migration can't be started as activity_pay_history table isn't empty"
+    } else {
+      activityPayCount = activityPayRepository.count()
+      if (activityPayCount <= 0) {
+        message = "Activities pay rate history migration can't be started as activity_pay table is empty"
+      } else {
+        transactionHandler.newSpringTransaction {
+          log.info("Activities pay rate history migration started for $activityPayCount records")
```
KristianGregor, line 174: That could be a big transaction. May put some memory pressure on the instance. You should make a note of the memory impact on preprod when you do that (if it doesn't fail outright). If it does fail for memory reasons, you may be able to get around it by temporarily allocating more mem to the API pods, just remember to scale them back down after you're finished.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/resource/MigrateActivityController.kt
rajaram-justice: Sounds good. Let me move it to UtilityController.

munslowl (approved):

PR #1637, opened 2025-07-21
SAA-3387: Migrate legacy case note id to UUID

rajaram-justice:


src/main/resources/migrations/common/V2025.07.18__combined_case_notes_related_changes.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE attendance              ADD COLUMN IF NOT EXISTS dps_case_note_id              VARCHAR(36);
+ALTER TABLE attendance_history      ADD COLUMN IF NOT EXISTS dps_case_note_id              VARCHAR(36);
+ALTER TABLE planned_suspension      ADD COLUMN IF NOT EXISTS dps_case_note_id              VARCHAR(36);
+ALTER TABLE planned_deallocation    ADD COLUMN IF NOT EXISTS dps_case_note_id              VARCHAR(36);
+ALTER TABLE allocation              ADD COLUMN IF NOT EXISTS deallocation_dps_case_note_id VARCHAR(36);
```
munslowl, line 5: Can these be uuids?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/repository/AttendanceRepository.kt
```diff
@@ -66,6 +66,33 @@ interface AttendanceRepository : JpaRepository<Attendance, Long> {
 
   fun findByPrisonerNumber(prisonerNumber: String): List<Attendance>
 
+  @Query(
+    value = """
+      SELECT a FROM Attendance a
+      WHERE a.caseNoteId IS NOT NULL
+    """,
+  )
+  fun findAllCaseNoteIdToMigrate(): List<Attendance>
```
munslowl, line 75: Do we think we can load all the attendance (~88k) record into mem ok?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/repository/AttendanceRepository.kt
```diff
@@ -66,6 +66,33 @@ interface AttendanceRepository : JpaRepository<Attendance, Long> {
...
+
+  @Modifying
+  @Query(
+    value = """
+      UPDATE Attendance a
+      SET a.dpsCaseNoteId = :dpsCaseNoteId
```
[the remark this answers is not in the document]
rajaram-justice: Possible. We can create a temporary index on case note id and then drop it as part of the main changes.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/repository/AttendanceRepository.kt
munslowl: Another possiblity is to maybe use interface projections which will be smaller and wont be managed by the entity manager.

There's also streams such as [here](https://github.com/ministryofjustice/hmpps-activities-management-api/blob/bdc1272d20cbc5ef8b723f1855f96e615d92bf6a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/repository/PurposefulActivityRepository.kt#L183)

PR #1707, opened 2025-08-29
SAA-3537: Enable SQS end allocations due to end job in PROD

munslowl:
- also add to check DLQ


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/integration/ManageAllocationsJobSqsIntegrationTest.kt
```diff
@@ -121,6 +126,42 @@ class ManageAllocationsJobSqsIntegrationTest : AbstractJobIntegrationTest() {
     verifyJobComplete(JobType.DEALLOCATE_ENDING)
   }
 
+  @Sql("classpath:test_data/seed-activity-id-11.sql")
+  @Test
+  fun `exception occurs when deallocating an allocation due to end and a message ends up on DLQ`() {
```
munslowl, line 131: Adding a test to check that fail causes message on DLQ.

PR #1775, opened 2025-10-06
SAA-3700: HMP SEND turn on A&A

natclamp-moj:


helm_deploy/values-preprod.yaml
munslowl, line 1: Need to do PROD alo

munslowl (approved):

PR #2058, opened 2026-04-15
SAA-4206: Amend create activity endpoint to skip location input

mitchell-dsilva:


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/ActivityServiceTest.kt
```diff
@@ -942,6 +942,80 @@ class ActivityServiceTest {
...
 
+  @Test
+  fun `create activity - success when external activity with outside work is true and no location`() {
+    val createExternalActivityRequest = mapper.read<ActivityCreateRequest>("activity/activity-create-request-6.json")
+      .copy(
+        startDate = TimeSource.tomorrow(),
+        inCell = false,
+        onWing = false,
+        offWing = false,
+        outsideWork = true,
```
ashleygyngell, line 955: I know activity-create-request-6 doesn't have dpsLocationId... but is it worth including it as `null` to make it clear its not testing on an internal location. Only realised it was missing because it was included in other tests.

munslowl (approved):

ashleygyngell (approved):

PR #2106, opened 2026-05-13
SAA-4238: Fix update appointment times in a series

munslowl:
Example error scenario:

- Create an appointment series, for 10 days, with start and end times of 10:00 to 11:00
- Update the third appointment **only** to set the end time to 10:30
- Update the start time for second appointment **to the end of the series** to 10:30

then the third appointment will **start and end** at 10:30.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsactivitiesmanagementapi/service/appointment/AppointmentService.kt
```diff
@@ -80,17 +81,19 @@ class AppointmentService(
...
-          }
-        }
+    appointmentsToUpdate.forEach {
+      val startTime = request.startTime ?: it.startTime
+      val endTime = request.endTime ?: it.endTime
+      require(endTime == null || endTime > startTime) {
+        val startDateStr = it.startDate.format(DateTimeFormatter.ofPattern("dd-MM-yyyy"))
+        "Appointment on $startDateStr would have a start time ($startTime) on or after the end time ($endTime)"
+      }
     }
```
munslowl, line 92: For each appointment that is being updated - check the start time is before the end time.

ashleygyngell (approved):
LGTM

mitchell-dsilva (approved):

PR #2216, opened 2026-07-28
SAA-4566: Fix for incorrect sequence of revision history

mitchell-dsilva:
**Issue:** There was an issue found when testing in DEV where sometimes the latest revision has a lower number as compared to an older revision. Its works fine locally and the issue only appears with multiple pods in deployed environments , where each pod pre-allocates its own block of 50 IDs from the sequence and hands them out independently. So when the successive requests got to different pods we are seeing inconsistency in revision numbers.

To fix this the history is now being returned sorted by updatedDateTime i.e. the revisionDateTime and not revision.


davidthomas22 (approved):
LGTM

